Issue #26, opened 2023-12-17
Use POST instead of GET for /ui endpoint


wwwillchen:
Got a cryptic error with Base-64 encoding which adds bloat to network payload, but fixed it in https://github.com/wwwillchen/mesop/commit/fdbce139dfa8ac92b2d2e6e2bfe5754cb99081d6. Also, it's a bit strange to expose this on a URL and may lead to issues down the road.

For future reference, here's a potential way to Base-64 encode:
https://github.com/google/closure-library/blob/7818ff7dc0b53555a7fb3c3427e6761e88bde3a2/closure/goog/crypt/base64.js#L161

wwwillchen:
https://github.com/mpetazzoni/sse.js

PR #40, opened 2024-01-27
/ui method changed to POST from GET

Maanikya:
Changed the method of flash app route /ui from GET to POST 

Ref #26


wwwillchen:
We also need to update the client callsite https://github.com/google/mesop/blob/7218e9401f1462c063828e8250da1ec9b7a0a831/mesop/web/src/services/channel.ts#L69 because EventSource only does a GET request and it doesn't do a POST request.

PR #78, opened 2024-03-21
Allow `flex-shrink` to be set to 0.

### LINKED ISSUE #69: Not able to set`flex-shrink=0`
When I try to set `flex-shrink=0`, it's not showing up in the generated CSS. I believe this is because 0 would be the same as undefined/null/etc in the current code. 

As a quick test, I was able to make a small change that worked (EDIT: Actually that does not work since the type is number, so the default I guess would be 0 -- I'm not too familiar with typescript):

```
  if (styleObj.getFlexShrink() || styleObj.getFlexShrink() === 0) {
    style += `flex-shrink: ${styleObj.getFlexShrink()};`;
  }
```

One reason for this issue is that that flex-shrink defaults to 1 (See https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink).

So it by not including the style, I'm not able to get the flex-shrink = 0 behavior that I was looking for.

This may be an issue with other styles. We may want to explicitly check for null/undefined for all styles. So if it is set, then we should use the value that is specified.

Source location: https://github.com/google/mesop/blob/24f77303b6b5f014ae882e74d35464b06a76632d/mesop/web/src/utils/styles.ts#L5

richard-to:
The `flex-shrink` css property could not be set to 0 since Mesop will ignore styles that are "empty".

Normally this is fine, but `flex-shrink` defaults to 1, which means the "empty" state will make `flex-shrink` equal to 1. This means that 0 cannot be set for `flex-shrink`.

To resolve this, we can change the date type to a string in the UI proto.

Example:

```
import mesop as me

@me.page(path="/flex_shrink")
def main():
  with me.box(style=me.Style(display="flex", flex_shrink="0")):
    me.text("Testing flex shrink")
```

Also here's a screenshot to check that the property is being set correctly:

Screenshot 2024-03-20 at 8 26 01 PM

Fixes #69

### DIFF
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -279,7 +279,7 @@ def to_style_proto(s: Style) -> pb.Style:
     flex_basis=s.flex_basis,
     flex_direction=s.flex_direction,
     flex_grow=s.flex_grow,
-    flex_shrink=s.flex_shrink,
+    flex_shrink=str(s.flex_shrink),
     flex_wrap=s.flex_wrap,
     font_family=s.font_family,
     font_size=_px_str(s.font_size),
--- a/mesop/protos/ui.proto
+++ b/mesop/protos/ui.proto
@@ -234,7 +234,7 @@ message Style {
     optional string flex_basis = 7;
     optional string flex_direction = 8;
     optional int32 flex_grow = 9;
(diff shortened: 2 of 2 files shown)


wwwillchen:
Thanks for contributing!

PR #86, opened 2024-03-27
Allow columns CSS property to accept int/str vals.

### LINKED ISSUE #79: The `columns` property should be string instead of int.
When I was looking at the `columns` property in https://developer.mozilla.org/en-US/docs/Web/CSS/columns, it seems that it allows more than integers.

Is there a reason only integers are allowed? Otherwise seems fairly straightforward to just convert the `columns` property to a string.

richard-to:
The columns property allows int/strs according to
https://developer.mozilla.org/en-US/docs/Web/CSS/columns

# Example code

To test this change, I tested out the following code.

```
import mesop as me

@me.page(path="/columns")
def main():
  with me.box(style=me.Style(columns="150px 3")):
    me.text("first_column")
    me.text("second_column")

  with me.box(style=me.Style(columns=2)):
    me.text("first_column")
    me.text("second_column")
```

# Example of columns defined using strings

Screenshot 2024-03-27 at 3 55 24 PM

# Example of columns defined using ints

Screenshot 2024-03-27 at 3 55 47 PM

# Note

I also tried to update the columns component in labs.layout.py, but ran into an issue there trying to change the field definition to int | str. So for now, just left it as int.

This led to this error: `Mesop Internal Error: ('Unhandled param type', int | str, 'field_name', 'columns')`

Which leads to: mesop/mesop/editor/component_configs.py

It's probably related to the logic around here. Seems like we handle the int | str | None case, but not int | str.

```
    elif param_type is str or (
      # special case, for int|str|None (used for styles, e.g. pixel value), use str
      args
      and len(args) == 3
      and args[0] is int
      and args[1] is str
      and args[2] is NoneType
    ):
```

Fixes #79

### DIFF
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -173,7 +173,7 @@ class Style:
   box_sizing: str | None = None
   color: str | None = None
   column_gap: int | str | None = None
-  columns: int | None = None
+  columns: int | str | None = None
   cursor: str | None = None
   display: Literal[
     # precomposed values
@@ -273,7 +273,7 @@ def to_style_proto(s: Style) -> pb.Style:
     box_sizing=s.box_sizing,
     color=s.color,
     column_gap=_px_str(s.column_gap),
-    columns=s.columns,
+    columns=str(s.columns),
     cursor=s.cursor,
     display=s.display,
     flex_basis=s.flex_basis,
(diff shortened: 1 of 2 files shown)


wwwillchen:
Thanks for this change!

Re: the labs columns - I think we should probably get rid of it / replace it with something better. The problem with how it's implemented is that it doesn't actually support the SxS layout that I think most people want when they use something like columns. I think using CSS Grid, instead of the CSS columns property is actually going to provide the more intuitive behavior, but let's chat about this some more. I think you've got some good idea around layouts :)

PR #88, opened 2024-03-29
Document component key (closes #74)

### LINKED ISSUE #74: Document how to use `key` (as part of components API)
It's useful to do things like resetting component state (by creating a new component instance by incrementing the key for example):

**Example**

```py
def reset(event):
  state = me.state(State)
  state.select_menu_key += 1

def main():
  state = me.state(State)
  
  me.select(key=str(state.select_menu_key), 
            options=[me.SelectOption("o1", "o1"), me.SelectOption("o2", "o2")])
  me.button(label="Reset", on_click=reset)
```

wwwillchen:
- Update component API reference to link to component key doc section

### DIFF
--- a/docs/guides/components.md
+++ b/docs/guides/components.md
@@ -53,3 +53,68 @@ def page1():
 ```
 
 This is similar to Angular's [Content Projection](https://angular.io/guide/content-projection).
+
+## Component Key
+
+Every native component in Mesop accepts a `key` argument which is a component identifier. This is used by Mesop to tell [Angular whether to reuse the DOM element](https://angular.io/api/core/TrackByFunction#description).
+
+### Resetting a component
+
+You can reset a component to the initial state (e.g. reset a [select](../components/select.md) component to the unselected state) by giving it a new key value across renders.
+
+For example, you can reset a component by "incrementing" the key:
+
+```py
+class State:
+  select_menu_key: int
+
+def reset(event):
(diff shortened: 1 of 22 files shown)


richard-to (approved):
LGTM - Definitely a good thing to document. They key attribute is very very useful.

docs/guides/components.md
```diff
@@ -53,3 +53,64 @@ def page1():
...
+def main():
+  state = me.state(State)
+  me.select(key=str(state.select_menu_key),
+            options=[me.SelectOption(label="o1", value="o1")])
+  me.button(label="Reset", on_click=reset)
+```
+
+### Event handlers
+
+Every Mesop event includes the key of the component which emitted the event. This makes it useful when you want to reuse an event handler for multiple instances of a component:
```
richard-to, line 84: Not sure if it belongs in this part of the doc. I'm tempted to suggest also explaining/showing the anti-pattern for this.

Specifically the case where we try to dynamically generate a function.

So variations like this where it is not obvious why this would end up taking redirecting to the last url used.

```
def link_widget(url: str):
   def on_click(event: me.ClickEvent):
     me.navigate(url)
  return me.button(url, on_click=on_click)
```

wwwillchen: Good idea - I added a section in the troubleshooting section and linked it bi-directionally.

PR #89, opened 2024-03-29
Document input race condition issue anti-pattern

wwwillchen:


richard-to:
LGTM. Good idea to document this use case.

richard-to (approved):

PR #91, opened 2024-04-01
Switch to ipv6 host

wwwillchen:
Why? Running Colab (internally) requires IPv6, which isn't currently supported when using "0.0.0.0"

Using `::` ensures that this will work for for both IPv6 and IPv4 environments, according to:
- https://stackoverflow.com/a/21689979

Background:
- https://superuser.com/questions/515379/what-is-the-difference-between-and-0-0-0-0-from-the-netstat-an-output
- https://en.wikipedia.org/wiki/0.0.0.0?useskin=vector

Closes https://github.com/google/mesop/issues/102

### DIFF
--- a/mesop/cli/cli.py
+++ b/mesop/cli/cli.py
@@ -102,7 +102,7 @@ def main(argv: Sequence[str]):
     log_startup(port=port())
     logging.getLogger("werkzeug").setLevel(logging.WARN)
 
-  flask_app.run(host="0.0.0.0", port=port(), use_reloader=False)
+  flask_app.run(host="::", port=port(), use_reloader=False)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


wwwillchen:
Ready for review - PTAL.

richard-to (approved):
LGTM

tsal:
> ensures that this will work for for both IPv6 and IPv4 environments

Except this doesn't work in colab if you're in an environment that doesn't have IPV6 support.

wwwillchen:
> > ensures that this will work for for both IPv6 and IPv4 environments
> 
> Except this doesn't work in colab if you're in an environment that doesn't have IPV6 support.

What environment are you using?

tsal:
> > > ensures that this will work for for both IPv6 and IPv4 environments
> > 
> > 
> > Except this doesn't work in colab if you're in an environment that doesn't have IPV6 support.
> 
> What environment are you using?

An environment that doesn't have IPv6 enabled, OS is macOS in this case, chrome browser.

wwwillchen:
Got it, I've filed an issue here: https://github.com/google/mesop/issues/491

PR #99, opened 2024-04-04
Add a text to image IO function.

richard-to:
Also adds an example usage of the function.

I kept the implementation the same as the text_io one. Mainly just copied the code and made a few adjustments. I just copied the code to make it easier for others to copy if they wanted to add more functionality.

Here is a sample screenshot that uses a transform function that just returns the same image every time (note: need to pick a a few sample images to use  and also host somewhere).

Screenshot 2024-04-04 at 2 06 55 PM

### DIFF
--- a/mesop/examples/__init__.py
+++ b/mesop/examples/__init__.py
@@ -19,5 +19,6 @@ from mesop.examples import simple as simple
 from mesop.examples import sxs as sxs
 from mesop.examples import testing as testing
 from mesop.examples import text_io as text_io
+from mesop.examples import text_to_image as text_to_image
 
 # Do not import error_state_missing_init_prop because it cause all examples to fail.
--- a/mesop/labs/text_to_image.py
+++ b/mesop/labs/text_to_image.py
@@ -1,0 +1,121 @@
+from typing import Callable
+
+import mesop as me
+
+
+@me.stateclass
+class State:
+  input: str
+  output: str
+  textarea_key: int
+
+
+def text_to_image(
+  transform: Callable[[str], str],
+  *,
+  title: str | None = None,
+):
+  """Creates a simple UI which takes in a text input and returns an image output.
+
+  This function creates event handlers for text input and output operations
+  using the provided function `transform` to process the input and generate the image
+  output.
+
+  Args:
+    transform: Function that takes in a string input and returns a URL to an image or a base64 encoded image.
+    title: Headline text to display at the top of the UI.
+  """
+
+  def on_input(e: me.InputEvent):
+    state = me.state(State)
+    state.input = e.value
+
+  def on_click_generate(e: me.ClickEvent):
+    state = me.state(State)
+    state.output = transform(state.input)
+
+  def on_click_clear(e: me.ClickEvent):
+    state = me.state(State)
+    state.input = ""
+    state.output = ""
+    state.textarea_key += 1
+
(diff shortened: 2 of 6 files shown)


mesop/labs/text_to_image.py
```diff
@@ -0,0 +1,132 @@
...
+  """
+
+  def on_input(e: me.InputEvent):
+    state = me.state(State)
+    state.input = e.value
+
+  def on_click_generate(e: me.ClickEvent):
+    state = me.state(State)
+    output = transform(state.input)
+    if isinstance(output, types.GeneratorType):
```
wwwillchen, line 38: For image, it's a little less obvious the use case for generators. Typically, I'd expect the image to be returned all in one piece?

WDYT about having output being a string first, and then making it support generators if we see a use case?

mesop/labs/text_to_image.py
```diff
@@ -0,0 +1,132 @@
...
+      state.output = cast(str, output)
+      yield
+
+  def on_click_clear(e: me.ClickEvent):
+    state = me.state(State)
+    state.input = ""
+    state.output = ""
+    state.textarea_key += 1
+
+  with me.box(
```
wwwillchen, line 54: I'm wondering if we should reuse some of these styles across the i/o components. I think if we do a third i/o component, it may make sense to dedupe some of the styles (where applicable) to make it easier to maintain.

mesop/examples/__init__.py
```diff
@@ -19,5 +19,6 @@
 from mesop.examples import sxs as sxs
 from mesop.examples import testing as testing
 from mesop.examples import text_io as text_io
```
wwwillchen, line 21: no action needed: but food for thought, I think `text_to_image` is a good name, but I'm now wondering if `text_io` was a good name 😅 . It may be worth making `text_to_text` an alias for `text_io` later.

wwwillchen (approved):
Awesome - LGTM

wwwillchen:
Also, please add markdown docs for this.
Example: https://github.com/google/mesop/blob/main/docs/components/text_io.md
Add a line here: https://github.com/google/mesop/blob/0d60e4a6627a508c9dc0e28b3631f553b9981ac3/mkdocs.yml#L17
Steps to run docs locally: https://google.github.io/mesop/internal/development/#docs

mesop/labs/text_to_image.py
richard-to: Good catch. I agree that accepting string only for now is good.

mesop/labs/text_to_image.py
richard-to: I agree that we can leave as is for now. And agree that we can dedupe some stuff that is reusable.

mesop/examples/__init__.py
richard-to: Yes, I agree. Yeah I think you mentioned text_to_text last time we talked.

richard-to:
> Also, please add markdown docs for this. Example: https://github.com/google/mesop/blob/main/docs/components/text_io.md Add a line here:
> 
> https://github.com/google/mesop/blob/0d60e4a6627a508c9dc0e28b3631f553b9981ac3/mkdocs.yml#L17
> 
> 
> Steps to run docs locally: https://google.github.io/mesop/internal/development/#docs

Added docs page.

PR #100, opened 2024-04-04
Add simple chat widget function.

richard-to:
Also adds an example usage of the chat widget.

## Functionality

- Provide title
- Provide a bot name
- Can stream chat response
- When generating a response, new prompts cannot
- be sent
- Relatively responsive to device width

## Minor issues

- Chat window does not scroll along with new text being added.
- Text input does not clear when pressing entering a prompt.

### Desktop

Screenshot 2024-04-04 at 3 10 10 PM

### Mobile

Screenshot 2024-04-04 at 3 10 29 PM

### Tablet

Screenshot 2024-04-04 at 3 10 56 PM

### DIFF
--- a/mesop/examples/chat.py
+++ b/mesop/examples/chat.py
@@ -1,0 +1,40 @@
+import random
+import time
+
+import mesop as me
+import mesop.labs as mel
+
+
+@me.page(path="/chat", title="Lorem Ipsum Chat")
+def page():
+  mel.chat(
+    lorem_ipsum_chat, title="Lorem Ipsum Chat", bot_user="Lorem Ipsum Bot"
+  )
+
+
+def lorem_ipsum_chat(input: str, history: list[mel.ChatMessage]):
+  lines = [
+    (
+      "Lorem ipsum dolor sit amet, consectetur adipiscing elit, "
+      "sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
+    ),
+    "Laoreet sit amet cursus sit amet dictum sit amet.",
+    "At lectus urna duis convallis.",
+    "A pellentesque sit amet porttitor eget.",
+    "Mauris nunc congue nisi vitae suscipit tellus mauris a diam.",
+    "Aliquet lectus proin nibh nisl condimentum id.",
+    "Integer malesuada nunc vel risus commodo viverra maecenas accumsan.",
+    "Tempor id eu nisl nunc mi.",
+    "Id consectetur purus ut faucibus pulvinar.",
+    "Mauris pharetra et ultrices neque ornare.",
+    "Facilisis magna etiam tempor orci.",
+    "Mauris pharetra et ultrices neque.",
+    "Sit amet facilisis magna etiam tempor orci.",
+    "Amet consectetur adipiscing elit pellentesque habitant morbi tristique.",
+    "Egestas erat imperdiet sed euismod.",
+    "Tincidunt praesent semper feugiat nibh sed pulvinar proin gravida.",
+    "Habitant morbi tristique senectus et netus et malesuada.",
+  ]
+  for line in random.sample(lines, random.randint(3, len(lines) - 1)):
+    time.sleep(0.3)
+    yield line + " "
--- a/mesop/labs/chat.py
+++ b/mesop/labs/chat.py
@@ -1,0 +1,227 @@
+import time
+from dataclasses import dataclass
+from typing import Callable, Generator, Literal
+
+import mesop as me
+
+Role = Literal["user", "assistant"]
+
+_ROLE_USER = "user"
+_ROLE_ASSISTANT = "assistant"
+
+_BOT_USER_DEFAULT = "mesop-bot"
+
+_COLOR_BACKGROUND = "#f0f4f8"
+_COLOR_CHAT_BUBBLE_YOU = "#f2f2f2"
+_COLOR_CHAT_BUBBLE_BOT = "#ebf3ff"
+
+_DEFAULT_PADDING = me.Padding(top=20, left=20, right=20, bottom=20)
+_DEFAULT_BORDER_SIDE = me.BorderSide(
+  width="1px", style="solid", color="#ececec"
+)
+
+_LABEL_BUTTON = "Send prompt"
+_LABEL_BUTTON_IN_PROGRESS = "Processing prompt..."
+_LABEL_INPUT = "Enter your prompt"
+
+_STYLE_APP_CONTAINER = me.Style(
+  background=_COLOR_BACKGROUND,
+  display="grid",
+  height="100vh",
+  grid_template_columns="repeat(1, 1fr)",
+)
+_STYLE_TITLE = me.Style(padding=me.Padding(left=10))
+_STYLE_CHAT_BOX = me.Style(
+  height="100%",
+  overflow_y="scroll",
+  padding=_DEFAULT_PADDING,
+  margin=me.Margin(bottom=20),
+  border_radius="10px",
+  border=me.Border(
+    left=_DEFAULT_BORDER_SIDE,
+    right=_DEFAULT_BORDER_SIDE,
+    top=_DEFAULT_BORDER_SIDE,
+    bottom=_DEFAULT_BORDER_SIDE,
+  ),
+)
+_STYLE_CHAT_INPUT = me.Style(width="100%")
+_STYLE_CHAT_INPUT_BOX = me.Style(padding=me.Padding(top=30))
+_STYLE_CHAT_BUBBLE_NAME = me.Style(
+  font_weight="bold",
+  font_size="12px",
+  padding=me.Padding(left=15, right=15, bottom=5),
+)
+_STYLE_CHAT_BUBBLE_PLAINTEXT = me.Style(margin=me.Margin(top=15, bottom=15))
+
+
+def _make_style_chat_ui_container(has_title: bool) -> me.Style:
+  """Generates styles for chat UI container depending on if there is a title or not.
+
+  Args:
+    has_title: Whether the Chat UI is display a title or not.
+  """
+  return me.Style(
+    display="grid",
+    grid_template_columns="repeat(1, 1fr)",
+    grid_template_rows="1fr 14fr 1fr" if has_title else "5fr 1fr",
+    margin=me.Margin(top=0, bottom=0, left="auto", right="auto"),
+    width="min(1024px, 100%)",
+    height="100vh",
+    background="#fff",
+    box_shadow=(
+      "0 3px 1px -2px #0003, 0 2px 2px #00000024, 0 1px 5px #0000001f"
+    ),
+    padding=_DEFAULT_PADDING,
+  )
+
+
+def _make_style_chat_bubble_wrapper(role: Role) -> me.Style:
+  """Generates styles for chat bubble position.
+
+  Args:
+    role: Chat bubble alignment depends on the role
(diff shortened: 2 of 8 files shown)


mesop/examples/chat.py
```diff
@@ -0,0 +1,40 @@
+import random
+import time
+
+import mesop as me
+import mesop.labs as mel
+
+
+def lorem_ipsum_chat(input):
```
wwwillchen, line 8: nit: I would move this below `def page`; we can use this example and display it in the docs, and it's better to show the main code upfront.

wwwillchen: Also: add a type annotation for input (mostly for better documentation)

mesop/examples/chat.py
```diff
@@ -0,0 +1,40 @@
...
+  ]
+  for line in random.sample(lines, random.randint(3, len(lines) - 1)):
+    time.sleep(0.5)
+    yield line + " "
+
+
+@me.page(path="/chat", title="Lorem Ipsum Chat")
+def page():
+  mel.chat(
+    lorem_ipsum_chat, title="Lorem Ipsum Chat", bot_username="Lorem Ipsum Bot"
```
wwwillchen, line 39: bike-shed: WDYT about using`bot_name` as the parameter name? it's shorter without losing clarity.

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+
+  def _on_update(e: me.InputEvent):
+    state = me.state(State)
+    setattr(state, e.key, e.value)
+
+  return _on_update
+
+
+def chat(
+  transform: Callable[[str], Generator[str, None, None] | str],
```
wwwillchen, line 156: I think transform should have two parameters:
1. prompt/input: str
2. history: list[ChatMessage] - it could potentially be a list of tuples like how [Gradio](https://www.gradio.app/guides/creating-a-chatbot-fast#defining-a-chat-function) does it, but I think defining a dataclass is more robust (e.g. if we need a new field , it won't break existing apps).

mesop/examples/chat.py
wwwillchen, line 1: I think this example is good for testing, but it would be good to have a more realistic example later (e.g. calling the actual gemini API) - maybe doing it in a Colab so users can plug in their own API key?

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+)
+
+_LABEL_BUTTON = "Send prompt"
+_LABEL_BUTTON_IN_PROGRESS = "Processing prompt..."
+_LABEL_INPUT = "Enter your prompt"
+
+_STYLE_APP_CONTAINER = me.Style(
+  background=_COLOR_BACKGROUND,
+  display="grid",
+  height="100vh",
```
wwwillchen, line 28: why not use 100%?

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+    background="#fff",
+    box_shadow=(
+      "0 3px 1px -2px #0003, 0 2px 2px #00000024, 0 1px 5px #0000001f"
+    ),
+    padding=_DEFAULT_PADDING,
+  )
+
+
+def _make_style_chat_bubble_wrapper(
+  role: Literal["user", "assistant"]
```
wwwillchen, line 79: I'd extract this into a reusable type, e.g.

```py
Role = Literal["user", "assistant"]
```

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+  """
+  background = (
+    _COLOR_CHAT_BUBBLE_YOU if role == _ROLE_USER else _COLOR_CHAT_BUBBLE_BOT
+  )
+  return me.Style(
+    width="80%",
+    font_size="13px",
+    background=background,
+    border_radius="15px",
+    padding=me.Padding(
```
wwwillchen, line 108: remove? (empty should be equivalent of 0px)

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+    yield
+
+    start_time = time.time()
+    output_message = transform(state.input)
+    assistant_message = Message(role=_ROLE_ASSISTANT, username=bot_username)
+    output.append(assistant_message)
+    state.output = output
+    for content in output_message:
+      assistant_message.content += content
+      if (time.time() - start_time) >= 0.5:
```
wwwillchen, line 193: I'd leave a comment here on why you picked 0.5. 

fwiw: I found in practice the right number is a bit tricky (too low, and we're spending too much time on the render loop; too high and we're not giving as much feedback as we could) so we may want to make this customizable in the future.

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+            if msg.role == _ROLE_ASSISTANT:
+              me.text(msg.username, style=_STYLE_CHAT_BUBBLE_NAME)
+            with me.box(style=_make_chat_bubble_style(msg.role)):
+              if msg.role == _ROLE_USER:
+                me.text(msg.content, style=_STYLE_CHAT_BUBBLE_PLAINTEXT)
+              else:
+                me.markdown(msg.content)
+
+      with me.box(style=_STYLE_CHAT_INPUT_BOX):
+        me.input(
```
wwwillchen, line 215: To clear the input, you could try doing something like:

```py
value = "" if state.clear_input else None
```

and then in `on_click`, you do state.clear_input = True, yield, state.clear_input = False, etc.

Alternatively, you could use this trick: https://google.github.io/mesop/guides/components/#resetting-a-component (it would lose the focus though).

mesop/labs/chat.py
```diff
@@ -0,0 +1,228 @@
...
+    ),
+  )
+
+
+@dataclass
+class Message:
+  """Chat message metadata."""
+
+  role: Literal["assistant", "user"] = "user"
+  username: str = ""
```
wwwillchen, line 129: It's a little duplicative to store username here since it's the same across all the messages for a given role, you could store it on the State class (this is a bit more efficient, and IMO clearer)

wwwillchen (changes requested):
Exciting! Left some comments.

Similar to https://github.com/google/mesop/pull/99#issuecomment-2038432204, let's add docs for this.

Also, this is getting a bit complicated, so I would recommend adding an e2e test, example:
https://github.com/google/mesop/tree/main/mesop/tests/e2e

info on testing: https://google.github.io/mesop/internal/testing/

mesop/labs/chat.py
richard-to: Fair point. Updated.

mesop/labs/chat.py
richard-to: I tried your suggestion but it didn't work for me with the text input. I ended up using the key increment trick which worked.

I agree the downside is that it loses focus. But I think it loses focus when the user presses the send button right now. It definitely would be good to be able to control the focus somehow.

mesop/labs/chat.py
richard-to: Yeah, I think I just copied it from one of the Mesop chat examples somewhere. I lowered it to .25 for now.

mesop/examples/chat.py
richard-to: Yeah, I agree. I'll add a Github issue. Doing a colab one would be cool. Could also do one for the text to image one using Vertex AI API for image generation,

mesop/labs/chat.py
richard-to: Hmm. I agree that providing a history is a good idea. One downside of using a dataclass is that it forces the implementor of the transform function to account for Mesop specific behavior. So they'd need to add some light translation of the history to work with their LLM API whether formatting history into the prompt is necessary. But Gradio also has the same problem using primitives.

So I agree that a dataclass seems better.

Updated.

mesop/labs/chat.py
wwwillchen: SG this works for now, but yeah, we should add a focus API. I've filed a ticket: https://github.com/google/mesop/issues/106

mesop/labs/chat.py
```diff
@@ -0,0 +1,224 @@
...
+      bottom=_DEFAULT_BORDER_SIDE,
+    ),
+  )
+
+
+@dataclass
+class ChatMessage:
+  """Chat message metadata."""
+
+  role: Literal["assistant", "user"] = "user"
```
wwwillchen, line 121: Use the Role type alias?

mesop/labs/chat.py
```diff
@@ -0,0 +1,224 @@
...
+    border=me.Border(
+      left=_DEFAULT_BORDER_SIDE,
+      right=_DEFAULT_BORDER_SIDE,
+      top=_DEFAULT_BORDER_SIDE,
+      bottom=_DEFAULT_BORDER_SIDE,
+    ),
+  )
+
+
+@dataclass
```
wwwillchen, line 117: I'd make this  kw-only to make it more flexible for future evolution.

```
@dataclass(kw_only=True)
```

mesop/labs/chat.py
wwwillchen: I think Mesop developers doing some translation is a bit inevitable because I've seen a wide range of how history is formatted across internal and external APIs.

wwwillchen (approved):
Nice - LGTM. BTW, I've given you Maintainer's role so you can merge your own PRs now.

richard-to:
> Exciting! Left some comments.
> 
> Similar to [#99 (comment)](https://github.com/google/mesop/pull/99#issuecomment-2038432204), let's add docs for this.
> 
> Also, this is getting a bit complicated, so I would recommend adding an e2e test, example: https://github.com/google/mesop/tree/main/mesop/tests/e2e
> 
> info on testing: https://google.github.io/mesop/internal/testing/

Ok, got the e2e tests added to this pull request.

richard-to:
@wwwillchen I got the tests playwright tests to pass. I fixed it by removing the worker explicit worker config of 8. I think that ended up slowing down the CI VM because there were too many requests. I don't think it affected the performance of the CI tests. The test run was 8:14. Previous passing CI runs for other pull requests ranged from 8:30-9:00 minutes. So I think this should be OK.

Let me know if that sounds good to you. Otherwise I'll look into other solutions.

Other things I tried, I did try increasing the time outs for my tests, but I think it was struggling to update the input text. The thing about the input text updates, is that we need to use an explicit timeout (e.g. something like await page.waitForTimeout(2000);) If we don't wait, then the input text may not be saved to backend state yet. The button press will be a noop since text is required.

wwwillchen:
> @wwwillchen I got the tests playwright tests to pass. I fixed it by removing the worker explicit worker config of 8. I think that ended up slowing down the CI VM because there were too many requests. I don't think it affected the performance of the CI tests. The test run was 8:14. Previous passing CI runs for other pull requests ranged from 8:30-9:00 minutes. So I think this should be OK.
> 
> Let me know if that sounds good to you. Otherwise I'll look into other solutions.
> 
> Other things I tried, I did try increasing the time outs for my tests, but I think it was struggling to update the input text. The thing about the input text updates, is that we need to use an explicit timeout (e.g. something like await page.waitForTimeout(2000);) If we don't wait, then the input text may not be saved to backend state yet. The button press will be a noop since text is required.

SGTM. BTW, I didn't realize until just now that leaving off the workers actually defaults to 50% of cores according to https://github.com/microsoft/playwright/issues/28122. We could try to do something like 100% when local and 50% (or even 1, for no parallelism) in CI, but this is easy to tweak around later. Let's submit it :)

Issue #75, opened 2024-03-19
Add `table` component


wwwillchen:
- https://material.angular.io/components/table/overview
- In particular, Python developers oftentimes want to display a dataframe as a table UI

richard-to:
I'd like to try taking on this issue. I think it'd be a good way for me to get a better understanding of the codebase.

wwwillchen:
Fantastic! I think this will be very useful - this was requested by an internal user and I would also like to use this as well :)

A couple of ideas around the API, but please come up with your own ideas!

- I'd consider having a simple API, e.g. `me.table(df)` where df is a Pandas dataframe. This is a very common use case. As prior art, take a look at what [Streamlit has done](https://docs.streamlit.io/library/api-reference/data/) with its dataframe and table components.
- As a follow-up feature, it may be nice (I have a personal use case :)) to provide some kind of `on_row_click` API so users can do custom interactivity (e.g. navigate to a page). There's a lot of potential interactivities (e.g. pagination, column sorting, filtering, etc), so feel free to be iterative about this and implement something simple (e.g. a static table) to begin with.

In case you haven't seen it, take a look at the docs for creating a new component: https://google.github.io/mesop/internal/new_component/#potential-exploration-areas

In particular, the scaffold_component.sh script is quite helpful.

richard-to:
I definitely agree about starting off simple first. Definitely agree that there's a lot of functionality that can be added. The Streamlit stuff is very cool.

wwwillchen:
Basic table component done in above PR - we can open more issues based on user feedback.

PR #103, opened 2024-04-05
Add basic table with Pandas support

richard-to:
This change adds a barebones Angular table that can take a Pandas DataFrame as input.

Current issues:

- Table headers not bold
- Style support (haven't looked into this yet)

Example usage:

```
import pandas
import mesop as me

@me.page(path="/components/table/e2e/table_app")
def app():
  with me.box(
    style=me.Style(
      padding=me.Padding(top=10, right=10, left=10, bottom=10), width=500
    )
  ):
    d = {
      "Bools": [True, False],
      "Ints": [3, 4],
      "Floats": [2.3, 4.5],
      "Strings": ["Hello", "World"],
    }
    df = pandas.DataFrame(data=d)
    me.table(me.TableDataSourcePandasDataFrame(df=df))
```

Screenshot 2024-04-04 at 6 13 26 PM

Issue Reference: #75

### DIFF
--- a/mesop/components/table/table.py
+++ b/mesop/components/table/table.py
@@ -1,0 +1,39 @@
+from typing import Any
+
+import mesop.components.table.table_pb2 as table_pb
+from mesop.component_helpers import insert_component, register_native_component
+
+
+# Don't include type hint since Pydantic can't properly type check the Pandas data
+# frame. In addition, we don't want to include Pandas as a dependency into Mesop.
+@register_native_component
+def table(data_frame: Any):
+  """
+  This function creates a table from Pandas data frame
+
+  Args:
+      data_frame: Pandas data frame.
+  """
+  insert_component(
+    type_name="table",
+    proto=table_pb.TableType(
+      displayed_columns=list(data_frame.columns),
+      data_source=_to_data_source(data_frame),
+    ),
+  )
+
+
+def _to_data_source(data_frame) -> list[table_pb.TableRow]:
+  """Convert Pandas data frame for display as a table.
+
+  All values will be converted to strings for display purposes on the frontend.
+  """
+  columns = list(data_frame.columns)
+  data = []
+  for df_row in data_frame.itertuples():
+    data.append(
+      table_pb.TableRow(
+        row={col_name: str(getattr(df_row, col_name)) for col_name in columns}
+      )
+    )
+  return data
--- a/build_defs/defaults.bzl
+++ b/build_defs/defaults.bzl
@@ -96,3 +96,7 @@ THIRD_PARTY_PY_PYTEST = [
 THIRD_PARTY_PY_MYPY_PROTOBUF = [
     requirement("mypy-protobuf"),
 ]
+
+THIRD_PARTY_PY_PANDAS = [
+    requirement("pandas"),
+]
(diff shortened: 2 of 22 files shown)


mesop/components/table/table.py
```diff
@@ -0,0 +1,68 @@
...
+        elif isinstance(value, float):
+          row[col_name] = table_pb.TableValue(float_value=value)
+        else:
+          raise ValueError(f"Unsupported Mesop table data type: {type(value)}")
+      data.append(table_pb.TableRow(row=row))
+    return data
+
+
+@register_native_component
+def table(data_source: TableDataSourcePandasDataFrame):
```
wwwillchen, line 52: The parameter should just take in a Pandas DataFrame, which would be a much cleaner callsite: `me.table(df)`. We can wrap it internally with a dataclass if needed.

mesop/components/table/table.py
```diff
@@ -0,0 +1,68 @@
...
+          row[col_name] = table_pb.TableValue(int_value=value)
+        elif isinstance(value, float):
+          row[col_name] = table_pb.TableValue(float_value=value)
+        else:
+          raise ValueError(f"Unsupported Mesop table data type: {type(value)}")
+      data.append(table_pb.TableRow(row=row))
+    return data
+
+
+@register_native_component
```
wwwillchen, line 51: Not sure if validation is failing for you, but I had to skip validation for the plot component which also used a Protocol: https://github.com/google/mesop/blob/0d60e4a6627a508c9dc0e28b3631f553b9981ac3/mesop/components/plot/plot.py#L19

You may need to modify https://github.com/google/mesop/blob/0d60e4a6627a508c9dc0e28b3631f553b9981ac3/mesop/component_helpers/helper.py#L299 to allow accepting an argument for `skip_validation` a la https://github.com/google/mesop/blob/0d60e4a6627a508c9dc0e28b3631f553b9981ac3/mesop/component_helpers/helper.py#L100-L116

wwwillchen: I'm not really sure what's the right thing to do here. If we want to do a proper validation check (e.g. checking isinstance with the actual Pandas DataFrame class, or Matplotlib Figure), then we'll end up importing some pretty hefty third-party library into Mesop (which every Mesop developer will depend on, even if they don't use it).

I think skipping validation checking is reasonable - they'll probably get a more cryptic error, but in the future we could do something a little bit magical like check the class name via `__name__` (https://g.co/gemini/share/6ac3898f6693), but this feels a little bit too fancy for now.

wwwillchen (commented):
Overall - looks like a good start; left some pointers on the pydantic validation issue.

Biggest question, I think, is whether to import pandas into Mesop. In general, I'd like to keep the dependency list small, but I'm not sure if it'll be tricky to convert various types of pandas values into something easily displayable (e.g. a string, number)

mesop/components/table/table.py
richard-to: Sure. That sounds fine. I mainly did this initially because of the Pydantic type checks. Thought I could trick it into working, but it didn't work. So I'll updated to just pass in the dataframe.

mesop/components/table/table.py
richard-to: Cool. I'll just just skip the validation check. Definitely agree about the third party dependency issue.

wwwillchen (approved):
LGTM

PR #112, opened 2024-04-12
Add __version__ attribute so Mesop users can easily tell which version of Mesop they're using

### LINKED ISSUE #104: Add `__version__`
Commonly used so that users can inspect what version of Mesop is loaded (useful for debugging), e.g.

- https://github.com/google/generative-ai-python/blob/7d2c17e4738316455212526c4f909ce668b62676/google/generativeai/__init__.py#L85C1-L86C1
- https://github.com/tensorflow/tensorboard/blob/942431445d5153e519705072c6455d6cd57391d3/tensorboard/version.py#L18

wwwillchen:
Fixes #104 

pip_package is a bit strangely organized (it was modeled after https://github.com/tensorflow/tensorboard/tree/master/tensorboard/pip_package)

### DIFF
--- a/docs/internal/publishing.md
+++ b/docs/internal/publishing.md
@@ -4,7 +4,7 @@ Follow these instructions for releasing a new version of Mesop publicly via PyPI
 
 ## Bump the version
 
-Update `mesop/pip_package/setup.py` version field to the next version.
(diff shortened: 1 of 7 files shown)


richard-to (approved):
LGTM

PR #113, opened 2024-04-12
Make MesopException the parent class of more specific mesop exceptions & expose as public API

### LINKED ISSUE #111: Expose Mesop Exception classes in public API
This will allow Mesop app developers to catch Mesop exceptions and handle them.

- http://google3/third_party/py/mesop/exceptions/__init__.py

wwwillchen:
Closes #111.

This makes it possible for Mesop app developers to catch Mesop exceptions and having them all subclass from a single Mesop exception class also makes it more convenient to catch any kind of Mesop exception.

### DIFF
--- a/mesop/__init__.py
+++ b/mesop/__init__.py
@@ -111,6 +111,18 @@ from mesop.events import (
 from mesop.events import (
   InputEvent as InputEvent,
 )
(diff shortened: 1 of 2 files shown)


richard-to (approved):
LGTM

PR #116, opened 2024-04-18
Update hot reloading logic in execute_module.py

richard-to:
This change separates the hot reloading logic for the mesop module and external app modules.

Before this fix, if the app module is outside of the mesop module, then changes to code in labs would not be reflected in the hot reload since the mesop directory is not also the app module.


wwwillchen:
Thanks for the fix. LGTM, but could you add a test case to https://github.com/google/mesop/blob/5457ce5ec138429158cdb8f5f415e14ec5e6d4cd/mesop/cli/execute_module_test.py#L16?

PR #122, opened 2024-04-20
Create shorthands for Margin, Padding, and Border

### LINKED ISSUE #119: Support shorthand syntax for Padding and Margin
Copied from discussion doc:

One thing we can do is to have helper functions for some of our UI data classes like Margin and Padding. Oftentimes, it's quite repetitive, where you're writing things like:

```py
with me.box(
style=me.Style(margin=me.Margin(top=16, left=16, right=16, bottom=16))
):
```

Flutter has a nice API for doing this with EdgeInsets (basically Padding) [1] that we can borrow and do something like this:

```py
me.Margin.all(14)
```

and:

```py
me.Margin.symmetric(horizontal=12, vertical=6)
```

I haven't done this out of laziness :) but also because it would break visual editor support (it's possible to support it in visual editor, but it would be a bit of work to update the codemod logic). Although TBH, I don't know if anyone is really using visual editor.

[1] [https://api.flutter.dev/flutter/painting/EdgeInsets-class.html](https://www.google.com/url?q=https://api.flutter.dev/flutter/painting/EdgeInsets-class.html&sa=D&source=docs&ust=1713467325333225&usg=AOvVaw26oIAA7wRbbz2YV4BeqSv2)

wwwillchen:
Fixes #119

### DIFF
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -61,6 +61,37 @@ class Border:
   bottom: BorderSide | None = None
   left: BorderSide | None = None
 
+  @staticmethod
+  def all(value: BorderSide) -> "Border":
+    """
+    Creates a Border instance with all sides having the same style.
+
+    Args:
+        value: The style to apply to all sides of the border.
+
+    Returns:
+        Border: A new Border instance with the specified style applied to all sides.
+    """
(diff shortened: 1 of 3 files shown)


richard-to (approved):
LGTM

Issue #82, opened 2024-03-22
Add uploader component


richard-to:
An uploader component could be really useful for use cases where one needs to upload an image/file/audio/video for sending to a model.

It does not seem there is an Angular Material component for this though.

richard-to:
Add initial uploader component. There are definitely improvements we can make which we can create separate issues.

PR #124, opened 2024-04-20
Add uploader component.

richard-to:
This component will render a basic file input. On upload the component will send the data in data URL format to Mesop server where the user can then validate the data and use it for their application.

Use cases include multi-modal use cases where you want to upload an image, audio, or video to an multimodal model.

The uploader does not support the following features:

- Drag and drop
- Image/audio/video preview
- Multi file upload
- Does not try to limit file size on the frontend
- Image will be uploaded immediately on upload (i.e. no click event to trigger the upload)
- Also no styling exposed yet.

Screenshot 2024-04-28 at 4 30 10 PM

Ref: #82


mesop/components/uploader/uploader.py
```diff
@@ -0,0 +1,72 @@
...
+    key=key.key, **json.loads(event.string_value)
+  ),
+)
+
+
+@register_native_component
+def uploader(
+  *,
+  key: str | None = None,
+  accept: str = "",
```
wwwillchen, line 45: I see you're following the native web API here and this is where I think it may make sense to deviate a little bit to make it more pythonic. IIUC,  `accept` is a comma-separated list-like string (for whatever reason HTML attributes don't really have a native list type). I think having this be an actual list[str] is more pythonic and then naming it something like `accept_types` makes it more obvious this is a list and not a str.

As reference:
- https://docs.streamlit.io/develop/api-reference/widgets/st.file_uploader
- https://www.gradio.app/docs/uploadbutton

mesop/components/uploader/uploader.py
```diff
@@ -0,0 +1,72 @@
...
+
+@dataclass(kw_only=True)
+class UploadChangeEvent(MesopEvent):
+  """Event representing a change in the uploader component's value.
+
+  Attributes:
+      name: The file's name.
+      size: The size of the file in bytes.
+      type: The file's MIME type.
+      data_url: Uploaded content as data URL.
```
wwwillchen, line 23: One of the main value propositions with Mesop is that it interoperates well with the broader Python ecosystem (e.g. the table component directly accepts Pandas dataframe). In this case, I think returning a data URL feels a bit clunky and not very Pythonic.

Instead of returning a data_url, I think we should return a "file" or something file-like.

Take a look at what [streamlit does](https://docs.streamlit.io/develop/api-reference/widgets/st.file_uploader#:~:text=The%20UploadedFile%20class%20is%20a%20subclass%20of%20BytesIO%2C%20and%20therefore%20it%20is%20%22file%2Dlike%22.%20This%20means%20you%20can%20pass%20them%20anywhere%20where%20a%20file%20is%20expected.). They [subclass Python's io.BytesIO](https://github.com/streamlit/streamlit/blob/0367e69e9fd5ac9fc9fc24a62ed75cf0c0d44a84/lib/streamlit/runtime/uploaded_file_manager.py#L56) which gives it file-like semantics. It's nice because then you can use it in Python APIs like `pd.read_csv(uploaded_file)`

mesop/components/uploader/uploader.ng.html
```diff
@@ -0,0 +1,5 @@
+<input
```
wwwillchen, line 1: One thing we need to watch out for is that it's actually [quite difficult to style](https://blog.angular-university.io/angular-file-upload/#:~:text=Why%20don%27t%20we%20see%20file%20inputs%20very%20often%3F) the native file input element.

Can we use the angular material button as the base UI for this? It seems like what most people would want (and a button is fairly easy for app developers to customize).

wwwillchen (changes requested):
Nice work. I added a few suggestions. I'd take a look at [streamlit](https://docs.streamlit.io/develop/api-reference/widgets/st.file_uploader) and [gradio's](https://www.gradio.app/docs/uploadbutton) component for reference.

mesop/components/uploader/uploader.py
richard-to: Good point. I agree that that is a better API.

mesop/components/uploader/uploader.ng.html
richard-to: That's a good point. I've definitely done the hidden file input trick before. Thanks for the link for how to do it in Angular.

I ended up using the angular button. But I had some questions about this:

1. Should I use the mesop button instead (if so how would I get it imported?)
2. I didn't expose any styling for the button yet. Mainly, I think we can probably improve our uploader with some more features like drag/drop, multiple uploads, and such, which I think would change the look of the component. For now I just kept it as basic as possible.

mesop/components/uploader/uploader.py
richard-to: Thanks for the pointer. I agree that makes more sense. Initially I didn't realize I could send bytes back to the server, but once you pointed our the proto serialization, then that made it easy. 

I ended up storing the file contents as BytesIO (but didn't subclass the file metadata class as bytesio). So api is slightly more verbose. Would end up being `pd.read_csv(uploaded_file.contents)`.

mesop/components/uploader/uploader.ng.html
wwwillchen: LGTM. I think using the Angular button is fine here. SG with keeping the styling simple, we could expose the style attribute later, but something basic is a good start.

mesop/components/uploader/e2e/uploader_app.py
```diff
@@ -0,0 +1,39 @@
...
+      me.text(f"File size: {state.size}")
+      me.text(f"File type: {state.mime_type}")
+
+    with me.box(style=me.Style(margin=me.Margin.all(10))):
+      me.image(src=state.contents)
+
+
+def handle_upload(event: me.UploadEvent):
+  state = me.state(State)
+  for file in event.files:
```
wwwillchen, line 35: This is kind of funky, it's basically overwriting the same properties on state over again. I think this technically workers since this is just a single file upload right now, but it's kind of confusing as an example. Instead, I'd do something like assert that files is length 1 and only read the first element.

mesop/components/uploader/uploader.py
```diff
@@ -0,0 +1,91 @@
...
+  """
+
+  name: str
+  size: int
+  mime_type: str
+  contents: io.BytesIO
+
+
+@dataclass(kw_only=True)
+class UploadEvent(MesopEvent):
```
wwwillchen, line 33: I think it's better to have the API be `file` and just a singular UploadedFile. I know you mentioned supporting multiple file uploads in the future, I think the current API should make single file upload (which IMO will be the common case) as ergonomic as possible.

In the future, if we want to support multiple file uploads, we have a couple options. We could add another attribute here `files` and then `file` would simply be an alias to the first element of `files`. Or we could have two different events (e.g. UploadEvent, UploadMultipleFilesEvent) and emit the different events by overloading the uploader function (e.g. have a parameter like `accept_multiple_files`). I'm thinking the first option isn't bad - having two attributes `file` and `files` on the same event, but we can decide this later.

mesop/components/uploader/uploader.py
wwwillchen: I'm curious why you decided to not subclass? I'm not sure if subclassing BytesIO is hacky or not, but it does make it more convenient to use.

wwwillchen (commented):
Left a few comments mostly on API. It's a tricky API, but this is looking quite good.

mesop/components/uploader/e2e/uploader_app.py
richard-to: Good point. I took your suggestion from the other comment to just use `file` for now.

mesop/components/uploader/uploader.py
richard-to: I have a preference for composition over inheritance. So in this case it seemed the contents was really the BytesIO part. And the rest was just metadata.

However, I can definitely see the argument for subclassing the uploaded file as well. And I don't have a strong opinion on this either way.

So I've updated it to be a subclass instead.

mesop/components/uploader/uploader.py
richard-to: That's great point. I've definitely found some annoyance with how the Javascript api does this too. I've updated to use a file parameter. I agree that using files with an alias to file would be fine when we add multi file uploads.

wwwillchen (approved):
LGTM - thanks!

PR #125, opened 2024-04-20
Add on table cell click event

richard-to:
The table cell click event will return the column name and row Index which will allow us to refer back to the same row or column in the Pandas data frame.

Screenshot 2024-04-20 at 3 16 13 PM

Ref: #118

### DIFF
--- a/mesop/components/table/table.proto
+++ b/mesop/components/table/table.proto
@@ -2,11 +2,27 @@ syntax = "proto2";
 
 package mesop.components.table;
 
+message TableClickEvent {
+    optional int64 row_index = 1;
+    optional int64 col_index = 2;
+}
+
 message TableRow {
-    map<string, string> row = 1;
+    // Pandas Index.
+    optional int64 index = 1;
+    // Column values are stored as a list corresponding to the `displayed_columns`
+    // field.
+    //
+    // This is done because the Pandas may change the column name in certain situations
+    // such as:
+    //
+    // - Naming a column `Index` which conflicts with the special Pandas `Index` column.
+    // - Naming a column with spaces, such as "Date Time"
+    repeated string cell_values = 2;
 }
 
 message TableType {
     repeated string displayed_columns = 1;
     repeated TableRow data_source = 2;
+    optional string on_table_click_event_handler_id = 3;
 }
--- a/mesop/components/table/table.py
+++ b/mesop/components/table/table.py
@@ -1,24 +1,66 @@
-from typing import Any
+from dataclasses import dataclass
+from typing import Any, Callable
 
 import mesop.components.table.table_pb2 as table_pb
-from mesop.component_helpers import insert_component, register_native_component
+from mesop.component_helpers import (
+  insert_component,
+  register_event_handler,
+  register_event_mapper,
+  register_native_component,
+)
+from mesop.events import MesopEvent
+
+
+@dataclass(kw_only=True)
+class TableClickEvent(MesopEvent):
+  """Event representing a click on the table component cell.
+
+  Attributes:
+      row_index: DataFrame row index of the clicked cell in the table.
+      col_index: DataFrame col index of the clicked cell in the table.
+      key (str): key of the component that emitted this event.
+  """
+
+  row_index: int
+  col_index: int
+
+
+def map_table_click_event(event, key):
+  click_event = table_pb.TableClickEvent()
+  click_event.ParseFromString(event.bytes_value)
+  return TableClickEvent(
+    key=key.key,
+    row_index=click_event.row_index,
+    col_index=click_event.col_index,
+  )
+
+
+register_event_mapper(TableClickEvent, map_table_click_event)
 
 
 # Don't include type hint since Pydantic can't properly type check the Pandas data
 # frame. In addition, we don't want to include Pandas as a dependency into Mesop.
 @register_native_component
-def table(data_frame: Any):
+def table(
+  data_frame: Any, *, on_click: Callable[[TableClickEvent], Any] | None = None
+):
   """
   This function creates a table from Pandas data frame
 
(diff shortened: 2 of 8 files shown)


mesop/components/table/table.py
```diff
@@ -27,13 +65,19 @@ def _to_data_source(data_frame) -> list[table_pb.TableRow]:
...
+  same cell on the server to handle user events.
   """
   columns = list(data_frame.columns)
   data = []
   for df_row in data_frame.itertuples():
     data.append(
       table_pb.TableRow(
-        row={col_name: str(getattr(df_row, col_name)) for col_name in columns}
+        row={
+          **{col_name: str(getattr(df_row, col_name)) for col_name in columns},
```
wwwillchen, line 78: There's an edge case here if the one of the columns is called "Index". Pandas seems to rename the column into something else (i.e. _1) to avoid a name collision with the special Index column.

Repro:

https://colab.research.google.com/drive/1TDDNOSFm2jnSEuLwTTqBgPexKUJgr10n#scrollTo=Mizd7VJuzgto

I would add this to table_app.py just to make sure this case is handled correctly.

mesop/components/table/table.ts
```diff
@@ -28,4 +32,18 @@ export class TableComponent {
...
+
+  onClickCell(row: TableRow, index: number): void {
+    // On cell click events, return the column name and row Index so we can refer back
+    // to the same cell on the server.
+    const userEvent = new UserEvent();
+    userEvent.setHandlerId(this.config().getOnTableClickEventHandlerId()!);
+    userEvent.setKey(this.key);
+    const value = {
+      col_name: this._config.getDisplayedColumnsList()[index],
+      row_index: parseInt(row.getRowMap().get('Index')),
```
wwwillchen, line 44: This is a bit problematic because in downstream, I believe the Closure Compiler will minify these property names (e.g. "col_name" could become "a1"). You could technically wrap these in quotes (i.e. `"col_name:"`) to have the compiler not obfuscate them, but it's still not great because this isn't that typesafe (e.g. a typo in the object key here isn't caught at compile time).

A  different approach is to add a new field to UserEvent with the type `bytes` and then serialize/deserialize a table-specific proto message (defined in table.proto). This similar to what we do the [Type#bytes](https://github.com/google/mesop/blob/64467d43f00d807add09bcf9f77e5e460d039b65/mesop/protos/ui.proto#L203) field.

wwwillchen (changes requested):
Mostly looks good - couple of issues though.

mesop/components/table/table.ts
richard-to: Nice! Thanks for the great tip. That's definitely a lot better.

mesop/components/table/table.py
richard-to: Good catch.

I ended up storing the columns as a repeated list. That should also fix a bug where column names with spaces wouldn't work such as "Date Times".

mesop/components/table/table.proto
```diff
@@ -2,11 +2,27 @@ syntax = "proto2";
...
+    // Pandas Index.
+    optional int64 index = 1;
+    // Column values are stored as a list corresponding to the `displayed_columns`
+    // field.
+    //
+    // This is done because the Pandas may change the column name in certain situations
+    // such as:
+    //
+    // - Naming a column `Index` which conflicts with the special Pandas `Index` column.
+    // - Naming a column with spaces, such as "Date Time"
```
wwwillchen, line 20: I think it's a little confusing to call this row. I'd name it something like `cell_values.

mesop/components/table/table.py
```diff
@@ -1,24 +1,66 @@
...
+from mesop.events import MesopEvent
+
+
+@dataclass(kw_only=True)
+class TableClickEvent(MesopEvent):
+  """Event representing a click on the table component cell.
+
+  Attributes:
+      row_index: DataFrame row index of the clicked cell in the table.
+      col_name: DataFrame col index of the clicked cell in the table.
```
wwwillchen, line 20: update: should be col_index now

mesop/components/table/table.ts
```diff
@@ -28,4 +33,17 @@ export class TableComponent {
   config(): TableType {
     return this._config;
   }
+
+  onClickCell(row: TableRow, index: number): void {
+    // On cell click events, return the column name and row Index so we can refer back
```
wwwillchen, line 38: "column name" -> "column index"

mesop/components/table/table.ts
```diff
@@ -28,4 +33,17 @@ export class TableComponent {
...
+    // to the same cell on the server.
+    const userEvent = new UserEvent();
+    userEvent.setHandlerId(this.config().getOnTableClickEventHandlerId()!);
+    userEvent.setKey(this.key);
+    const clickEvent = new TableClickEvent();
+    clickEvent.setColIndex(index);
+    clickEvent.setRowIndex(row.getIndex() as number);
```
wwwillchen, line 45: I'd add a comment here on why this cast is necessary.

mesop/components/table/table.py
```diff
@@ -1,24 +1,66 @@
...
-def table(data_frame: Any):
+def table(
+  data_frame: Any, *, on_click: Callable[[TableClickEvent], Any] | None = None
+):
   """
   This function creates a table from Pandas data frame
 
   Args:
       data_frame: Pandas data frame.
+      on_click: [click](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click_event) is a native browser event.
```
wwwillchen, line 52: I'd mention that this is triggered whenever a table cell is clicked (similar to your comment in TableClickEvent).

wwwillchen (approved):
LGTM - left a few minor comments

PR #130, opened 2024-04-21
Fix setup_proto_py_modules.sh

wwwillchen:
@richard-to not sure if you've setup VS Code type-checking - there's [some instructions here](https://google.github.io/mesop/internal/development/#venv), but I just recently fixed it with this PR and this earlier docs fix: https://github.com/google/mesop/commit/11c8d36d98dbe2477c09224cd41ee22e2082ab15


richard-to (approved):
LGTM

PR #131, opened 2024-04-23
Update default UI for radio component

wwwillchen:
- Changes default label position to "after" which is what [Angular Material defaults](https://github.com/angular/components/blob/4719da2c34bee87095a2368334715e415f18a4c3/src/material/radio/radio.ts#L121) to and what checkbox currently defaults to.
- Sets the display type to `block` as this is the most common behavior.

I will open a separate PR so that radio (& other components) can be styled, so Mesop app developers can override this.


richard-to (approved):
Cool. The assets thing you added is interesting. LGTM.

PR #137, opened 2024-04-23
Increase timeout in chat_test due to flakiness

wwwillchen:
Not sure if there's a better way but I noticed it's been flaky.
Example: https://github.com/google/mesop/actions/runs/8795330360?pr=135

### DIFF
--- a/mesop/examples/e2e/chat_test.ts
+++ b/mesop/examples/e2e/chat_test.ts
@@ -13,7 +13,7 @@ test('Chat UI can send messages and display responses', async ({page}) => {
   await page.getByRole('button', {name: 'Send prompt'}).click();
   await expect(
     page.getByRole('button', {name: 'Processing prompt...'}),
-  ).toBeAttached({timeout: 3000});
+  ).toBeAttached({timeout: 5000});
   await expect(page.locator('//input')).toHaveValue('');
   expect(page.locator('//div[text()="Lorem ipsum"]')).toHaveCount(1);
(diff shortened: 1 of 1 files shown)


mesop/examples/e2e/chat_test.ts
```diff
@@ -13,7 +13,7 @@ test('Chat UI can send messages and display responses', async ({page}) => {
   await page.getByRole('button', {name: 'Send prompt'}).click();
   await expect(
     page.getByRole('button', {name: 'Processing prompt...'}),
-  ).toBeAttached({timeout: 3000});
+  ).toBeAttached({timeout: 5000});
```
richard-to, line 16: I didn't get a chance to look at the playwright output, but it's possible time out issue could be here:

```
 // Need to wait for the input state to be saved before clicking.
 await page.waitForTimeout(2000);
```

The reason is that if we don't wait long enough, "Lorem ipsum" won't get saved to state. So when we click the button, there will be no input saved, which is a noop in the chat app logic.

That line is could be particularly flaky since there is no signal that the state has been saved. It just waits the 2 seconds.

richard-to (approved):
LGTM. We can probably monitor if the flakiness continues.

wwwillchen:
> LGTM. We can probably monitor if the flakiness continues.

SG - I think this will fix _some_ of the flakiness (in at least one test failure screenshot I saw the button display "Processing prompt... " which seems like the test just timed out a little too early).

PR #145, opened 2024-04-24
Deflakes chat_test.ts

wwwillchen:
Changes:

- Creates a minimal chat app with no sleep to speed up the test
- Skips assertions on the in-progress state 'Processing prompt...' - this seems inherently flaky because it's possible to miss it (if it happens too quickly now that there's no sleep)

This test isn't quite as high fidelity as before (but still covers most of the functionality) but it greatly alleviates the flakiness.

Checked (passed 10/10 times):

```sh
$ yarn playwright test  mesop/examples/e2e/chat_test.ts --repeat-each=10
```


richard-to (approved):
Nice. Good idea to use a more minimal example for the e2e test without the sleep delay.

Issue #51, opened 2024-02-02
Show interactive examples in components docs


wwwillchen:

wwwillchen:
Need to update markdown to link to the new code examples in `demo/` and the live demo in: https://mesop-y677hytkra-uc.a.run.app/

wwwillchen:
Fixed in #233

PR #147, opened 2024-04-24
Create a Mesop Demos app

wwwillchen:
Creates a Mesop Demos app to showcase various examples which consists of the live Mesop app and its corresponding code side-by-side.

Live demo: https://mesop-y677hytkra-uc.a.run.app/

- Most of the examples here are forked from existing examples, but cleaned up to make it more user-friendly.
- Although it's somewhat duplicative to have these examples, I think it's somewhat better to have separate examples for documentation purposes vs. testing purposes as the latter may test edge cases whereas the documentation is focused on the common use cases.
- These new examples are type-checked but not tested.

Addresses part of #51. 

Reminder to self: update markdown to link to these demos.


richard-to (approved):
LGTM - Looks great. I definitely like having usable examples along with the code. That was definitely one downside of the current docs where there's the code example but no example of what it looks like.

wwwillchen:
Thanks, I've added a demo for each component, mostly based on our current examples but cleaned up for readability. I think if we have a few more real-world examples that will be great.

PR #157, opened 2024-04-25
Create optimized / prod-Angular app for prod mode

### LINKED ISSUE #61: Compile Mesop Angular app for pip release
Right now, we compile the development version of Angular. We should use the app_bundle target similar to what [tensorboard](https://github.com/tensorflow/tensorboard/blob/d73e21318ee2a30e3cf1fa3ae563d6fd70b1f88c/tensorboard/defs/defs.bzl#L101) has done.

This will make the JS binary smaller and faster.

wwwillchen:
Fixes #61.

TLDR: compiling Angular app with Bazel is painful, especially with protos.

Previously, even in prod mode, the Mesop app was compiled in Angular dev mode, which is not ideal, particularly for performance reasons (Angular prod mode is AOT compiled instead of dev mode which is JIT compiled).

Because there isn't a well-lit path for compiling Angular apps in Bazel, we end up copying what tensorboard has done: https://github.com/tensorflow/tensorboard/blob/4023658707d54976622c3b3a32f5a4a215689da7/tensorboard/defs/defs.bzl#L93 which is essentially to use Angular internal build toolchain for OSS.

However, we couldn't use Angular's internal `app_bundle` Bazel def directly because Mesop needs a custom esbuild plugin to resolve the generated JS paths for protos, so we ended up forking Angular's app_bundle Bazel def: https://github.com/angular/dev-infra/tree/301473cb5fbcd6cca8b70a6369dcdf3dc2d12e08/bazel/app-bundling

Note: this will require some downstream changes: 1) needs a `ng_js_binary_prod` shim in build_defs/ and 2) update the copybara script for mesop/web/src/app/prod/index.html because "prod_bundle/bundle.js" is now "prod_bundle.js" (to be fair, the previous path was kind of funky).

### DIFF
--- a/build_defs/app_bundle/BUILD
+++ b/build_defs/app_bundle/BUILD
@@ -1,0 +1,12 @@
+package(default_visibility = ["//visibility:public"])
+
+exports_files([
+    "terser_config.json",
+    "esbuild.config-tmpl.mjs",
+])
+
+# Make source files available for distribution via pkg_npm
+filegroup(
+    name = "files",
+    srcs = glob(["*"]),
+)
--- a/build_defs/app_bundle/README.md
+++ b/build_defs/app_bundle/README.md
@@ -1,0 +1,1 @@
+Forked from https://github.com/angular/dev-infra/tree/301473cb5fbcd6cca8b70a6369dcdf3dc2d12e08/bazel/app-bundling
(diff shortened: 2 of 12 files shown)


richard-to (approved):
Can't say I fully understand everything that's going on in the prod build process here. But was definitely educational to read through the changes here. Sounds like a lot of fun getting this to work for Mesop.

wwwillchen:
> Can't say I fully understand everything that's going on in the prod build process here. But was definitely educational to read through the changes here. Sounds like a lot of fun getting this to work for Mesop.

Yeah, there's quite a few pieces that are unfortunately not very straightforward. I'll try to explain to you the build process in-depth sometime (both for OSS and downstream).

Issue #84, opened 2024-03-22
Improve Mesop performance for apps with lots of content


wwwillchen:
People are sometimes building apps with Mesop which sends down lots of content (e.g. hundreds of kbs). Because Mesop is naively sending the entire component tree with every user interaction, this can result in a slow app experience.

There's a couple ideas for improving Mesop performance:

- #62 
- #63

wwwillchen:
Overall performance is quite good now with the component diffing + gzipping #231

Issue #62, opened 2024-02-24
Send proto diffs from server to client


wwwillchen:
Diff:

- Component tree
- States

Common case:

- In text_io, it's common to append to the output string and being able to diff even a single field would be helpful.

richard-to:
I was thinking about this issue some more. Are requests to from client to server handled synchronously? Or is async?

I'm assuming the events are sent async, so there's no queuing of events to ensure event 1 is processed and then event 2 is processed.

That could pose an issue for the proto diffs. But I guess this may be an existing behavior even with the full request being sent back. Not sure though, I'd have to test out the behavior a bit.

wwwillchen:
I think TCP guarantees that the events are processed in order: https://hpbn.co/building-blocks-of-tcp/#head-of-line-blocking

fwiw, I've not seen any issue with out of ordering

richard-to:
I like your suggestion earlier about sending a hashes on requests. Since I guess the neither the component tree nor the state gets updated client side.

It does seem like in that case, we could store state/component-tree in memory and not need to worry about server restarts or server affinity.

And I guess we'd need some kind of retry behavior in cases where the state is lost on the server.

So given that, it does seem like we'll need the state diffing behavior.

wwwillchen:
@richard-to should we close this ticket? seems like everything is done now.

richard-to:
Hah, sure. Yeah, I think this can be closed. I'll just create other issues for related improvements and/or fixes.

PR #167, opened 2024-04-28
Add component diffing on event responses.

richard-to:
This change is gated by a flag called `enable_component_tree_diffs`. When enabled, the full component tree will not be returned in event responses. Instead only the changed component fields will be returned to the frontend. The frontend will then apply the diff to update its copy of the component tree.

A few notes:

- Wasn't sure how to add unit tests for typescript. Is there a framework we're using?
- I'd like to run the playwright tests with the diff flag enabled and without it enabled (right now it's enabled by default in the test). Haven't looked into this too much yet. Seems like there's a way to do it 
  - Passing in a different config but that would lead to a lot of duplication (so probably rule this out)
  - Seems like we could pass in some environment variable to enable/disable the flag. Then we'd need to update the CI to run the tests twice.

Ref #62, #84

### DIFF
--- a/mesop/component_helpers/helper.py
+++ b/mesop/component_helpers/helper.py
@@ -326,3 +326,82 @@ runtime().register_event_mapper(
     key=key.key,
   ),
 )
+
+
+_COMPONENT_DIFF_FIELDS = (
+  "key",
+  "source_code_location",
+  "style",
+  "style_debug_json",
+  "type",
+)
+
+
+def diff_component(
+  component1: pb.Component, component2: pb.Component
+) -> pb.ComponentDiff:
+  """Finds diffs between two component trees.
+
+  Args:
+    component1: Old version of the component tree.
+    component2: New version of the component tree that we want component1 to match.
+
+  Returns:
+    Changes needed to make component1 equal to component2.
+  """
+  diff = pb.ComponentDiff()
+
+  # Check each field for differences. For now if there are any differences, we will
+  # mark the entire field for replacement.
+  for field in _COMPONENT_DIFF_FIELDS:
+    if getattr(component1, field) != getattr(component2, field):
+      diff.diff_type = pb.ComponentDiff.DiffType.DIFF_TYPE_UPDATE
+      setattr(
+        diff,
+        f"update_strategy_{field}",
+        pb.ComponentDiff.UpdateStrategy.UPDATE_STRATEGY_REPLACE,
+      )
+      if isinstance(getattr(component2, field), Message):
+        getattr(diff, field).CopyFrom(getattr(component2, field))
+      else:
+        setattr(diff, field, getattr(component2, field))
+
+  # Handle differences with child components.
+  for index, child_component in enumerate(component1.children):
+    if index >= len(component2.children):
+      diff.diff_type = pb.ComponentDiff.DiffType.DIFF_TYPE_UPDATE
+      diff.children.append(
+        pb.ComponentDiff(
+          index=index, diff_type=pb.ComponentDiff.DiffType.DIFF_TYPE_DELETE
+        )
+      )
+    else:
+      child_diff = diff_component(child_component, component2.children[index])
+      if (
+        child_diff
+        and child_diff.diff_type != pb.ComponentDiff.DiffType.DIFF_TYPE_NONE
+      ):
+        diff.diff_type = pb.ComponentDiff.DiffType.DIFF_TYPE_UPDATE
+        child_diff.index = index
+        diff.children.append(child_diff)
+
+  # Find child components that do not exist in component 1. These will need to be added
+  # to component 1.
+  #
+  # It is also very important that the add component diffs occur after the
+  # delete diffs since that additions would mess up the deletions.
+  #
+  # Although in practice child components diffs can contain either add or delete
+  # operations, but not both.
+  for index, component2_child in enumerate(component2.children):
+    if index >= len(component1.children):
+      diff.diff_type = pb.ComponentDiff.DiffType.DIFF_TYPE_UPDATE
+      diff.children.append(
+        pb.ComponentDiff(
+          index=index,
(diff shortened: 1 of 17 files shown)


wwwillchen:
Here's an example for a TS unit test:
https://github.com/google/mesop/blob/a795a41be0e55480b28d9b103eb379f7853c8dce/mesop/web/src/dev_tools/dev_tools_spec.ts#L4
https://github.com/google/mesop/blob/a795a41be0e55480b28d9b103eb379f7853c8dce/mesop/web/src/dev_tools/BUILD#L28-L43

wwwillchen:
Re: running w/ tests on and off - I think the environmental variable idea would work, you can also look at https://playwright.dev/docs/test-parameterize#introduction for some other ideas of how to parameterize test.

Also, I'd recommend adding some e2e tests with reasonably complex trees, multiple levels in the tree, different types of components & mutations (e.g. adding/deleting children) because most of our e2e tests are quite simple (usually 1 component with minimal hierarchy and mutations).

wwwillchen (changes requested):
Left some comments. Overall, this is looking quite promising - great work!

mesop/component_helpers/helper.py
```diff
@@ -326,3 +326,82 @@ def register_event_mapper(
...
+
+  Returns:
+    Changes needed to make component1 equal to component2.
+  """
+  diff = pb.ComponentDiff()
+
+  # Check each field for differences. For now if there are any differences, we will
+  # mark the entire field for replacement.
+  for field in _COMPONENT_DIFF_FIELDS:
+    if getattr(component1, field) != getattr(component2, field):
```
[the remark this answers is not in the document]
richard-to: Based on the behavior I'm seeing, that seems to be the case.

I did a quick search in the public repo and I seem some __eq__ implementations: https://github.com/search?q=repo%3Aprotocolbuffers%2Fprotobuf%20%20__eq__&type=code

mesop/runtime/context.py
```diff
@@ -1,12 +1,23 @@
 from typing import Any, Callable, Generator, TypeVar, cast
 
+from absl import flags
+
 import mesop.protos.ui_pb2 as pb
 from mesop.dataclass_utils import (
   serialize_dataclass,
   update_dataclass_from_json,
 )
 from mesop.exceptions import MesopDeveloperException, MesopException
```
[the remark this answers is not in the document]
richard-to: Ok, I set to it true to by default. One reason I set it to false was mainly auto caution in case there was some edge case bug we didn't catch. But I guess it's probably better to enable it since most people wouldn't know to enable it in the first place.

mesop/server/server.py
```diff
@@ -35,11 +36,18 @@ def render_loop(
     try:
       runtime().run_path(path=path, trace_mode=trace_mode)
       title = runtime().get_path_title(path=path)
```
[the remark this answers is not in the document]
richard-to: Ah good catch. Updated.

mesop/protos/ui.proto
```diff
@@ -364,3 +365,51 @@ message LiteralElement {
...
     optional FieldType type = 1;
 }
+
+// Used for tracking differences between two components.
+//
+// Next ID: 15
+message ComponentDiff {
+    // Used for tracking the index of a child components.
+    // If -1, that means the component is the root component, which will have no parent.
+    optional int32 index = 1 [default = -1];
```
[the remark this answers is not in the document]
richard-to: Good point. I guess the -1 didn't really have an effect anyway. Defaulting to the normal default of 0 works fine.

richard-to:
> Re: running w/ tests on and off - I think the environmental variable idea would work, you can also look at https://playwright.dev/docs/test-parameterize#introduction for some other ideas of how to parameterize test.
> 
> Also, I'd recommend adding some e2e tests with reasonably complex trees, multiple levels in the tree, different types of components & mutations (e.g. adding/deleting children) because most of our e2e tests are quite simple (usually 1 component with minimal hierarchy and mutations).

Thanks for the pointer to the typescript tests. That worked for me.

---------

Haven't got a chance to look into the Playwright flag stuff yet. Also haven't added more complex e2e examples, but agreed that it would be good to have them. Will get to those later this week.

mesop/runtime/context.py
```diff
@@ -62,9 +83,24 @@ def node_slot_children_count(self) -> int | None:
   def set_current_node(self, node: pb.Component) -> None:
     self._current_node = node
 
+  def set_previous_node_from_current_node(self) -> None:
```
wwwillchen, line 86: I don't think this method is being called anywhere? I tried running your PR locally, but I don't actually see any component diffs happening, I think because this method isn't actually being used.

richard-to: Doh, good catch. Looks like a regression when I added the reset_nodes method. It was supposed to call set_previous_node_from_current_node there.

richard-to:
Added an e2e test to better test more complex diffing scenarios.

mesop/runtime/context.py
wwwillchen: I patched your CL but I still don't see any component diffs. If you check the Chrome DevTools console logs, the protos received from the server are logged there. I don't see any component diffs in the proto.

richard-to: Ok, I think I got it working now 🤞🏾. Turns I didn't catch it because I was testing on the chat app which does the streaming response. But with the non-streaming response events it was not returning a diff. So basically the diff was not working on the first loop since previous node was not being set.

So that's the one thing I haven't figured out yet. I had to add back the original trace_mode loop code rather than existing early. However, I'm not sure why that works. I'm probably missing something pretty subtle.

wwwillchen: SGTM, we can debug this some more later.

wwwillchen (approved):
LGTM - great job!

PR #178, opened 2024-05-03
Use watchdog for a more robust file watching for mesop/bin

### LINKED ISSUE #150: Support hot reloading from multiple files using `mesop` CLI (via pip install)
Use https://github.com/gorakhargosh/watchdog?

wwwillchen:
Closes #150.

Note: this doesn't work with Python relative imports which seems relatively less popular than Python absolute imports.

### DIFF
--- a/mesop/bin/bin.py
+++ b/mesop/bin/bin.py
@@ -3,9 +3,11 @@ import os
 import sys
 import threading
 import time
-from typing import Sequence
+from typing import Sequence, cast
 
 from absl import app, flags
+from watchdog.events import FileSystemEvent, FileSystemEventHandler
+from watchdog.observers import Observer
 
 import mesop.protos.ui_pb2 as pb
 from mesop.cli.execute_module import execute_module, get_module_name_from_path
@@ -16,7 +18,6 @@ from mesop.runtime import (
   reset_runtime,
   runtime,
 )
-from mesop.server.server import is_processing_request
 from mesop.server.wsgi_app import create_app
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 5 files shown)


mesop/bin/bin.py
```diff
@@ -84,37 +85,113 @@ def main(argv: Sequence[str]):
...
+
+  # Initially track all the files on the file system and then rely on watchdog.
+  for root, dirnames, files in os.walk(workspace_dir_path):
+    # Filter out directories starting with "." because they
+    # can be special directories like .venv directories which
+    # can have lots of Python files that are definitely not application Python modules.
+    dirnames[:] = [d for d in dirnames if not d.startswith(".")]
+
+    for file in files:
+      if file.endswith(".py"):
```
richard-to, line 169: [question] How does this handle the `__init__.py` case?

mesop/bin/bin.py
```diff
@@ -84,37 +85,113 @@ def main(argv: Sequence[str]):
...
+    src_path = cast(str, event.src_path)
+    if src_path.endswith(".py"):
+      print(f"Stopped watching deleted Python module: {event.src_path}")
+      remove_app_module(
+        workspace_dir_path=self.workspace_dir_path,
+        app_module_path=event.src_path,
+      )
+
+
+def fs_watcher(absolute_path: str):
```
richard-to, line 154: [question] So absolute path is the full path to the "main.py" file?

richard-to (approved):
Looks good. I think I only had one question about the __init__.py case.

mesop/bin/bin.py
wwwillchen: yup

mesop/bin/bin.py
wwwillchen: great catch - I forgot to handle `__init__.py`

wwwillchen:
Thanks for the review

PR #186, opened 2024-05-04
Don't send component configs in prod mode

### LINKED ISSUE #175: Don't send down componentConfigs in prod mode
https://github.com/google/mesop/blob/80fe54643b7ebf204de0a6670d4b20c9f892ae56/mesop/server/server.py#L45

wwwillchen:
Closes #175 

Component configs is only needed for editor mode.

### DIFF
--- a/mesop/cli/cli.py
+++ b/mesop/cli/cli.py
@@ -65,7 +65,7 @@ def monitor_stdin():
 
 
 def main(argv: Sequence[str]):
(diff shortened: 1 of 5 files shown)


richard-to (approved):
LGTM

wwwillchen:
I changed it so it defaults to prod_mode = True, so we don't have to update the downstream sync.

Issue #159, opened 2024-04-26
Add Sticky Columns to Table Component


richard-to:
This is a feature in the Angular Table Component but has not been enabled yet in Mesop.

See "Table with sticky columns" here: https://material.angular.io/components/table/examples

PR #194, opened 2024-05-04
Add sticky headers and columns for Table component

richard-to:
Ref #159

### DIFF
--- a/mesop/components/table/table.py
+++ b/mesop/components/table/table.py
@@ -38,11 +38,25 @@ def map_table_click_event(event, key):
 register_event_mapper(TableClickEvent, map_table_click_event)
 
 
+@dataclass(kw_only=True)
+class TableHeader:
+  sticky: bool = False
+
+
+@dataclass(kw_only=True)
+class TableColumn:
+  sticky: bool = False
+
+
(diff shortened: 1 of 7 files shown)


mesop/components/table/table.py
```diff
@@ -38,19 +38,39 @@ def map_table_click_event(event, key):
...
+  columns: dict[str, TableColumn] | None = None,
 ):
   """
   This function creates a table from Pandas data frame
 
   Args:
       data_frame: Pandas data frame.
       on_click: Triggered when a table cell is clicked. The [click event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click_event) is a native browser event.
+      header: Configures table header to be sticky or not.
+      columns: Configures table columns to be sticky or not.
```
wwwillchen, line 68: I would add a comment on what the key for the dict should be (e.g. column name)

wwwillchen (approved):
LGTM - small comment on documenting the API. thx

mesop/components/table/table.py
richard-to: Good call. Updated.

PR #195, opened 2024-05-05
Remove columns from mesop.labs.layout

wwwillchen:
This was an experimentation from a while back but I don't think this is the right way to support columns because it's usually not what people expect. This basically splits a big body of text across multiple columns, but usually I think people want a grid-like layout (even a single row) where each column is a body of text.


richard-to (commented):
LGTM

PR #196, opened 2024-05-05
Add LLM Rewriter example.

richard-to:
This extends the chatbot example to include the ability to rewrite LLM responses.

# Notes

I'm not too familiar with is needed for an LLM rewriter. But here's my initial attempt. Initially planning to do a side by side preview like in some of the screenshots I saw.

But I felt like a modal was slightly better since it's a bit tougher to make the side by side work well in a responsive fashion. But also a modal is a common UI use case (I'll probably spin off a separate modal example).

# Features

- Edit icon
- Click bot response to edit
- Modal with rewrite UI
- Side by side comparison of original message and rewritten message
- Markdown preview edits
- Updated message bubble contains indicators that its edited
  - Chat bubble turns purple
  - "(edited)" is appended to the bot name

# Unedited 
Screenshot 2024-05-05 at 4 49 59 PM

# Modal

Screenshot 2024-05-05 at 4 57 26 PM

# Edited

Screenshot 2024-05-05 at 4 51 08 PM

### DIFF
--- a/mesop/examples/llm_rewriter.py
+++ b/mesop/examples/llm_rewriter.py
@@ -1,0 +1,400 @@
+import random
+import time
+from dataclasses import dataclass
+from typing import Literal
+
+import mesop as me
+
+Role = Literal["user", "assistant"]
+
+
+@dataclass(kw_only=True)
+class ChatMessage:
+  """Chat message metadata."""
+
+  role: Role = "user"
+  content: str = ""
+  edited: bool = False
+
+
+@me.stateclass
+class State:
+  input: str
+  output: list[ChatMessage]
+  in_progress: bool
+  rewrite: str
+  rewrite_message_index: int
+  preview_rewrite: str
+  preview_original: str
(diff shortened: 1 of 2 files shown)


mesop/examples/llm_rewriter.py
```diff
@@ -0,0 +1,404 @@
...
+            type="flat",
+            disabled=state.in_progress,
+            on_click=on_click_submit_chat_msg,
+          )
+
+
+# Event Handlers
+
+
+def on_input_update(State):
```
wwwillchen, line 112: even though this works, imo, the code is harder to understand vs the naive approach of separate handlers, if someone wants to fork this example

wwwillchen (approved):
lgtm % 1 suggestion

mesop/examples/llm_rewriter.py
richard-to: Great point. Updated

PR #202, opened 2024-05-09
Do not show error box once error is no longer current

### LINKED ISSUE #201: Close error box once error is no longer valid

wwwillchen:
Closes #201.

The previous behavior was overly complicated in trying to show multiple errors, but it resulted in an annoying developer experience where you needed to manually close the red error box (or reload the page) once you fixed the error.

In the vast majority of cases, you're dealing with one error at a time, and if you really have multiple errors at once (e.g. JS errors), then you can still debug them one at a time with the error box still


richard-to (approved):
LGTM - Yeah I didn't even realize you had to close out the error box or that it could show multiple errors.

PR #210, opened 2024-05-09
Create code component

### LINKED ISSUE #96: Provide syntax highlighting in markdown code blocks
Potential API:

```py
me.code(language="python")
```

Prior Art:

- https://docs.streamlit.io/library/api-reference/text/st.code

wwwillchen:
Closes #96

### DIFF
--- a/mesop/components/code/code.py
+++ b/mesop/components/code/code.py
@@ -1,0 +1,10 @@
+from mesop.component_helpers import component
+from mesop.components.markdown.markdown import markdown
+
+
+@component
+def code(code: str = "", *, language: str = "python"):
+  """
+  Creates a code component which displays code with syntax highlighting.
+  """
+  markdown("``` " + language + "\n" + code + "\n" + "```")
--- a/docs/components/code.md
+++ b/docs/components/code.md
@@ -1,0 +1,13 @@
+## Overview
+
+Code is used to render code with syntax highlighting. `code` is a simple wrapper around [markdown](./markdown.md).
+
+## Examples
+
+```python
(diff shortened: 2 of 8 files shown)


mesop/components/code/code.py
```diff
@@ -0,0 +1,10 @@
+from mesop.component_helpers import component
+from mesop.components.markdown.markdown import markdown
+
+
+@component
+def code(code: str = "", *, language: str = "python"):
+  """
+  Creates a code component which displays code with syntax highlighting.
+  """
+  markdown("``` " + language + "\n" + code + "\n" + "```")
```
richard-to, line 10: Seems like it could be possible for someone to enter something that could break this code block if the user was accepting user input for the code block params.

For example if I did a language of ```python```.

But seems like it should be ok since it would still go through the markdown component, I think some kind of injection attack would not work. The code block would just not render correctly.

richard-to (approved):
LGTM - Nice. Definitely a welcome feature to be able to render code with syntax highlighting (especially for the Mesop demo apps).

mesop/components/code/code.py
wwwillchen: Good point, I agree that this isn't a security issue since we're relying on Angular's DOM sanitization https://angular.io/guide/security#sanitization-and-security-contexts, which is done in the markdown component.

wwwillchen:
> LGTM - Nice. Definitely a welcome feature to be able to render code with syntax highlighting (especially for the Mesop demo apps).

yup, that's why I started working on it now :)

PR #213, opened 2024-05-09
Deprecate text_io and introduce text_to_text as the replacement

### LINKED ISSUE #205: Deprecate `mel.text_io` and introduce `mel.text_to_text`
- Basically the same component, except naming change and API clean-up, where text_to_text will have `generator_mode="append"` to align with the behavior in chat.

wwwillchen:
Fixes #205.

- The `text_io` name isn't consistent and as clear with our other labs component, `text_to_image`. This keeps `text_io` as a deprecated API.
- The behavior in `text_io` with appending outputs from a generator returned by a transform function is inconsistent with our `chat` component which appends the outputs together, so I've aligned the new behavior in `text_to_text` and made it configurable. Defaulting to `append` is also better because it can be more performant (e.g. in theory, don't need to do string diffing to figure out the string delta to send to the client).

Reminder to self:

- Need to update the demo links which point to "/text_io"


richard-to (approved):
LGTM

PR #237, opened 2024-05-11
Add `html` component

### LINKED ISSUE #156: Add a component to directly embed html
There's two parts:

- [x] Embedding sanitized HTML (strips out JS)
- [ ] Embedding unsanitized HTML (includes JS) in a sandbox (iframe)

The complexity with unsanitized HTML is that Trusted Types blocks it, which means we need to either use Angular's Trusted Type policy or introduce our own Trusted Type policy. The issue is how this will work downstream (we may need to use the downstream Trusted Type policy).

wwwillchen:
This allows you to render sanitized HTML, however it doesn't allow potentially dangerous HTML like JavaScript.

In a separate PR, I'll update the `embed` component so you can essentially do an `<iframe srcdoc="$htmlcontent">`, which will allow you to run arbitrary JS in a sandboxed way.

Partially addresses #156.

### DIFF
--- a/docs/components/html.md
+++ b/docs/components/html.md
@@ -1,0 +1,17 @@
+## Overview
+
+The HTML component allows you to add custom HTML to your Mesop app.
+
+> Note: the HTML is [sanitized by Angular](https://angular.dev/best-practices/security#sanitization-example) for web security reasons so potentially unsafe code like JavaScript is removed.
+
+## Examples
+
+<iframe class="component-demo" src="https://mesop-y677hytkra-uc.a.run.app/html"></iframe>
+
+```python
+--8<-- "demo/html_demo.py"
+```
+
+## API
+
+::: mesop.components.html.html.html
--- a/mesop/components/html/html.ng.html
+++ b/mesop/components/html/html.ng.html
@@ -1,0 +1,1 @@
+<div [innerHTML]="config().getHtml()" [style]="getStyle()"></div>
--- a/mesop/components/html/html.ts
+++ b/mesop/components/html/html.ts
@@ -1,0 +1,34 @@
+import {Component, Input} from '@angular/core';
+import {
+  Key,
+  Style,
+  Type,
+} from 'mesop/mesop/protos/ui_jspb_proto_pb/mesop/protos/ui_pb';
+import {HtmlType} from 'mesop/mesop/components/html/html_jspb_proto_pb/mesop/components/html/html_pb';
+import {formatStyle} from '../../web/src/utils/styles';
+
+@Component({
+  selector: 'mesop-html',
+  templateUrl: 'html.ng.html',
+  standalone: true,
+})
+export class HtmlComponent {
+  @Input({required: true}) type!: Type;
+  @Input() key!: Key;
+  @Input() style!: Style;
+  private _config!: HtmlType;
+
+  ngOnChanges() {
+    this._config = HtmlType.deserializeBinary(
+      this.type.getValue() as unknown as Uint8Array,
+    );
+  }
+
+  config(): HtmlType {
+    return this._config;
+  }
+
+  getStyle(): string {
+    return formatStyle(this.style);
+  }
+}
--- a/demo/html_demo.py
+++ b/demo/html_demo.py
@@ -1,0 +1,11 @@
+import mesop as me
+
+
+@me.page(path="/html_demo")
+def app():
+  me.html(
+    """
+Custom HTML
+<a href="https://google.github.io/mesop/" target="_blank">mesop</a>
+"""
+  )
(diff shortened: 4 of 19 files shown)


docs/components/html.md
```diff
@@ -0,0 +1,17 @@
+## Overview
+
+The HTML component allows you to add custom HTML to your Mesop app.
+
+> Note: the HTML is rendered inside an iframe for web security reasons so it cannot interact with other Mesop components.
+
+## Examples
+
+<iframe class="component-demo" src="https://mesop-y677hytkra-uc.a.run.app/link"></iframe>
```
richard-to, line 9: [nit] Looks like these are pointing to "link" when it should be html?

richard-to:
> 1. There's no styling inside the iframe. Should we inject the stylesheet in here?

What do you mean by "inject the stylesheet"?

When using the iframe with source doc, is the user able to reference assets (Javascript, CSS) from other domains or would they need to reference from same domain? Or is that not allowed for security reasons?

I guess if linking to external resources is not allowed, then one could write the js/css in the html head. Probably not that user friendly, but since this is an advanced use case, probably ok. Although would be hard to manage.

> 2. The iframe height is arbitrary (browser default). Should we try to automatically resize the iframe based on the content or should Mesop app developers set a height?

I think the app developer should able to set the height. So if there is auto resize as a default and the user could resize as an override, that may be more convenient.

richard-to (approved):
LGTM

wwwillchen:
will re-open once I've been able to address the comments.

wwwillchen:
@richard-to could you take another look? I updated this PR quite a bit and switched to sanitizing the HTML instead of iframing the HTML (I'll try this approach in a separate https://github.com/google/mesop/pull/405). This is better for some use cases because it gets rendered in the same frame, but it doesn't support JS so it's not that flexible.

mesop/components/html/html.ts
```diff
@@ -0,0 +1,51 @@
...
+  @Input() style!: Style;
+  @ViewChild('iframe', {read: ElementRef}) iframe!: ElementRef;
+  private _config!: HtmlType;
+  private srcDoc!: string;
+
+  ngOnChanges() {
+    this._config = HtmlType.deserializeBinary(
+      this.type.getValue() as unknown as Uint8Array,
+    );
+    // const previousSrcDoc = this.srcDoc;
```
richard-to, line 27: Can this commented out code be removed?

mesop/components/html/html.ts
```diff
@@ -0,0 +1,51 @@
...
+    //   this.srcDoc !== previousSrcDoc &&
+    //   this.iframe &&
+    //   this.iframe.nativeElement
+    // ) {
+    //   this.loadIframe();
+    // }
+  }
+
+  ngAfterViewInit() {
+    // this.loadIframe();
```
richard-to, line 41: Ditto here

mesop/web/src/safe_iframe/safe_iframe.ts
```diff
@@ -30,4 +36,24 @@ function setIframeSrcImpl(iframe: HTMLIFrameElement, src: string) {
...
+  iframe.sandbox.add(
+    'allow-same-origin',
+    'allow-scripts',
+    'allow-forms',
+    'allow-popups',
+    'allow-popups-to-escape-sandbox',
+    'allow-storage-access-by-user-activation',
+  );
+
+  // Check if there's any santiziation that's needed.
```
richard-to, line 56: nit: typo sanitization

mesop/components/html/html.ng.html
```diff
@@ -0,0 +1 @@
+<div [innerHTML]="config().getHtml()" [style]="getStyle()"></div>
```
richard-to, line 1: So this is no longer rendered in an iframe? I see some iframe changes in this PR still. How are those related?

richard-to (approved):
LGTM - Just to make sure I understand. The HTML component here renders HTML directly without an iframe. But it is sanitized so no javascript and other malicious stuff can be added?

mesop/components/html/html.ts
wwwillchen: oops, I was too hasty - thanks, done.

mesop/web/src/safe_iframe/safe_iframe.ts
wwwillchen: (I'll add this file in #405

mesop/components/html/html.ng.html
wwwillchen: Yeah, no longer rendered in an iframe. The iframe changes were left over by accident - I've removed them.

wwwillchen:
> LGTM - Just to make sure I understand. The HTML component here renders HTML directly without an iframe. But it is sanitized so no javascript and other malicious stuff can be added?

Yup, that's right. Right now we've been telling people to use me.markdown as a way to render sanitized HTML, which is 1) kind of confusing (since it's not obvious to most developers) and 2) kind of annoying since markdown adds a `<p>` which adds margin, and having a raw div makes it more straightforward to style.

I think we still need JS for a bunch of use cases, so we'll need to handle that separately.

PR #242, opened 2024-05-12
Add LLM Playground example

richard-to:
Here's the LLM Playground example. I based it roughly off of the GCP Vertex AI UI (only added some of the fields they had). I also looked briefly at the Chat GPT one.

# Main Page

Screenshot 2024-05-11 at 5 00 56 PM

# Code View

Screenshot 2024-05-11 at 7 24 40 PM

### DIFF
--- a/demo/llm_playground.py
+++ b/demo/llm_playground.py
@@ -1,0 +1,504 @@
+import random
+import time
+from typing import Callable
+
+import mesop as me
+
+
+@me.stateclass
+class State:
+  title: str = "LLM Playground"
+  # Prompt / Response
+  input: str
+  response: str
+  # Tab open/close
+  prompt_tab: bool = True
+  response_tab: bool = True
+  # Model configs
+  selected_model: str = "gemini-1.5"
+  selected_region: str = "us-east4"
+  temperature: float = 0
+  token_limit: int = 1
+  stop_sequence: str = ""
+  stop_sequences: list[str]
+  # Modal
+  modal_open: bool = False
+  # Workaround for clearing inputs
+  clear_prompt_count: int = 0
+  clear_sequence_count: int = 0
+
+
+@me.page(path="/llm_playground", title="LLM Playground")
+def page():
+  state = me.state(State)
+
+  # Modal
+  with modal(modal_open=state.modal_open):
+    me.text("Get code", type="headline-5")
+    if "gemini" in state.selected_model:
+      me.text(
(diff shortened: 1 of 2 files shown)


mesop/examples/llm_playground.py
```diff
@@ -0,0 +1,504 @@
...
+        icon="code", tooltip="Code", label="CODE", on_click=on_click_show_code
+      )
+
+    # Main Content
+    with me.box(style=_STYLE_MAIN_COLUMN):
+      # Prompt Tab
+      with tab_box(header="Prompt", key="prompt_tab"):
+        me.textarea(
+          label="Write your prompt here, insert media and then click Submit",
+          # Workaround: update key to clear input.
```
wwwillchen, line 92: As an aside (not blocking this PR), I think we should brainstorm a proper API for clearing inputs. It's quite tricky to figure this out and it's a relatively common use case.

One thing we could do is providing a command API (which you can call inside an event handler), e.g. 

```py
me.clear_input(key="prompt")
```

This would work well if the key is globally unique. If the key is not... then we'd probably fire a warning and clear either all, one or none of them.

wwwillchen (approved):
LGTM - this is great.

Same as other PR, let's put in demo/ to make it easier to deploy.

Also, I think this will be a good one to turn into a Colab notebook because then people can call the real API with their API key using Colab user secrets:

https://drlee.io/how-to-use-secrets-in-google-colab-for-api-key-protection-a-guide-for-openai-huggingface-and-c1ec9e1277e0

richard-to:
I didn't realize the AI Studio UI was so much different from the GCP Vertex AI UI. I feel like the AI Studio UI has a lot more functionality. I may try adding different version based on AI Studio UI later. I kind of like the structured prompt ui that the AI Studio UI has. Would be interesting trying to replicate that. But I think this is reasonable enough for now. Yeah, I'll add a version that works with an external colab.

wwwillchen:
> I didn't realize the AI Studio UI was so much different from the GCP Vertex AI UI. I feel like the AI Studio UI has a lot more functionality. I may try adding different version based on AI Studio UI later. I kind of like the structured prompt ui that the AI Studio UI has. Would be interesting trying to replicate that. But I think this is reasonable enough for now. Yeah, I'll add a version that works with an external colab.

👍 - I've never used the Vertex AI UI but I do like the AI studio one, and I think the AI studio is quite popular because it's easy to get started with.

PR #243, opened 2024-05-12
Add markdown editor example

richard-to:
While working on the LLM Rewriter, it occurred to me it would be pretty easy to make a markdown editor using Mesop. Thought it would make a decent example in the vein of the typical todo list type example.

Screenshot 2024-05-08 at 9 20 15 PM

### DIFF
--- a/demo/main.py
+++ b/demo/main.py
@@ -33,6 +33,7 @@ import image as image
 import input as input
 import llm_rewriter as llm_rewriter
 import markdown_demo as markdown_demo  # cannot call it markdown due to python library naming conflict
+import markdown_editor as markdown_editor
 import plot as plot
 import progress_bar as progress_bar
 import progress_spinner as progress_spinner
@@ -75,6 +76,7 @@ FIRST_SECTIONS = [
     name="Use cases",
     examples=[
       Example(name="llm_rewriter"),
+      Example(name="markdown_editor"),
     ],
   ),
 ]
--- a/demo/markdown_editor.py
+++ b/demo/markdown_editor.py
@@ -1,0 +1,190 @@
+from dataclasses import dataclass
+
+import mesop as me
+
+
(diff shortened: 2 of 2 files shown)


wwwillchen (approved):
Neat! Only suggestion is to move this to the demo/ folder so it's easier to deploy on our new demo playground site (& you can still run it locally)

richard-to:
Ok, added it to the demo folder. Wasn't sure if I did it right though.

mesop/examples/__init__.py
```diff
@@ -1,4 +1,5 @@
 from demo import main as main
+from demo import markdown_editor as markdown_editor
```
wwwillchen, line 2: you don't need to import it here

wwwillchen: Instead import it in https://github.com/google/mesop/blob/d132af54d52d7155fca569aa48ec64c3f2968e81/demo/main.py#L22

And then add it as an Example inside one of the Section, e.g. https://github.com/google/mesop/blob/d132af54d52d7155fca569aa48ec64c3f2968e81/demo/main.py#L77 and then it'll show up in the side menu. The embed should work then.

wwwillchen (commented):
added a comment on how to add it to the demo playground

mesop/examples/__init__.py
richard-to: Thanks. That worked.

A few side notes:

- For the markdown editor, the UI gets a bit squished in the demo preview view on a laptop. Could be nice to have a way to hide the code view. Or do a "full screen" embed view.
- The hot reloading doesn't seem to track the changes in demo folder. I'm forced to stop and start the CLI. I think execute_module.py would need updates to handle it. Although I think maybe you have related hot reloading bug already in Github for this. Haven't checked.

PR #254, opened 2024-05-16
Make labs.chat text easier to read

wwwillchen:
I find it somewhat hard to read the chat text - WDYT about adjusting it higher? (We could also make it configurable via a parameter, but generally I think chat UIs are around this font size).

**Before:**

Screenshot 2024-05-15 at 7 30 57 PM

**After:**

Screenshot 2024-05-15 at 7 29 25 PM


richard-to:
Yeah, in general I prefer bigger text myself. I think your adjustment looks good. Can't really remember why I chose that size in the first place.

PR #255, opened 2024-05-16
Fix hot reloading for demo when using Bazel CLI

### LINKED ISSUE #248: Hot reloading not working for files in demo folder
If I make changes to a file a in the demo folder, the Mesop app breaks and I need to restart the CLI again.

I think the issue is because the demo folder isn't really covered in execute_module.py since it's not in the mesop directory. Haven't looked at it too much more yet.

It seems like it's not able to load any pages: `User Error: Accessed path: /embed/llm_playground not registered`

wwwillchen:
Fixes #248


mesop/cli/cli.py
```diff
@@ -71,6 +85,12 @@ def main(argv: Sequence[str]):
   if not FLAGS.prod:
     enable_debug_mode()
 
+  if FLAGS.reload_demo_modules:
```
richard-to, line 88: [nit] May want to add a comment on why we need to do this for demo modules since I don't think it would be intuitive to people new to the code.

richard-to (approved):

mesop/cli/cli.py
wwwillchen: added a comment.

PR #256, opened 2024-05-16
Allow user to toggle preview fullscreen

### LINKED ISSUE #251: Preview pane in Demo App can be a bit too thin for more involved examples
Specifically markdown editor and LLM Playground look a bit too squished in preview mode, especially the latter.

A couple options:

1. Add way to minimize/maximize the code example pane
2. Add a button to expand the embed preview into a full width modal pane
3. Could also add link to the preview outside of the demo app index

For simplicity, probably option 1 seems sufficient.

I think option 3 may not work for some of the current use cases of the demo app where embed it. Option 2 seems a bit weird from a UX perspective (but I guess it'd work similar to say full screen view of a YouTube video UX pattern).

# Example screenshots:

Screenshot 2024-05-14 at 8 29 33 PM

Screenshot 2024-05-14 at 8 29 40 PM

wwwillchen:
Fixes #251


demo/main.py
```diff
@@ -188,20 +190,41 @@ def body(current_demo: str):
...
+        align_items="center",
+        border=me.Border(bottom=BORDER_SIDE),
+      )
+    ):
+      me.text(
+        "Preview",
+        style=me.Style(
+          font_weight=500,
+          padding=me.Padding.all(8),
+          border=me.Border(),
```
richard-to, line 221: [question] What does the empty Border() class do?

demo/main.py
```diff
@@ -188,20 +190,41 @@ def body(current_demo: str):
...
+        ),
+      )
+      with me.content_button(type="icon", on_click=toggle_fullscreen):
+        me.icon(
```
richard-to, line 225: [nit] May want to add a tooltip with the icon. I think it's a pretty common enough icon that may be not necessary, but it's a bit more user friendly.

In terms of terminology, I think "fullscreen" would be slightly misleading from a user since this just hides the code sidebar. In YouTube, I guess this would be similar to "theater mode". But probably wouldn't work in our case. So that's maybe one reason not to include a tooltip.

richard-to (approved):

PR #261, opened 2024-05-16
Implement gzipping for static file serving

### LINKED ISSUE #231: Serve static assets (e.g. JS, CSS) using gzip
in particular, serving the JS file can get a ~4x compression by gzipping:
https://stackoverflow.com/a/59635292

wwwillchen:
- Fixes #231 
- This produces a significant size savings, particularly for the JS bundle (~4x: 1.6MB -> 400kb for prod mode).
- Although ideally gzipping is handled outside of the Flask server and at another layer (e.g. a web server like NGINX),  there's benefits to doing it in the Flask app because in development use case, particularly with Colab, it can be quite slow to load the JS bundle.
- To avoid paying the cost of gzipping the same files over and over again (~30ms on my local computer), this implements a singleton cache. This is OK assuming that the file contents of what we're caching do not change (for these static files, this should hold true).

### DIFF
--- a/mesop/server/BUILD
+++ b/mesop/server/BUILD
@@ -24,3 +24,9 @@ py_test(
     srcs = ["wsgi_app_test.py"],
     deps = [":server"] + THIRD_PARTY_PY_PYTEST,
 )
+
+py_test(
+    name = "static_file_serving_test",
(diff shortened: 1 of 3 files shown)


richard-to (approved):
LGTM

PR #264, opened 2024-05-16
Fix code demo rendering.

richard-to:
The code demo was not rendering correctly since there is an existing code module in Python.

Screenshot 2024-05-16 at 10 43 00 AM


wwwillchen (approved):
Thanks!

PR #265, opened 2024-05-16
Fix select demo rendering.

richard-to:
The select demo was not rendering correctly since there is an existing module in Python.

Screenshot 2024-05-16 at 10 50 54 AM


wwwillchen (approved):
Thanks!

PR #267, opened 2024-05-17
Remove click event on modal bg for LLM Playground.

richard-to:
The click event will be triggered on the nested elements, so this prevents copy and pasting of the code.

For now just using a click event on the close button.

Also logged an issue here: https://github.com/google/mesop/issues/268

### DIFF
--- a/demo/llm_playground.py
+++ b/demo/llm_playground.py
@@ -64,7 +64,7 @@ def page():
             temperature=state.temperature,
           )
         )
-    me.button(label="Close", type="raised", on_click=on_click_modal_bg)
+    me.button(label="Close", type="raised", on_click=on_click_modal)
 
   # Main content
   with me.box(style=_STYLE_CONTAINER):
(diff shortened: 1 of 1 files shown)


demo/llm_playground.py
```diff
@@ -137,11 +137,7 @@ def page():
       with me.box(style=_STYLE_SLIDER_INPUT_BOX):
         with me.box(style=_STYLE_SLIDER_WRAP):
           me.slider(
-            min=0,
-            max=2,
-            step=0.1,
-            style=_STYLE_SLIDER,
```
wwwillchen, line 143: I think you need to rebase?

wwwillchen (approved):

demo/llm_playground.py
richard-to: Not related to this scenario. But I think I need to fix my rebase settings. Keep running into issues when I try to rebase upstream main to update a feature branch.

PR #270, opened 2024-05-17
Setup reasonable default web security policies

### LINKED ISSUE #225: Configure CSP and Trusted Types
https://angular.io/guide/security#content-security-policy

- Flask
- https://github.com/GoogleCloudPlatform/flask-talisman

wwwillchen:
Fixes #225. I opened up #271 because it's the remaining major web security piece we should shore up.

Notes:
- CSP is quite easy to mess up due to weird syntax it uses, so I've done a snapshot test to avoid accidental regressions.
- Using CSP nonce requires us to generate a random string and set it on the server ([ref](https://angular.io/guide/security#content-security-policy:~:text=Set%20the%20ngCspNonce%20attribute%20on%20the%20root%20application%20element%20as%20%3Capp%20ngCspNonce%3D%22randomNonceGoesHere%22%3E%3C/app%3E.%20Use%20this%20approach%20if%20you%20have%20access%20to%20server%2Dside%20templating%20that%20can%20add%20the%20nonce%20both%20to%20the%20header%20and%20the%20index.html%20when%20constructing%20the%20response.))

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,9 @@ repos:
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
+        # do not fix snapshot files because these are golden
+        # files which need to match the test output exactly.
+        exclude: /snapshots/
       # Suppress check-yaml because of false error:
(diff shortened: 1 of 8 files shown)


richard-to (approved):
LGTM

PR #272, opened 2024-05-17
Upgrade ruff pre-commit hook

wwwillchen:
It was fighting some of the formatting changes I was making (due to a version skew between the version of the Ruff via VS code extension and the pre-commit Ruff)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
       - id: check-case-conflict
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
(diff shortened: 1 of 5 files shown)


richard-to:
This change seems to include some other changes (like some of the security ones and also some link up dates).

wwwillchen:
Fixed the commit history

richard-to (approved):
Looks like a failing test. But doesn't seem related to the changes here.

wwwillchen:
> Looks like a failing test. But doesn't seem related to the changes here.

fixed it - the new version of ruff formats things a bit differently so that's why it was failing.

PR #273, opened 2024-05-17
Stop sending component configs with every render event in Editor mode

### LINKED ISSUE #269: Stop sending component_configs with every render event in editor mode
This adds a significant overhead a few hundred ms for each POST request to /ui and it's not necessary, even in editor mode.

wwwillchen:
Fixes #269.

The previous implementation was naive - sending down all the component configs with every render event which is wasteful and eats >300ms. We can use the component configs from the first render.

The only downside to this change is that if the user adds a new component (e.g. using `@me.component`) it won't send down the new component config, but this isn't that big of a deal because 1) the component will work, it just won't show up in the right-hand devtools/editor panel, and 2) the developer can refresh the page (triggering all the component configs to be resent).

### DIFF
--- a/mesop/server/server.py
+++ b/mesop/server/server.py
@@ -32,7 +32,10 @@ def configure_flask_app(
   flask_app = Flask(__name__)
 
   def render_loop(
-    path: str, keep_alive: bool = False, trace_mode: bool = False
+    path: str,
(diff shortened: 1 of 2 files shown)


richard-to (approved):
LGTM

Issue #95, opened 2024-04-02
Support custom web components


wwwillchen:
_Updated:_ Proposal doc for supporting [custom web components](https://docs.google.com/document/d/1Nc7Ub8DMNSxAmFuPRdyrlZXh_AoxVjZM-YEeWF8dAyI/edit?usp=sharing) - please leave any feedback either on the doc or on this thread. Thanks!

wwwillchen:
Along the lines of the previous comment, it may be worthwhile to expose a lightweight custom native component pathway for people trying to write vanilla JS/HTML/CSS components (i.e. non-Angular components), in addition to another potential pathway to bring in custom Angular components into Mesop.

**API sketch:**

```py
def custom_component(key: str, text: str):
  me.html(f```
<div><span>{text}</span><button>native button</button></div>

<script>
  const key = {key};
  document.querySelector('button').addEventListener((event) => {
    window.mesop.dispatchEvent(new window.mesop.UserEvent({key}));
  });
</script>
```)

`me.html` would be what's proposed in #156  and would end up iframing the html content for security and isolation purposes.

I think supporting `me.html` is a pretty good idea on its own (since we already support `me.embed`), and this would mean adding a communication channel between the iframed HTML and the main Mesop app.

Could this also support regular Angular components? Potentially, if the user injects a `<script src="./custom_angular_component.js"></script>` then they could use it. However, compiling an Angular component as a standalone script doesn't seem trivial, particularly for components part of larger Angular codebases. Angular has a feature called [Angular Elements](https://angular.io/guide/elements) that lets you compile Angular components into native HTML web components, but I'm unsure how widely used this is.

**Open questions:**

- Could this support folks trying to use (currently unsupported) viz libraries like Plotly? (it could result in a pretty large JS file being sent down via Mesop which would be slow, unless they use a CDN)

wwwillchen:
**Prior art:**

Streamlit: 
- https://docs.streamlit.io/develop/concepts/custom-components/intro
- https://github.com/streamlit/component-template
- https://github.com/tconkling/streamlit-discrete-slider

Gradio:

- https://www.gradio.app/guides/custom-components-in-five-minutes

justinfagnani:
BTW, you could just support native web components directly / only and then let Angular (as well as other frameworks) be used via Angular Elements or other web components wrappers / compilers.

There's already a Material 3 web components reference implementation by the Material team here: https://github.com/material-components/material-web

[...]

PR #287, opened 2024-05-19
Setup APIs to allow custom component downstream

wwwillchen:
I think we will need some custom components for our hackathon that aren't yet ready to be public APIs, I will open a draft PR to demonstrate this.

Related to #95

### DIFF
--- a/mesop/component_helpers/helper.py
+++ b/mesop/component_helpers/helper.py
@@ -227,6 +227,21 @@ def insert_composite_component(
   )
 
 
+def insert_custom_component(
+  component_name: str,
+  proto: Message,
+  key: str | None = None,
+  style: Style | None = None,
+):
+  return insert_component(
+    # Prefix with <custom> to ensure there's never any overlap.
+    type_name="<custom>" + component_name,
+    proto=proto,
+    key=key,
+    style=style,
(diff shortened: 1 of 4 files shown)


mesop/component_helpers/helper.py
```diff
@@ -227,6 +227,21 @@ def insert_composite_component(
...
 
+def insert_custom_component(
+  component_name: str,
+  proto: Message,
+  key: str | None = None,
+  style: Style | None = None,
+):
+  return insert_component(
+    # Prefix with <custom> to ensure there's never any overlap.
+    type_name="<custom>" + component_name,
```
richard-to, line 238: Question, so when someone is using the custom component. Say the mic component. Would they still do something like me.mic()?

richard-to (approved):

mesop/component_helpers/helper.py
wwwillchen: No, they would import the mic component from their module, e.g.

```py
from my_package.mic_component import mic
```

Btw, the type_name here is used internally to map between Python and JS, but it doesn't actually need to correspond with the component function name, etc. (we keep it the same for consistency).

Issue #283, opened 2024-05-18
Support hotkey (e.g. enter button)


wwwillchen:

wwwillchen:
Since we have web components, I'd probably rely on users to write custom JS to support hotkeys.

PR #296, opened 2024-05-21
Add an on_enter handler for input type components

orangerd:
Add an `on_enter` handler for `<input type="text">` type components, so that input boxes can be "submitted" using the keyboard. Currently, there's no handler for "Enter", so input boxes are usually "submitted" using a relevant button hooked to a handler.

### DIFF
--- a/mesop/components/input/e2e/input_app.py
+++ b/mesop/components/input/e2e/input_app.py
@@ -11,10 +11,15 @@ def on_input(e: me.InputEvent):
   state.input = e.value
 
 
+def on_enter(e: me.EnterEvent):
+  state = me.state(State)
+  state.input = "boo"
+
+
 @me.page(path="/components/input/e2e/input_app")
 def app():
   s = me.state(State)
-  me.input(label="Basic input", on_input=on_input)
+  me.input(label="Basic input", on_input=on_input, on_enter=on_enter)
   me.text(text=s.input)
 
   me.textarea(
--- a/mesop/components/input/e2e/input_test.ts
+++ b/mesop/components/input/e2e/input_test.ts
@@ -4,4 +4,6 @@ test('test input interactivity works', async ({page}) => {
   await page.goto('/components/input/e2e/input_app');
   await page.getByLabel('Basic input').fill('hi');
   expect(await page.getByText('hi').textContent()).toEqual('hi');
+  await page.getByLabel('Basic input').press('Enter');
+  expect(await page.getByText('boo').textContent()).toEqual('boo');
 });
--- a/mesop/__init__.py
+++ b/mesop/__init__.py
@@ -57,6 +57,7 @@ from mesop.components.divider.divider import divider as divider
 from mesop.components.embed.embed import embed as embed
 from mesop.components.icon.icon import icon as icon
 from mesop.components.image.image import image as image
+from mesop.components.input.input import EnterEvent as EnterEvent
 from mesop.components.input.input import input as input
 from mesop.components.input.input import native_textarea as native_textarea
 from mesop.components.input.input import textarea as textarea
--- a/mesop/components/input/input.ng.html
+++ b/mesop/components/input/input.ng.html
@@ -48,6 +48,7 @@
     [value]="config().getValue()"
     [readonly]="config().getReadonly()"
     (input)="onInput($event)"
+    (keyup)="onKeyUp($event)"
   />
   }
 </mat-form-field>
--- a/mesop/components/input/input.proto
+++ b/mesop/components/input/input.proto
@@ -3,7 +3,7 @@ syntax = "proto2";
 
 package mesop.components.input;
 
-// Next id: 24
+// Next id: 25
 message InputType {
   optional bool disabled = 1;
   optional string id = 2;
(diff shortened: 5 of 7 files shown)


mesop/components/input/e2e/input_app.py
wwwillchen, line 1: Let's add a test case (corresponding to this file):
https://github.com/google/mesop/blob/main/mesop/components/input/e2e/input_test.ts

info on running tests: https://google.github.io/mesop/internal/testing/

wwwillchen (commented):
Overall - looks good, if you could add a test, that would be great. thanks.

mesop/components/input/e2e/input_app.py
orangerd: Thanks Will, added! Couldn't seem to run the playwright tests locally, but I'll try to ask around.

mesop/components/input/e2e/input_test.ts
```diff
@@ -4,4 +4,7 @@ test('test input interactivity works', async ({page}) => {
   await page.goto('/components/input/e2e/input_app');
   await page.getByLabel('Basic input').fill('hi');
   expect(await page.getByText('hi').textContent()).toEqual('hi');
+  await page.getByLabel('Basic input').press('Enter');
+  await page.waitForTimeout(3000);
```
wwwillchen, line 8: is this needed? the next line should be enough

orangerd: Yeah you're right - removed!

wwwillchen:
Sorry for the slower review - I've been re-thinking whether this is the right API.

I think it's useful to have this functionality, but I'm hesitating a little bit with how this will fit in when we support keyboard shortcuts more holistically. For example, we could make this more generic and have this be `on_key_up` which emits a KeyUpEvent that includes the `key` value (e.g. "Enter").

For example, I noticed in [AI studio](https://aistudio.google.com/), they distinguish between "enter" (new line), "shift+enter" (enters the text without running), and "ctrl+enter" (enters the text & runs). I think it could be useful to support this in the future. It's a little weird API wise to have something like `on_enter` and `on_key_up` in the future if we want to support this. 

@richard-to WDYT about the API?

BTW @richard-to came up with a pretty clever workaround of detecting when there's `"\n"` in the `InputEvent` value and then using that to trigger the submit action in our internal app.

richard-to:
> I originally wrote a comment suggesting an alternative approach of `on_key_up` and having a `KeyUpEvent` which would be more generic than `on_enter`, but that becomes very similar to `on_input` and it makes it ambiguous whether on_key_up would be called before on_input, etc. I think keeping this simple as `on_enter` LGTM, eventually we'll want to have some kind of broader hotkey support (#283), but this seems like a good initial step.
> 
> @richard-to WDYT about the API here?

I was thinking that we'd have something more generic and similar to what's available on the Javascript side, which would be on key up.  But that's a great point about on_input.

On key up and on input are not technically the same. It sort of feels like a on compromise of on key up/press/down in some ways.

I do wonder if it would be possible to pass in modifier keys and key presses like enter with on_inputFor example, it's possible we could pass in a list of the modifier keys and key presses. I think that is possible, but due to the necessary debouncing and latency, I'm not sure how easy to use it would be from an API usage standpoint.

So do I think the proposed hot key idea could work more generically.

------

I think the on_enter is still very useful since it is a common use case. So eventually could be seen as a shortcut for whatever hot key implementation.

Maybe one suggestion feature-wise could be support the also common "Modifier" + Enter use case that Will mentioned. But I think that could be added on later to the current api if necessary.

wwwillchen:
> > I originally wrote a comment suggesting an alternative approach of `on_key_up` and having a `KeyUpEvent` which would be more generic than `on_enter`, but that becomes very similar to `on_input` and it makes it ambiguous whether on_key_up would be called before on_input, etc. I think keeping this simple as `on_enter` LGTM, eventually we'll want to have some kind of broader hotkey support (#283), but this seems like a good initial step.
> > @richard-to WDYT about the API here?
> 
> I was thinking that we'd have something more generic and similar to what's available on the Javascript side, which would be on key up. But that's a great point about on_input.
> 
> On key up and on input are not technically the same. It sort of feels like a on compromise of on key up/press/down in some ways.
> 
> I do wonder if it would be possible to pass in modifier keys and key presses like enter with on_input. For example, it's possible we could pass in a list of the modifier keys and key presses. I think that is possible, but due to the necessary debouncing and latency, I'm not sure how easy to use it would be from an API usage standpoint.
> 
> So do I think the proposed hot key idea could work more generically.
> 
> I think the on_enter is still very useful since it is a common use case. So eventually could be seen as a shortcut for whatever hot key implementation.
> 
> Maybe one suggestion feature-wise could be support the also common "Modifier" + Enter use case that Will mentioned. But I think that could be added on later to the current api if necessary.

OK sounds good, let's go with this `on_enter` for now. We could add  the modifier keys (e.g. ctrl, shift) to the EnterEvent to support the advanced use cases I was thinking about, but we can do that in a follow-up PR.

wwwillchen:
@orangerd you'll need to fix a pre-commit issue. BTW you can install the pre-commit hooks to catch it locally, see: https://google.github.io/mesop/internal/development/#commit-hooks

orangerd:
Ah sorry - yes installed the pre-commits

wwwillchen:
Thanks! Merging

PR #306, opened 2024-05-25
Add some new style properties

### LINKED ISSUE #302: Add transform / rotate CSS properties
- https://developer.mozilla.org/en-US/docs/Web/CSS/rotate
- https://developer.mozilla.org/en-US/docs/Web/CSS/transform

### LINKED ISSUE #303: Add justify-self / align-self properties
- https://developer.mozilla.org/en-US/docs/Web/CSS/align-self
- https://developer.mozilla.org/en-US/docs/Web/CSS/justify-self

richard-to:
Properties added:

- rotate
- transform
- justify_self
- align_self

One thing to note here is that I noticed justify_self/justify_items and align_self/align_items do not have the exact same property values. Also there were some property values missing from the ItemAlignmentValues Literal definition. I ended up adding the missing values and creating a separate Literal definition for justify_self/justify_items. I believe this should be ok since it's just a Literal definition, so API should be the same still.

Also added some animation examples using rotate and transform.

- Fixes #302
- Fixes #303

### DIFF
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -22,6 +22,42 @@ ItemAlignmentValues = Literal[
   "center",
   "start",
   "end",
+  "flex-start",
+  "flex-end",
+  "self-start",
+  "self-end",
+  "baseline",
+  "first baseline",
+  "last baseline",
+  "safe center",
+  "unsafe center",
+  "inherit",
+  "initial",
+  "revert",
+  "revert-layer",
+  "unset",
+]
+ItemJustifyValues = Literal[
+  "normal",
+  "stretch",
+  "center",
+  "start",
+  "end",
+  "flex-start",
+  "flex-end",
+  "self-start",
+  "self-end",
+  "left",
+  "right",
+  "baseline",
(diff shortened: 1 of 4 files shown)


mesop/component_helpers/style.py
```diff
@@ -22,6 +22,46 @@
...
+  "self-start",
+  "self-end",
+  "left",
+  "right",
+  "baseline",
+  "first baseline",
+  "last baseline",
+  "safe center",
+  "unsafe center",
+  "legacy left",
```
wwwillchen, line 57: I would remove the legacy values - I can't really think of why someone would use it in Mesop (IIUC, the web has this for legacy compatibility reasons)

wwwillchen (approved):

mesop/component_helpers/style.py
richard-to: Good call. I was waffling on keeping/leaving those. Ended up keeping them, I think you make good points that Mesop users likely shouldn't need to use legacy properties unless they were copying some bad code.

PR #307, opened 2024-05-25
Allow user to set value for slider.

### LINKED ISSUE #241: Can't set starting value for slider component
The slider component has no way to set a starting value. So it always starts on the right end. Users may want a different starting position/value, such as the center.

richard-to:
This change allows the value to be set for the slider. The implemented behavior here is similar to the input/textarea value field.

The value will update the value of the slider only if the value being passed in changes.

This is for the following use cases:

1. Setting an initial value
2. Updating the value programmatically. For example if you wanted a textbox to control the slider value
3. Only updating the value if it changes, allows the slider to move smoothly (same issue as the text inputs)

Fixes #241

### DIFF
--- a/mesop/components/slider/slider.ts
+++ b/mesop/components/slider/slider.ts
@@ -1,4 +1,5 @@
 import {MatSliderModule} from '@angular/material/slider';
+import {FormsModule} from '@angular/forms';
 import {Component, Input} from '@angular/core';
 import {
   UserEvent,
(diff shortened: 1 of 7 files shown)


mesop/components/slider/slider.ts
```diff
@@ -15,14 +16,19 @@ import {formatStyle} from '../../web/src/utils/styles';
...
-  imports: [MatSliderModule],
+  imports: [MatSliderModule, FormsModule],
 })
 export class SliderComponent {
   @Input({required: true}) type!: Type;
   @Input() key!: Key;
   @Input() style!: Style;
+
+  // Need to make this value accessible by ngModel.
+  value?: number = 0;
```
wwwillchen, line 27: I don't think you need the `?` since you're defining a default value

wwwillchen (approved):

mesop/components/slider/slider.ts
richard-to: I mainly just used `?` because the proto fields seems to be defined with `?`. So mainly did it to avoid the casting. However, I updated it now to just as cast to number on the getValue() proto call.

PR #308, opened 2024-05-26
Enable allow-listing parent origins for iframing

### LINKED ISSUE #271: Provide a mechanism for allow-listing origins which can iframe a Mesop app
https://cheatsheetseries.owasp.org/cheatsheets/Clickjacking_Defense_Cheat_Sheet.html

# API

## Option 1 - configure at page-level

```py
@me.page(path="/foo", security_policy=me.SecurityPolicy(allowed_frame_ancestors=["github.com"]))
def app():
   ...
```

This is nice because it's flexible.

## Option 2 - configure at app-level

This could be controlled by a command-line flag or by calling a mesop API at the top-level of a program (e.g. `me.register_allowed_frame_ancestors`).

## Thoughts

I think Option 1 is more flexible and seems more intuitive

# Considerations

## Default

We should eventually have the default be that there's no allowed frame ancestor (except for self, which is always permitted). However, we should have good documentation on setting this up before making this the default behavior.

## Editor mode

For editor mode, we would probably always allow any frame ancestor, regardless of what's specified, because this is required for Mesop to load in Colab which will iframe Mesop in a randomly-generated origin so it's not possible to specify this ahead of time. Given that apps should never be deployed in editor mode, this seems like a reasonable choice.

wwwillchen:
- In editor mode, allow any origin to iframe Mesop (in particular because Colab iframes content like Mesop in a randomly generated origin, but also to make local development easier, where you may want to iframe from another localhost port).
- In prod mode, allow the self origin and any allow-listed origins for that particular page.
- Adds docs

Fixes #271.

### DIFF
--- a/docs/guides/web_security.md
+++ b/docs/guides/web_security.md
@@ -1,0 +1,38 @@
+# Web Security
+
+Mesop by default configures its apps to follow a set of web security best practices.
+
+## How
+
+At a high-level, Mesop is built on top of Angular which provides [built-in security protections](https://angular.io/guide/security) and Mesop configures a strict [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).
+
+Specifics:
+
+- Mesop APIs do not allow arbitrary JavaScript execution in the main execution context. For example, the [markdown](../components/markdown.md) component sanitizes the markdown content and removes active HTML content like JavaScript.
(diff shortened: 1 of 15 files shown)


richard-to (approved):
LGTM - Do you think this change breaks any existing apps downstream?

wwwillchen:
> LGTM - Do you think this change breaks any existing apps downstream?

Good point - I've added a `default_allowed_iframe_parents` parameter for `configure_static_file_serving` so we can configure reasonable defaults downstream (essentially preserving our existing behavior).

Issue #310, opened 2024-05-26
Support XSRF mitigations


wwwillchen:
- Referer checking
- Anti-XSRF tokens

References:

- https://flask-seasurf.readthedocs.io/en/latest/
- https://github.com/GoogleCloudPlatform/flask-talisman
 
Considerations:

- Make sure to configure flask.session cookies to be `secure` and `httponly`
- How do I expire an XSRF token?

wwwillchen:
Implementation:

- Generate a CSRF token and set it on the Flask session
- Emit the CSRF token to Angular (using HTTP header?)
- Have Angular app POST the CSRF token 
- Flask server app checks the CSRF token

Notes:

- Use [Tink-Python](https://github.com/tink-crypto/tink-py) to generate secret key and tokens

```py
from flask import Flask, session

app = Flask(__name__)
app.secret_key = 'your-secret-key'  # Use tink? Or does this need to be same across apps?

@app.before_request
def set_csrf_token():
    if 'csrf_token' not in session:
        session['csrf_token'] = generate_csrf_token()
```

wwwillchen:
https://security.stackexchange.com/questions/158045/is-checking-the-referer-and-origin-headers-enough-to-prevent-csrf-provided-that

wwwillchen:
Right now, the origin check seems to be adequate as it hasn't caused any issues after the initial implementation PRs. Doing CSRF tokens could cause some usability issues as the tokens could expire in the middle of a Mesop user session which would be a bad user experience. Closing this as done.

PR #313, opened 2024-05-27
Protect against CSRF by checking origin of POST request

wwwillchen:
Addresses part of #310.

### DIFF
--- a/mesop/server/server.py
+++ b/mesop/server/server.py
@@ -1,5 +1,6 @@
 import base64
 import time
+import urllib.parse as urlparse
 from typing import Generator, Sequence
 
 from flask import Flask, Response, abort, request, stream_with_context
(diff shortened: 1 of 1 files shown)


mesop/server/server.py
```diff
@@ -156,6 +157,10 @@ def generate_data(ui_request: pb.UiRequest) -> Generator[str, None, None]:
 
   @flask_app.route("/ui", methods=["POST"])
   def ui_stream() -> Response:
+    # Prevent CSRF by checking the request origin matches the origin
+    # of the URL root (where the Flask app is being served from)
+    if not is_same_origin(request.headers.get("Origin"), request.url_root):
```
richard-to, line 162: Seems like browsers prevent modification of Origin. But would I be able to fake the Origin header using something like cURL?

richard-to (approved):

mesop/server/server.py
wwwillchen: I assume it's possible to spoof an Origin header, but that's not really a CSRF vulnerability at that point, because if someone is running cURL (or some other network tool) to spoof the HTTP header, then either the user is trying to intentionally do a cross-site request (which isn't really a security vulnerability, just sort of mis-using a site), or the user has already been seriously compromised (e.g. someone can already do remote code execution on their computer), in which case there's nothing we can do.

PR #314, opened 2024-05-27
Improve our internal documentation

wwwillchen:
Anything else that you think should be refreshed?

### DIFF
--- a/docs/internal/development.md
+++ b/docs/internal/development.md
@@ -1,5 +1,7 @@
 # Development
 
+I recommend following (or at least reading) all the steps in this doc if you plan on actively developing Mesop.
+
 ## Setup
 
 ### Bazel/ibazel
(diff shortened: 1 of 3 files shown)


richard-to (approved):

richard-to:
For the development page, maybe add a line explaining the use cases where you would use Dev Mode instead of Regular Mode.

One thing I have on my TODO list is to get my VS Code workspace running correctly. Been working without linting / protos / etc set up. So I'll try to run the instructions there.

I skimmed through the rest of the documentation and it seems current to me.

wwwillchen:
> For the development page, maybe add a line explaining the use cases where you would use Dev Mode instead of Regular Mode.
> 
> One thing I have on my TODO list is to get my VS Code workspace running correctly. Been working without linting / protos / etc set up. So I'll try to run the instructions there.
> 
> I skimmed through the rest of the documentation and it seems current to me.

I'm going to open a follow-up PR to clean-up our dev mode, it actually doesn't work anymore and I don't think there's a ton of benefit to it with a bit of tweaking our main CLI.

Try to setup the VS code workspace :) It's a bit of work but then it's really nice to get proper intellisense, type-checking!

PR #315, opened 2024-05-28
Add `scroll_into_view` command

### LINKED ISSUE #106: Add a focus / scrollToComponent API
For things like a chat use case #100, it would be convenient to scroll and/or change the focus.

It could be something like:

```
me.scrollToComponent(key)
```

The only thing is if key is _not_ unique (right now it only needs to be unique amongst its sibling components), then it will create an ambiguity of where to scroll. We could scroll to the first one (somewhat arbitrarily) and then display a warning somewhere (we'd need to create some kind of warning UI, similar to what we have for displaying errors, but less intrusive).

wwwillchen:
Closes #106.

- Also adds additional documentation for the existing `navigate` command.

### DIFF
--- a/mesop/examples/scroll_into_view.py
+++ b/mesop/examples/scroll_into_view.py
@@ -1,0 +1,21 @@
+import mesop as me
+
+
+@me.page(path="/scroll_into_view")
+def app():
+  me.button("Scroll to middle line", on_click=scroll_to_middle)
+  me.button("Scroll to bottom line", on_click=scroll_to_bottom)
+  for _ in range(100):
+    me.text("Filler line")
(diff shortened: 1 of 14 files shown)


mesop/examples/scroll_into_view.py
```diff
@@ -0,0 +1,13 @@
+import mesop as me
+
+
+@me.page(path="/scroll_into_view")
+def app():
+  me.button("Scroll to bottom line", on_click=on_click)
```
richard-to, line 6: [suggestion] Maybe also add a scroll to middle as well.

richard-to (approved):
LGTM - Nice. Can't wait to add this to mel.chat()

PR #318, opened 2024-05-28
Provide an `on_load` hook for page load

### LINKED ISSUE #97: Provide an on load hook for processing data on page load

wwwillchen:
Fixes #97.

### DIFF
--- a/mesop/examples/docs/on_load.py
+++ b/mesop/examples/docs/on_load.py
@@ -1,0 +1,28 @@
+import time
+
+import mesop as me
+
+
+def fake_api():
+  yield 1
+  time.sleep(1)
+  yield 2
+  time.sleep(2)
+  yield 3
+
+
+def on_load(e: me.LoadEvent):
+  for val in fake_api():
+    me.state(State).default_values.append(val)
+    yield
+
+
+@me.page(path="/docs/on_load", on_load=on_load)
+def app():
+  me.text("onload")
+  me.text(str(me.state(State).default_values))
+
+
+@me.stateclass
+class State:
+  default_values: list[int]
--- a/mesop/server/server.py
+++ b/mesop/server/server.py
@@ -9,8 +9,10 @@ import mesop.protos.ui_pb2 as pb
 from mesop.component_helpers import diff_component
 from mesop.editor.component_configs import get_component_configs
 from mesop.editor.editor_handler import handle_editor_event
+from mesop.events import LoadEvent
 from mesop.exceptions import format_traceback
 from mesop.runtime import runtime
+from mesop.warn import warn
 
 LOCALHOSTS = (
   # For IPv4 localhost
@@ -18,6 +20,8 @@ LOCALHOSTS = (
   # For IPv6 localhost
   "::1",
 )
+
+STREAM_END = "data: <stream_end>\n\n"
 
 
 def is_processing_request():
@@ -68,7 +72,7 @@ def configure_flask_app(
       )
       yield serialize(data)
       if not keep_alive:
-        yield "data: <stream_end>\n\n"
+        yield STREAM_END
     except Exception as e:
       print(e)
       if e in exceptions_to_propagate:
@@ -91,7 +95,7 @@ def configure_flask_app(
     ui_response = pb.UiResponse(error=error)
 
     yield serialize(ui_response)
-    yield "data: <stream_end>\n\n"
+    yield STREAM_END
 
   def serialize(response: pb.UiResponse) -> str:
     encoded = base64.b64encode(response.SerializeToString()).decode("utf-8")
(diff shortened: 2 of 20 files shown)


mesop/server/server.py
```diff
@@ -106,7 +109,23 @@ def generate_data(ui_request: pb.UiRequest) -> Generator[str, None, None]:
...
+          result = page_config.on_load(LoadEvent(path=ui_request.path))
+          # on_load is a generator function then we need to iterate through
+          # the generator object.
+          if result:
+            for _ in result:
+              yield from render_loop(
+                path=ui_request.path, init_request=True, keep_alive=True
+              )
+              runtime().context().set_previous_node_from_current_node()
+              runtime().context().reset_current_node()
```
richard-to, line 123: [question] Did you test cases where component diffing could come into play? I feel like that's an area for potential regression. Most likely should be fine. But a good place to have some test coverage.

mesop/server/server.py
```diff
@@ -126,15 +145,34 @@ def generate_data(ui_request: pb.UiRequest) -> Generator[str, None, None]:
...
+                result = page_config.on_load(LoadEvent(path=path))
+                # on_load is a generator function then we need to iterate through
+                # the generator object.
+                if result:
+                  for _ in result:
+                    yield from render_loop(
+                      path=path, init_request=True, keep_alive=True
+                    )
+                    runtime().context().set_previous_node_from_current_node()
+                    runtime().context().reset_current_node()
```
richard-to, line 170: [question] Ditto here, did you test cases where component diffing could come into play? I feel like that's an area for potential regression. Most likely should be fine. But a good place to have some test coverage.

mesop/server/server.py
```diff
@@ -126,15 +145,34 @@ def generate_data(ui_request: pb.UiRequest) -> Generator[str, None, None]:
         runtime().context().reset_current_node()
 
         result = runtime().context().run_event_handler(ui_request.user_event)
+        has_run_navigate_on_load = False
         for _ in result:
           path = ui_request.path
           for command in runtime().context().commands():
             if command.HasField("navigate"):
               path = command.navigate.url
+              page_config = runtime().get_page_config(path=path)
```
richard-to, line 154: [nit] I feel like this section, lines 154-170 is could be cleaned up a bit code-wise. No real suggestion though. It's not entirely the same as the on load behavior above. But seems very similar. I think ok for now since we don't want to make a poor abstraction.

I think the one part I have trouble understanding is the use of `has_run_navigate_on_load`. What happens on when there are multiple navigate commands? Does that mean we're navigating to multiple different pages? If so, seems like we'd need to run the load event each time if each page had an onload?

mesop/examples/docs/on_load.py
```diff
@@ -0,0 +1,16 @@
+import mesop as me
+
+
+def on_load(e: me.LoadEvent):
+  me.state(State).default_values = ["a", "b"]
```
richard-to, line 5: I feel like for an example, we want a better example here. In this case it's static, so technically one could use `field(lambda: ["a", "b"])` in the state class instead of the load event.

But I guess in the more dynamic case, one could also use: ``field(lambda: random.shuffle(["a", "b"])` as well.

I think that's the one thing about the `on_load`. I'm wondering what's a really good use case for it. I feel like there is one though. 

I mean using the `field()` function is bit weird since we import from dataclasses. But it does work.

richard-to (approved):
LGTM overall. Just a few questions I had mostly.

mesop/examples/docs/on_load.py
wwwillchen: Good point, I updated this simulate a fake streaming API that's slow.

This is the main reason I added on_load because you'll often want to do some kind slow API call (e.g. RPC, open a file, etc), and the current trick of loading state inside a component state has no way of letting you yield intermediate states (esp. loading/in-progress states).

mesop/server/server.py
wwwillchen: I do a test across navigations here: mesop/tests/e2e/on_load_test.ts

wwwillchen: Found an existing regression and filed: https://github.com/google/mesop/issues/371

mesop/server/server.py
wwwillchen: Updated tests in mesop/tests/e2e/on_load_test.ts to be more comprehensive

mesop/server/server.py
wwwillchen: So to clarify, you can do this:

```py
def on_load():
  me.navigate("/p1")
  yield
  me.navigate("/p2")
  yield
```

This probably isn't a great UX, but it's allow-able and p1's onload will be executed on the first tick and then p2's onload will be executed on the second tick (note: p1 is cleared from the commands after it's been run once).

- Added a warning if there's multiple navigate commands found at once.

wwwillchen:
Thanks for the review!

PR #335, opened 2024-06-01
Check is same site instead of same origin for CSRF check

wwwillchen:
When deploying on Google Cloud Run - the origin is https, however the url root is http, so that's why was it failing the is same origin check.

Related to #310

### DIFF
--- a/mesop/server/server.py
+++ b/mesop/server/server.py
@@ -161,12 +161,12 @@ def configure_flask_app(
 
   @flask_app.route("/ui", methods=["POST"])
   def ui_stream() -> Response:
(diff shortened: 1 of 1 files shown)


richard-to (approved):
LGTM - Do you think it is generally ok to check that just the site matches? Is this a common scenario?

wwwillchen:
> LGTM - Do you think it is generally ok to check that just the site matches? Is this a common scenario?

I think it's okay-ish, but not great. My guess is this is a fairly common scenario where there's some kind of proxy in front of the server that flask is running on but I haven't tested this out on other Cloud services.

Eventually, we should do other CSRF mitigations outlined in #310.

Issue #361, opened 2024-06-06
Update grpc-tools to 1.12.4 to support Linux Arm builds


richard-to:
I was trying to install Mesop for development on Linux (Debian). It kept failing. I tracked it down to this https://github.com/grpc/grpc-node/issues/1405.

I updated the dependency to 1.12.4 and was able to complete the build. Are we allowed to update or do we need to keep that pinned at 1.11.3?

wwwillchen:
I think we can upgrade

richard-to:
One thing I noticed just now is that it seems create this warning message:

```
external/com_google_protobuf_javascript/generator/js_generator.cc:3932:23: warning: comparison of integer expressions of different signedness: 'int' and 'std::vector<const google::protobuf::FileDescriptor*>::size_type' {aka 'long unsigned int'} [-Wsign-compare]
 3932 |     for (int i = 0; i < files.size(); i++) {
```

Issue #101, opened 2024-04-04
Add support for Github Codespaces / Docker image for quick development start up


richard-to:
To make it easier for other devs to get started with development, it could be helpful to configure a Github Codespace. I think that also depends on creating a Docker image, which by itself is also helpful for quick development start up. It's main benefit is that devs wouldn't need to worry about debugging dependency versioning / OS specific issues.

PR #363, opened 2024-06-06
Add devcontainer support for Mesop dev.

richard-to:
- This change allows one click installation with Github Codespaces
  - Using the default resources, it does take a long time to install
- This change also allows remote container dev in VSCode
  - Running ibazel in the container is  a bit slow
  - Does not support Linux ARM yet. See #361
  - Also can't push commits unless ssh keys are set up
    - Need to see if there is a better way to do that
- Will add docs later
- Probably still need to work out various kinks

Ref #101

# Mesop Demo from Codespaces
Screenshot 2024-06-05 at 8 16 17 PM

# Mesop VSCode workspace from Codespaces
Screenshot 2024-06-05 at 8 16 34 PM

### DIFF
--- a/scripts/devcontainer_setup.sh
+++ b/scripts/devcontainer_setup.sh
@@ -1,0 +1,25 @@
+# postCreateCommands for devcontainer.json
+
+# Ensure that the node modules directory is writable.
+#
+# We're storing node modules on a Docker volume (not to be confused with bind mounts on
+# the host OS) for better performance. When this volume gets created, it is owned by
+# root.
+sudo chown mesop-dev:mesop-dev node_modules
+
+# Update third party python packages needed for Mesop.
+bazel run //build_defs:pip_requirements.update
+
+# Virtual Env to use with VS Code.
+bazel run //mesop/cli:cli.venv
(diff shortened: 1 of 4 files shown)


scripts/devcontainer_setup.sh
```diff
@@ -0,0 +1,21 @@
+# postCreateCommands for devcontainer.json
+
+# Ensure that the node modules directory is writable.
+sudo chown mesop-dev:mesop-dev node_modules
+
```
wwwillchen, line 5: fwiw, I think generally you don't need to run this, but it doesn't hurt either.

wwwillchen (approved):
Thanks!

scripts/devcontainer_setup.sh
richard-to: The issue here is that Docker volume that gets created is owned by root. There does not appear to be a way to change this in the configuration since the node_modules directory gets created on the bind mount on the host with the source code.

If it wasn't on the bind mount, we could create the directory and change permissions ahead of time like we do with the vs code extensions  volume.

PR #372, opened 2024-06-07
Only show channel progress indicator if it's taking >500ms to avoid jarring UX

### LINKED ISSUE #365: Debounce progress indicator
Context: https://www.reddit.com/r/LocalLLaMA/comments/1d8g5j1/comment/l7cbwh4/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

wwwillchen:
closes #365

### DIFF
--- a/mesop/web/src/services/channel.ts
+++ b/mesop/web/src/services/channel.ts
@@ -20,6 +20,11 @@ import {applyComponentDiff} from '../utils/diff';
 const anyWindow = window as any;
 const DEV_SERVER_HOST = anyWindow['MESOP_SERVER_HOST'] || '';
 
+// Pick 500ms as the minimum duration before showing a progress/busy indicator
+// for the channel.
+// See: https://github.com/google/mesop/issues/365
+const WAIT_TIMEOUT_MS = 500;
+
 interface InitParams {
   zone: NgZone;
   onRender: (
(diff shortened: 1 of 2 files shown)


richard-to:
Yeah this is an interesting issue. I agree that on the on input it is can be feel jarring and visually makes Mesop feel slower. I often wait for the indicator to stop before I press submit. However, that's because I know it may not include all my text if I submit too early. 

Especially for long running operations, I definitely think a good practice is to force disable other inputs. Otherwise you run the possibility of user's queuing up many actions and that can be weird. This can be a useful indicator that something is happening.

I haven't tested out how this feels (haven't tested out the branch), but I'll keep an eye on it when I'm testing out Mesop apps.

richard-to (approved):
LGTM

wwwillchen:
> Yeah this is an interesting issue. I agree that on the on input it is can be feel jarring and visually makes Mesop feel slower. I often wait for the indicator to stop before I press submit. However, that's because I know it may not include all my text if I submit too early.
> 
> Especially for long running operations, I definitely think a good practice is to force disable other inputs. Otherwise you run the possibility of user's queuing up many actions and that can be weird. This can be a useful indicator that something is happening.
> 
> I haven't tested out how this feels (haven't tested out the branch), but I'll keep an eye on it when I'm testing out Mesop apps.

Yeah, makes sense. You should document that pattern of disabling other inputs in our docs :)

PR #382, opened 2024-06-08
Add steps for creating an RC for pip publishing process

### LINKED ISSUE #327: Setup a mesop-rc pip package
This will enable us to Mesop pre-release versions

wwwillchen:
@richard-to I think it would be good for you to do a pip release, just so you're familiar with it. Let me know if you have any questions! (I will also need to grant you PyPI access).

Closes #327.

### DIFF
--- a/docs/internal/publishing.md
+++ b/docs/internal/publishing.md
@@ -2,9 +2,15 @@
 
 Follow these instructions for releasing a new version of Mesop publicly via PyPI (e.g. `pip install mesop`).
 
-## Bump the version
+If you haven't done this before, follow the [first-time setup](#first-time-upload-setup).
 
-Update `mesop/version.py` by incrementing the version number. We follow semver.
+## Update version to RC
+
+Update [`mesop/version.py`](https://github.com/google/mesop/blob/main/mesop/version.py) by incrementing the version number. We follow semver.
+
+You want to first create an RC (release candidate) to ensure that it works.
+
+For example, if the current version is: `0.7.0`, then you should increment the version to `0.8.0rc0` which will create an RC, which is treated as a [pre-release by PyPI](https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases).
 
 ## Install locally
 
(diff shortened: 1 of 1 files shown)


richard-to:
Yeah that sounds good. The instructions look mostly straightforward.

docs/internal/publishing.md
```diff
@@ -2,9 +2,15 @@
...
+If you haven't done this before, follow the [first-time setup](#first-time-upload-setup).
 
-Update `mesop/version.py` by incrementing the version number. We follow semver.
+## Update version to RC
+
+Update [`mesop/version.py`](https://github.com/google/mesop/blob/main/mesop/version.py) by incrementing the version number. We follow semver.
+
+You want to first create an RC (release candidate) to ensure that it works.
+
+For example, if the current version is: `0.7.0`, then you should increment the version to `0.8.0rc0` which will create an RC, which is treated as a [pre-release by PyPI](https://packaging.python.org/en/latest/specifications/version-specifiers/#pre-releases).
```
richard-to, line 13: [question] I got a bit confused here. I'm also not too familiar with release candidate versioning stuff.

In the example, you say if the version is 0.7.0, then we should increment to version 0.8.0rc0. Is that always true. Would it be possible do a 0.7.1rc0 for a minor patch change or is the rc process only for incremental major/minor versions?

docs/internal/publishing.md
```diff
@@ -50,25 +56,21 @@ rm -rf /tmp/mesoprelease-test/venv-twine \
...
+## Change the version from RC to regular release
 
-```sh
-pip install --index-url https://test.pypi.org/simple/ mesop
-```
+If you find an issue, then redo the above steps and create another RC candidate: `0.8.0rc0` -> `0.8.0rc1`.
+
+If all the testing looks good, then you can update [`mesop/version.py`](https://github.com/google/mesop/blob/main/mesop/version.py) and change the version from RC to a regular release, for example:
+
+`0.8.0rc0` -> `0.8.0`
```
richard-to, line 71: [question] Does this end up replace the release candidate? Or does the release candidate still remain?

richard-to (approved):
LGTM

docs/internal/publishing.md
wwwillchen: I think you can do 0.7.1rc0

Because we are pre v1, we aren't super strict about versioning. Technically in semver 0.7 to 0.8 can contain breaking changes, but in my mind, it could just also be a bunch of new features.

I think we can just stick to bumping the middle version, unless it's a really small release (e.g. 1 bug fix)

docs/internal/publishing.md
wwwillchen: It doesn't replace. In general, once you publish to pip, it should stay there forever to not break people downloading at pinned versions, but you can remove a release if you accidentally published something, but that should be rare.

Issue #347, opened 2024-06-05
Create an installation guide page


wwwillchen:
Similar to Streamlit's [installation guide](https://docs.streamlit.io/get-started/installation), it would be helpful to show to create your own venv, which is more reliable than installing it in a global environment.

wwwillchen:
Can reference #351 as an alternative to Colab

Issue #337, opened 2024-06-01
Revamp docs


wwwillchen:
- Create an API section - fixed in #377 
  - Component properties (key, style)
  - `me.component()`
  - Page
  - commands (navigate, scroll_into_view)

- Create a State management section (expand from interactivity)
  - Explain how state is serialized
  - default values - don't use list
  - Using stateclass & dataclass together
  - Mesop instantiates state
  - Advanced tips & tricks (e.g. using static variables, storing it in FS/database)

 - Split interactivity into multiple pages
   -  Basic interactivity
   -  Streaming UI

- Roadmap
  - Improved state management - better DevEx and performance
  - More components and commands
  - Custom components

wwwillchen:
Revamped docs - will add roadmap later on.

PR #383, opened 2024-06-08
Revamp docs: getting started + guides + FAQ

### LINKED ISSUE #346: Create FAQ page
Frequently asked questions:

- How does Mesop work?
- What kinds of apps / use cases is Mesop suited for?
- Is Mesop production ready?
- How is the Python code executed? All Python code is executed on the server.
- How do I deploy a Mesop app?
- Do I have to deploy on Google Cloud Run?
- Does Mesop share data with Google or anyone else? Mesop itself does _not_ have any telemetry. Mesop loads fonts via Google Fonts and you can read about [Google Fonts privacy and data collection](https://developers.google.com/fonts/faq/privacy).
- Can I build a mobile app? Or desktop app?

wwwillchen:
Closes #346, #347, #337


README.md
```diff
@@ -1,75 +1,8 @@
 # Mesop: Build delightful web apps quickly in Python 🚀
 
-### Used at Google for rapid internal app development
+If you're interested in learning how to use Mesop, please check out our main docs at: google.github.io/mesop/
```
richard-to, line 3: [question] does google.github.io/mesop automatically get rendered as a link?

------------

I think this is an interesting change. Do you think people that don't know about Mesop will land on the repo first or the github docs? 

One slight concern I have with this change to the docs is that we make the assumption that people will know what Mesop is. So the README here which for repositories is a kind of home page, feels less friendly now. 

I think the benefit of this change is you don't have to manage both the Docs and also this README home page which may have similar information. And the Docs can be more detailed since info can be spread across multiple pages.

I also wondering if people will click on the link if they don't know what Mesop is.

docs/faq.md
```diff
@@ -0,0 +1,39 @@
...
+  - navigation
+---
+
+# Frequently Asked Questions
+
+## General
+
+### What kinds of apps is Mesop suited for?
+
+Mesop is particularly well-suited for ML/AI demos and internal tools because it enables developers without frontend experience to quickly build web apps. For use cases that prioritize developer experience and velocity, Mesop can be a good choice.
```
richard-to, line 12: [nit] cut particularly. I don't think it adds much.

docs/faq.md
```diff
@@ -0,0 +1,39 @@
...
+
+For apps with stringent demands, either in terms of performance or UI customization, then other UI frameworks may be more suitable.
```
richard-to, line 14: I feel like "UI customization" could be misconstrued here. I feel like you can heavily customize the UI using Styles. There are certainly limitations in the styling and also we don't have all the Angular components and we don't have a way for users to create custom components yet. So maybe need to be more clear.

I think the issue could be the use of "stringent". People may take a stringent in different ways. I feel like the better word may be something like "production". At the moment, Mesop is definitely not something people would want to run in a serious production application yet.

docs/guides/state_management.md
```diff
@@ -0,0 +1,118 @@
...
+    response = call_api()
+    state = me.state(State)
+    state.nested = NestedState(val=response.text)
+```
+
+If you didn't explicitly annotate NestedState as a dataclass, then you would get an error instantiating NestedState because there's no initializer defined.
+
+## Tips
+
+### Avoid mutable default values (e.g. list)
```
richard-to, line 112: [nit] I think if you tried to set a default like `var1: list[int] = [1]`, you'd get an error I think.

The dataclass solution is to use `field(default_factory=lambda: [1])` which would create a separate instance each time. Or also can use the onload.

So I feel this tip is could be worded a bit differently. Because we can set defaults values for mutable values.  Just not the more intuitive way,

docs/guides/state_management.md
```diff
@@ -0,0 +1,118 @@
...
+
+@me.stateclass
+class State:
+  nested: NestedState
+
+def app():
+  state = me.state(State)
+```
+
+> Reminder: because dataclass do not have default values, you will need to explicitly set default values, otherwise Mesop will not be able to instantiate an empty version of the class.
```
richard-to, line 97: [typo] "dataclasses do" or "dataclass does"?

docs/guides/state_management.md
```diff
@@ -0,0 +1,118 @@
+# State Management
+
+State management is a critical element of building interactive apps because it allows you store information about what the user did in a structured way.
+
+## Basic usage
+
+You can register a class using the class decorator `me.stateclass` which is like dataclass with special powers:
```
richard-to, line 7: [typo] "is like a dataclass" or "is a dataclass"

docs/guides/state_management.md
```diff
@@ -0,0 +1,118 @@
...
+    me.text(state.val)
+```
+
+## How State Works
+
+`me.stateclass` is a class decorator which tells Mesop that this class can be retrieved using the `me.state` method, which will return the state instance for the current user session.
+
+> If you are familiar with the dependency injection pattern, Mesop's stateclass and state API is essentially a minimalist dependency injection system which scopes the state object to the lifetime of a user session.
+
+Under the hood, Mesop is sending the state back and forth between the server and browser client so everything in a state class, recursively, must be serializable.
```
richard-to, line 30: [nit] Maybe cut recursively. Not sure if it adds anything.

docs/guides/state_management.md
```diff
@@ -0,0 +1,118 @@
...
+```py
+def on_click(e):
+    response = call_api()
+    state = me.state(State)
+    state.nested = NestedState(val=response.text)
+```
+
+If you didn't explicitly annotate NestedState as a dataclass, then you would get an error instantiating NestedState because there's no initializer defined.
+
+## Tips
```
richard-to, line 110: [nit] May want to add a note about how dataclasses/classes sometimes do not get deserialized correctly when nested in more complex ways. But I think that's a more of a bug, so probably don't need to add it.

docs/guides/interactivity.md
```diff
@@ -1,63 +1,104 @@
...
 
-## Slow / async patterns
+1. The server is setting the input value based on state.
+2. The client is setting the input value based on what the user is typing.
 
-These are patterns for dealing with common use cases such as calling a slow blocking API call or a streaming API call.
+The way to fix this is by *not* setting the input value from the server.
 
-### Loading
+The above example **corrected** would look like this :white_check_mark::
```
richard-to, line 60: [nit] May also want to include an example of setting a initial value for an input or updating the initial value. Also clearing using input may be good things to have.

richard-to (approved):
LGTM overall. Just some minor comments.

README.md
wwwillchen: that's fair, I kept the readme content and just added the links to the top (I think it linkifies, but i formatted it as markdown links just to make it look nicer)

docs/faq.md
wwwillchen: OK, I reworded "ui customization". I don't know if "production" is better because we are running things like Demo Gallery, which I would argue is a Mesop app in production (e.g. it has external traffic w/ good availability, etc.).

In my mind, it's basically - don't try to run a consumer-grade app e.g. with millions of users which requires a11y, i18n, SEO-optimization, etc. I've rephrased this below

docs/guides/interactivity.md
wwwillchen: let's add this later (this was a copy & paste from before)

docs/guides/state_management.md
wwwillchen: thx. added the solution to use field(default_factory). `onload` can produce a similar effect, but it's not quite the same since it can get executed multiple times for a user session (e.g. navigate to the same page twice).

docs/guides/state_management.md
wwwillchen: Yeah, this is a bug and we should at the least raise an error, and ideally handle it correctly.

PR #384, opened 2024-06-08
Pick Material 3 color palette by hand

wwwillchen:
Although Material 3 recommends generating color schemes automatically (which Angular Material has pre-generated a sample), in practice, hand-picking color schemes can be better.

# Before

image

# After

Screenshot 2024-06-08 at 1 02 34 AM
r:


richard-to (approved):
LGTM - Are there any other potential future benefits picking the color palette by hand like this -- (i.e. is this a step toward users being able to customize the theme)?

wwwillchen:
> LGTM - Are there any other potential future benefits picking the color palette by hand like this -- (i.e. is this a step toward users being able to customize the theme)?

We could, theoretically, expose the CSS variables for material theming and allow people to customize the theme that way (it would add more bloat though). I think the main challenge is that it's tough to create a good looking theme, creating this palette wasn't that easy for me. I think having a few themes that people can swap between might be the first step, also we still need to support dark theme 😀

PR #386, opened 2024-06-08
Update is same site

djm93dev:
### Updates to the `is_same_site` Function

**Consolidation of Initial Checks:**
The checks for whether `url1` or `url2` is `None` have been combined into a single line.


wwwillchen:
thanks for contributing

PR #389, opened 2024-06-08
Improve state performance by diffing state

richard-to:
- Uses DeepDiff to diff state objects
- Updates code to return state as final data stream response

# Chat Before

Screenshot 2024-06-08 at 12 57 03 PM

# Chat After

Screenshot 2024-06-08 at 12 53 35 PM

Ref #62


mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -127,3 +139,73 @@ def decode_mesop_json_state_hook(dct):
...
+  those manually for the DataFrame case.
+  """
+  if not is_dataclass(state1) or not is_dataclass(state2):
+    raise MesopException("Tried to diff state which was not a dataclass")
+
+  custom_actions = []
+
+  # Only use the `DataFrameOperator` if the state class has DataFrames since we don't
+  # want to import Pandas unnecessarily.
+  if "DataFrame" in str(state1.__class__.__annotations__):
```
wwwillchen, line 189: If the DataFrame is in a nested class (e.g. dataclass) inside a state class, Will this work?

mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -127,3 +139,73 @@ def decode_mesop_json_state_hook(dct):
...
+
+      return isinstance(level.t1, pd.DataFrame) and isinstance(
+        level.t2, pd.DataFrame
+      )
+    except ImportError:
+      # If Pandas is not installed, don't perform this check. We should log a warning.
+      return False
+
+  def give_up_diffing(self, level, diff_instance) -> bool:
+    if not level.t1.equals(level.t2):
```
wwwillchen, line 165: Just double checking - is this using https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.equals.html#pandas-dataframe-equals which is doing a deep comparison check (and not an identity check)?

mesop/server/server.py
```diff
@@ -81,6 +77,28 @@ def render_loop(
         error=pb.ServerError(exception=str(e), traceback=format_traceback())
       )
 
+  def create_update_state_event(diff: bool = False) -> str:
```
wwwillchen, line 80: nit: I think you can pull this out as a top-level fn definition since it doesn't look like it's using closure variables.

mesop/web/src/services/channel.ts
```diff
@@ -122,8 +123,33 @@ export class Channel {
...
+            if (states === undefined) {
+              console.error('Update state event is empty.');
+              return;
+            }
+
+            if (
+              uiResponse.getUpdateStateEvent()!.getUpdateStrategy() ===
+              UpdateStateEvent.UpdateStrategy.UPDATE_STRATEGY_DIFF
+            ) {
+              // `this.states` should be populated at this point since the first update
```
wwwillchen, line 138: This shouldn't happen normally, but it's probably worth checking that this.states and states have the same length. There could catch a weird edge case where someone tries to lazily register a stateclass.

mesop/web/src/services/channel.ts
```diff
@@ -157,7 +183,7 @@ export class Channel {
             onRender(this.rootComponent, this.componentConfigs);
             this.logger.log({
               type: 'RenderLog',
-              states: this.states,
+              states: this.states || new States(), // States is undefined on init render.
```
wwwillchen, line 186: nit: initialize this.states = new States() on L50 instead?

mesop/protos/ui.proto
```diff
@@ -145,13 +146,24 @@ message ContextLine {
...
+    repeated Command commands = 4;
 
     // Only sent in editor mode:
-    repeated ComponentConfig component_configs = 4;
+    repeated ComponentConfig component_configs = 5;
+}
+
+// UI response event for updating state.
+message UpdateStateEvent {
+    optional States states = 1;
```
wwwillchen, line 159: Instead of using an enum, you could model this as:

oneof state_type {
   States full_states = 1
   States state_diffs = 2;
}

I think it's a little clearer, but if you prefer enum, that's OK since this in an implementation detail.

mesop/runtime/context.py
```diff
@@ -120,9 +125,36 @@ def serialize_state(self) -> pb.States:
...
+    states = pb.States()
+    for (key, state), previous_state in zip(
+      self._states.items(), self._previous_states.values()
+    ):
+      states.states.append(pb.State(data=diff_state(previous_state, state)))
+      # If state has not changed no need to update the previous state since it should be
+      # the same.
+      #
+      # One thing to note is that there seems to be an issue multiple states with the
+      # same key can exist (i.e. usage in Mesop demo app). It's unclear if this causes
```
wwwillchen, line 138: This seems like a weird bug. Can you file an issue?

mesop/runtime/context.py
```diff
@@ -120,9 +125,36 @@ def serialize_state(self) -> pb.States:
...
+      # problem here.
+      if state != self._previous_states[key]:
+        self._previous_states[key] = copy.deepcopy(state)
```
wwwillchen, line 141: I don't understand why we need to update _previous_states because, IIUC, diff_state is called at most once per context lifetime.

mesop/runtime/context.py
```diff
@@ -120,9 +125,36 @@ def serialize_state(self) -> pb.States:
...
+    return states
+
   def update_state(self, states: pb.States) -> None:
-    for state, proto_state in zip(self._states.values(), states.states):
+    for state, previous_state, proto_state in zip(
+      self._states.values(), self._previous_states.values(), states.states
+    ):
       update_dataclass_from_json(state, proto_state.data)
+      # We should check if state and previous state are the same. There should be no
```
wwwillchen, line 149: Same as above, I don't understand why this is needed.

mesop/web/src/utils/diff.ts
```diff
@@ -74,3 +74,139 @@ export function applyComponentDiff(component: Component, diff: ComponentDiff) {
...
+
+// Interface for state diff objects.
+interface StateDiff {
+  path: (string | number)[];
+  action: string;
+  value: any;
+}
+
+// Applies state diffs to the state object.
+export function applyStateDiff(state_json: string, diff_json: string): string {
```
wwwillchen, line 94: nit: use camelCase variable names in JS

wwwillchen (commented):
Lots of good stuff here, few questions about the implementation and a handful of minor suggestions.

mesop/dataclass_utils/dataclass_utils.py
richard-to: Yeah, initially I was trying to to level.t1 == level.t2, but it kept failing with "ValueError: The truth value of a DataFrame is ambiguous".

I wasn't quite sure how to resolve it. Probably need to look into it more.

mesop/runtime/context.py
richard-to: Great catch. I had this in before I added the change to only update only send the state back once.

mesop/runtime/context.py
richard-to: This case is still needed for now since we need to update the previous state to match the state at the start of each ui request since the state gets reset.

mesop/web/src/utils/diff.ts
richard-to: Doh. One of these days I'll remember. How come prettier doesn't fix this automatically?

mesop/web/src/services/channel.ts
richard-to: Hmm, yeah. How should we handle that case?

mesop/dataclass_utils/dataclass_utils.py
richard-to: Ah good catch. Won't work. I updated to check for pandas globally once. Not sure if it makes difference from just repeatedly checking if the import exists.

mesop/protos/ui.proto
richard-to: Funny thing about this is that earlier this week, someone suggested I change my oneof to use enums since all the oneofs were the same data type. I didn't quite buy the reasoning. Anyway updated to oneof.

mesop/runtime/context.py
richard-to: Bug: https://github.com/google/mesop/issues/391

mesop/web/src/services/channel.ts
wwwillchen: If the lengths are not the same, then I would just throw an exception, since it's very likely we'd produce the wrong state result anyways.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: repeatedly checking is probably fine b/c Python will cache after the first import, but this is fine too.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: ah ok, using .equals seems good to me.

mesop/protos/ui.proto
wwwillchen: lol

mesop/web/src/utils/diff.ts
wwwillchen: prettier is only for formatting (basically an AST printer) and won't actually rename a variable, but we can use biome (which includes a linter) to do this! filed https://github.com/google/mesop/issues/392

mesop/runtime/context.py
wwwillchen: To clarify, I mean that for each ui request, you should have a new context, so previous_states should get set in `__init__` so you shouldn't need to update the previous state.

mesop/runtime/context.py
wwwillchen: Not sure if you meant to change this, but I think we can remove any mutations to previous_states - see other comment.

mesop/web/src/services/channel.ts
richard-to: I see. I thought you had some kind of mitigation in mind. Updated.

mesop/runtime/context.py
richard-to: Yes, it's true that a new context gets created. But the previous states will equal the default state when the state is initialized. But we need the previous state to be the state at the start of the request. So in the case of demo or an app with many pages, most should have the default previous state. But the current page will need to the state updated. So there's some optimization that we can do. However we'd probably need to resolve #391 first.

wwwillchen: ok yeah, you're right

mesop/runtime/context.py
wwwillchen: With #394 - do we still need to mutate previous_states here?

wwwillchen (approved):
LGTM - thanks!

mesop/runtime/context.py
richard-to: Yeah, even with #394 this wasn't needed the update to only send state at the end of the stream. It should be removed already. I think I just forgot to add a follow up comment here.

PR #390, opened 2024-06-08
Integrate scroll to feature with chat demos

richard-to:


mesop/labs/chat.py
```diff
@@ -189,13 +189,28 @@ def on_submit(e: me.ClickEvent | me.EnterEvent):
...
+        if scroll_count < 2:
+          me.scroll_into_view(key="scroll-to")
+          scroll_count += 1
         yield
     state.in_progress = False
+
+    # If the content was generated too quickly, then the scroll action may not have run.
+    # This will scroll to the end of the message, so if it is a very long message, it will
+    # scroll too far. However, we'd assume that a long message would take longer to render
+    # if streamed.
```
wwwillchen, line 208: I think what Gemini UI does is that it creates a blank space (e.g. box with 600px height) and then it scrolls to have the user prompt + blank space be at the top, and then it doesn't scroll anymore. This way you don't have the issue of "over-scrolling" (i.e. scrolling faster than the user can read), I think this is a nicer UX than the "keep on scrolling" pattern other apps use.

wwwillchen (approved):

mesop/labs/chat.py
richard-to: Great idea. That's simplifies things a lot.

PR #395, opened 2024-06-09
Configure custom fonts/stylesheets instead of hardcoding

### LINKED ISSUE #364: Configure Google fonts per page
Right now it's hard coded

wwwillchen:
Early on, I hardcoded several Google-specific fonts for a demo, and this basically unwinds some of the tech debt as it's just extra bloat for most Mesop apps, particularly externally.

Downstream, I'll need to set a default_stylesheets value for these Google-specific fonts to avoid affecting current Mesop apps in Google.

Fixes #364

### DIFF
--- a/mesop/web/src/app/editor/index.html
+++ b/mesop/web/src/app/editor/index.html
@@ -5,23 +5,14 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title>Mesop</title>
     <base href="/" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
       href="https://fonts.googleapis.com/css2?family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@48,400,0,0"
       rel="stylesheet"
     />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Google+Sans:wght@400;500;700"
-      rel="stylesheet"
-    />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Google+Sans+Text:wght@400;500;700"
-      rel="stylesheet"
-    />
-    <link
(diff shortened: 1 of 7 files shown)


mesop/web/src/app/editor/index.html
```diff
@@ -5,22 +5,13 @@
...
-    />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Google+Sans+Text:wght@400;500;700"
-      rel="stylesheet"
-    />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Google+Sans+Mono:wght@400;500;700"
-      rel="stylesheet"
-    />
+    <!-- Inject stylesheet (if needed) -->
```
richard-to, line 14: Since the user can now load in a stylesheet, technically that could allow them to update the markdown styles directly now which I think this is nice. But looks like the injection is before style.css, so style.css would overwrite some styles. Also wondering the if ./styles.css could also be injected. I think at the moment, this is fine since use of stylesheets would be pretty limited. I'm curious what you have in mind for stylesheets in general.

richard-to (approved):
LGTM

mesop/web/src/app/editor/index.html
wwwillchen: Good point, I put this after style.css now.

Re: stylesheets after this, I think we could support things like accepting a Stylesheet file (e.g. read a local css file and pass it into `@page(stylesheets=[...])`, but I'm not sure what's really needed here. I think we have pretty good coverage with the Style API, now that there's viewport size. I think the open questions are: 1) supporting styling states like `:hover` and 2) should we make it possible/easy to use libraries like Tailwind? For 1), I think _eventually_ we want to support this, but I haven't really heard of anyone asking for it. For 2), I think the question I have is whether we can do this in a fairly agnostic way as I don't want to couple Mesop too closely with Tailwind (or any other existing CSS library) as these things can go in and out of fashion and the way we stick very close to the web standards, IMO, is a strength.

PR #398, opened 2024-06-10
Implement autoplay option for audio component

T-Zaid:
This change implemented an `autoplay` function for audio component, as mentioned in #299.
I have added autoplay as a Boolean variable defaulted to `False`.

If autoplay feature is required, one have to call the function like,
```
me.audio(
    src="https://interactive-examples.mdn.mozilla.net/media/cc0-audio/t-rex-roar.mp3",
    autoplay = True
  )
```

### DIFF
--- a/mesop/components/audio/audio.ng.html
+++ b/mesop/components/audio/audio.ng.html
@@ -1,2 +1,6 @@
 <!-- always show controls, otherwise you can't interact with the audio. -->
-<audio controls [src]="config().getSrc()" [attr.autoplay]="config().getAutoplay()? 'autoplay' : null"></audio>
+<audio
+  controls
+  [src]="config().getSrc()"
+  [attr.autoplay]="config().getAutoplay()? 'autoplay' : null"
+></audio>
--- a/mesop/components/audio/audio.ng.html
+++ b/mesop/components/audio/audio.ng.html
@@ -2,5 +2,5 @@
 <audio
   controls
   [src]="config().getSrc()"
-  [attr.autoplay]="config().getAutoplay()? 'autoplay' : null"
+  [attr.autoplay]="config().getAutoplay()? true : null"
 ></audio>
--- a/mesop/components/audio/audio.ng.html
+++ b/mesop/components/audio/audio.ng.html
@@ -1,2 +1,2 @@
 <!-- always show controls, otherwise you can't interact with the audio. -->
-<audio controls [src]="config().getSrc()"></audio>
+<audio controls [src]="config().getSrc()" [attr.autoplay]="config().getAutoplay()? 'autoplay' : null"></audio>
--- a/mesop/components/audio/audio.py
+++ b/mesop/components/audio/audio.py
@@ -4,10 +4,7 @@ from mesop.component_helpers import insert_component, register_native_component
 
 @register_native_component
 def audio(
-  *,
-  src: str | None = None,
-  key: str | None = None,
-  autoplay: bool = False
+  *, src: str | None = None, key: str | None = None, autoplay: bool = False
 ):
   """
   Creates an audio component.
@@ -20,8 +17,5 @@ def audio(
   insert_component(
     key=key,
     type_name="audio",
-    proto=audio_pb.AudioType(
-      src=src,
-      autoplay=autoplay
-    ),
+    proto=audio_pb.AudioType(src=src, autoplay=autoplay),
   )
--- a/mesop/components/audio/audio.py
+++ b/mesop/components/audio/audio.py
@@ -7,12 +7,14 @@ def audio(
   *,
   src: str | None = None,
   key: str | None = None,
+  autoplay: bool = False
 ):
   """
   Creates an audio component.
 
   Args:
       src: The URL of the audio to be played.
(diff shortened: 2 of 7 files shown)


wwwillchen:
Looks like the pre-commit check for formatting failed, could you follow the steps for setting up your dev environment, particularly with [commit hooks](https://google.github.io/mesop/internal/development/#commit-hooks)? thanks.

T-Zaid:
I have run the pre-commit hooks now, all test passed locally.

tools/defaults.bzl
```diff
@@ -364,17 +364,17 @@ def ng_web_test_suite(browsers = [], deps = [], static_css = [], exclude_init_sc
...
-        css_content=$$(cat $${files[0]} | sed 's/"/\\\\"/g')
-        js_template='var cssElement = document.createElement("style"); \
-                    cssElement.type = "text/css"; \
-                    cssElement.innerHTML = "'"$$css_content"'"; \
-                    document.head.appendChild(cssElement);'
-         echo "$$js_template" > $@
+            cmd = """\r
+        files=($(execpaths %s))\r
+        # Escape all double-quotes so that the content can be safely inlined into the\r
+        # JS template. Note that it needs to be escaped a second time because the string\r
```
wwwillchen, line 370: can you undo these changes? they look un-related

mesop/components/audio/audio.ng.html
```diff
@@ -1,2 +1,6 @@
 <!-- always show controls, otherwise you can't interact with the audio. -->
-<audio controls [src]="config().getSrc()"></audio>
+<audio
+  controls
+  [src]="config().getSrc()"
+  [attr.autoplay]="config().getAutoplay()? 'autoplay' : null"
```
wwwillchen, line 5: Although using `'autoplay'` seems to work, I'd set this to `'true'` as it's clearer and what's specified by the standard: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/autoplay

wwwillchen (commented):
Thanks! Mostly looks good - could you add an example to https://github.com/google/mesop/blob/main/demo/audio.py to demonstrate how auto-playing works?

It's not super straightforward, because, on Chrome it doesn't work if I have an audio component with autoplay on load, because [Chrome essentially requires a page interaction](https://developer.chrome.com/blog/autoplay/) before allowing autoplay, so then I have to have a button -> on_click -> which causes an audio component with autoplay=True, before it can work.

tools/defaults.bzl
T-Zaid: These changes are made by pre-commit formatter. Looks weird, Ill remove it and run the pre-commit again to make sure of the changes.

T-Zaid:
> It's not super straightforward, because, on Chrome it doesn't work if I have an audio component with autoplay on load, because [Chrome essentially requires a page interaction](https://developer.chrome.com/blog/autoplay/) before allowing autoplay, so then I have to have a button -> on_click -> which causes an audio component with autoplay=True, before it can work.

Yes, I have read about it but couldn't come to a good conclusion about it. I would like to know your suggestion about some of the points I have written below,
- Adding a button would not be intuitive because by adding the `control` attribute, it's already giving an interactive button.
- If I add a custom button, and trying to click it programmatically, it gives error, `Uncaught (in promise) DOMException: play() failed because the user didn't interact with the document first.`, which is again defeating the purpose.

Should I change anything about the specific functionality?
or should keep that as it is, and mention all of these restrictions in the demo file you mentioned ?

mesop/components/audio/audio.py
```diff
@@ -4,21 +4,18 @@
...
-  src: str | None = None,
-  key: str | None = None,
+  *, src: str | None = None, key: str | None = None, autoplay: bool = False
 ):
   """
   Creates an audio component.
 
   Args:
       src: The URL of the audio to be played.
+      autoplay: boolean value indicating if the audio should be autoplayed or not.
```
wwwillchen, line 14: Add note here about the subtle issue with page interaction so that it shows up in the API docs.

wwwillchen:
> > It's not super straightforward, because, on Chrome it doesn't work if I have an audio component with autoplay on load, because [Chrome essentially requires a page interaction](https://developer.chrome.com/blog/autoplay/) before allowing autoplay, so then I have to have a button -> on_click -> which causes an audio component with autoplay=True, before it can work.
> 
> Yes, I have read about it but couldn't come to a good conclusion about it. I would like to know your suggestion about some of the points I have written below,
> 
> * Adding a button would not be intuitive because by adding the `control` attribute, it's already giving an interactive button.
> * If I add a custom button, and trying to click it programmatically, it gives error, `Uncaught (in promise) DOMException: play() failed because the user didn't interact with the document first.`, which is again defeating the purpose.
> 
> Should I change anything about the specific functionality? or should keep that as it is, and mention all of these restrictions in the demo file you mentioned ?

I'd just keep the functionality the same and just mention these in demo file and the inline pydocs. I don't think there's a way to circumvent the browser behavior.

T-Zaid:
I have addressed all the reviewed concerns in the latest push.

wwwillchen:
Thanks for contributing!

Issue #397, opened 2024-06-10
Install playwright in .devcontainer.json environment


richard-to:
Unable to run playwright in Github Code Spaces.

PR #404, opened 2024-06-10
Add steps to install playwright in .devcontainer.json

richard-to:
Ref: https://github.com/google/mesop/issues/397


Dockerfile
```diff
@@ -8,7 +8,25 @@ RUN apt-get update && \
   locales-all \
   lsof \
   tmux \
-  vim && \
+  vim \
+  # Playwright dependencies
+  # This is the qquivalent of `sudo yarn playwright install-deps`. We add these manually
```
wwwillchen, line 13: nit: typo

wwwillchen (approved):
LGTM - this seems a little painful to maintain if we upgrade playwright (but that's pretty rare I think).

Future idea: I don't know if this makes sense, but could we run Playwright from a separate Docker container and have it connect to the local dev container? This way we could use Playwright's included Docker container image https://playwright.dev/docs/docker

richard-to:
> LGTM - this seems a little painful to maintain if we upgrade playwright (but that's pretty rare I think).
> 
> Future idea: I don't know if this makes sense, but could we run Playwright from a separate Docker container and have it connect to the local dev container? This way we could use Playwright's included Docker container image https://playwright.dev/docs/docker

Yeah, I agree I was concerned about that as well. Unfortunately for some reason in Codespaces, the post set up command that gets run does not use root. And sudo doesn't work.

Weirdly when I do this locally, the post set up command seems to run as root, so for local dev container, I could just run the command.

Hmm, yeah using the separate docker container image seems like an interesting idea to look into. It's fairly easy  to create add container in the Docker Compose configuration. And we should be able to access the main container from the playwright container. It'd need a way to start up the server and some other things. So I think it mostly amounts to configuring things correctly for this use case.

PR #411, opened 2024-06-10
fix: serialize datetime

### LINKED ISSUE #387: set/tuple/datatime data types not JSON serializable
- The tuple is serializable but gets deserialized back to a list
- Set is not serializable at all
- Datetime also not serializable:

I think we can add support for these by doing some custom encoding parameters similar to what we did for Pandas in `MesopJSONEncoder`

For set and tuple, probably can do something like:

```
      if isinstance(obj, set):
        return {"[set]": obj}
```

Then in `decode_mesop_json_state_hook`

```
    if "[set]" in dct:
      return set(dct["[set]"]),
``

Haven't tested it though. Depends on how the encode decoding works. If it works it's way upwards, then this should work. But if the serialization/deserialization works from root to leaf, then wouldn't work for nested types. But I imagine it should work from leaf to root

--------

Datetime would be similar I think. But need to determine the format we want to store the date as. Some ISO standard or unix timetamp. Maybe some kind of ISO standard especially if we need to handle time zones.

----

Code link: https://github.com/google/mesop/blob/main/mesop/dataclass_utils/dataclass_utils.py

### PR DESCRIPTION
**Pull request by cullenwatson**

### CODE REVIEW

**Comment by richard-to** [comment]
Hi Cullen,

Thanks for your contribution. For this issue, I think the set/tuple cases are a bit trickier due to the state diffing and also the possibilities of nesting (which I see that you already have some tests for!). So you'll want to add thorough testing for the stae diffing use cases in mesop/dataclass_utils/diff_state_test.py and also mesop/web/src/utils/diff_state_spec.ts

For the state diffing on the client side, you'll need to add support for the set_added and set removed actions. See https://github.com/google/mesop/blob/main/mesop/web/src/utils/diff.ts#L94

Also the tuple case is interesting since I think the deep diff library we're using considers it an iterable added action. So there's no distinction between list and tuple.

For starters, I think the  addition of datetime sounds like a good starting PR. It's better to add changes in small self-contained PRs. Easier to review. Easier to consider more test cases. Then you can follow up with the changes for set support and then the tuple one.

**Comment by cullenwatson** [comment]
ok switched to just do datetime

File: mesop/dataclass_utils/dataclass_utils.py [hunk 1]
```diff
@@ -11,6 +12,7 @@
 
 _PANDAS_OBJECT_KEY = "__pandas.DataFrame__"
 _DIFF_ACTION_DATA_FRAME_CHANGED = "data_frame_changed"
+_DATETIME_OBJECT_KEY = "__mesop_datetime__"
```
richard-to, line 15: [nit] Can you move this line up below _PANDAS_OBJECT_KEY since those are related constants. I think for the name, maybe '__datetime.datetime__' since that's the module name.

richard-to (changes requested):
This looks good. Just one comment.

Plus also we should add a test to mesop/web/src/utils/diff_state_spec.ts to have an example that the datetime objects are added correctly, which I think they should but good to have the test case.

richard-to:
Also can you update the PR description so that the issue isn't close after we merge this one in (I probably should have just created separate issues but I'll leave as is for now.)

cullenwatson:
done

richard-to (approved):
Thanks. This looks good. I'll run the pre-submits and hopefully those will pass.

PR #422, opened 2024-06-12
Support for bokeh plot

techie-subhadeep:
This pull request adds support for creating interactive visualizations using [Bokeh](https://docs.bokeh.org/en/latest/index.html). The integration includes utilities for generating html plots and embedding them into the application.

### DIFF
--- a/mesop/components/bokeh_plot/bokeh_plot.py
+++ b/mesop/components/bokeh_plot/bokeh_plot.py
@@ -1,5 +1,6 @@
 from io import BytesIO
 from typing import Protocol
+
 from mesop.component_helpers import Style, component
 from mesop.components.html.html import html
 
@@ -25,4 +26,5 @@ def bokeh_plot(figure: figure, *, style: Style | None = None):
       plot component. If None, default styling (e.g. height, width) is used.
   """
   from bokeh.embed import file_html
+
   html(file_html(figure, "inline"), style=style)
--- a/mesop/components/bokeh_plot/bokeh_plot.py
+++ b/mesop/components/bokeh_plot/bokeh_plot.py
@@ -19,10 +19,10 @@ class figure(Protocol):
 @component(skip_validation=True)
 def bokeh_plot(figure: figure, *, style: Style | None = None):
   """
-  Creates a plot component from a Matplotlib figure.
+  Creates a plot component from a Bokeh figure.
 
   Args:
-    figure: A [Matplotlib figure](https://matplotlib.org/stable/api/figure_api.html#matplotlib.figure.Figure) which will be rendered.
+    figure: A [Bokeh figure](https://docs.bokeh.org/en/latest/docs/reference/plotting/figure.html) which will be rendered.
     style: An optional Style object that defines the visual styling for the
       plot component. If None, default styling (e.g. height, width) is used.
   """
--- a/mesop/components/bokeh_plot/bokeh_plot.py
+++ b/mesop/components/bokeh_plot/bokeh_plot.py
@@ -26,4 +26,4 @@ def bokeh_plot(figure: figure, *, style: Style | None = None):
     style: An optional Style object that defines the visual styling for the
       plot component. If None, default styling (e.g. height, width) is used.
   """
-  html(file_html(figure), style=style)
+  html(file_html(figure, "inline"), style=style)
--- a/mesop/components/bokeh_plot/bokeh_plot.py
+++ b/mesop/components/bokeh_plot/bokeh_plot.py
@@ -1,9 +1,7 @@
 from io import BytesIO
 from typing import Protocol
-from bokeh.plotting import figure
 from mesop.component_helpers import Style, component
 from mesop.components.html.html import html
-from bokeh.embed import file_html
 
 
 class figure(Protocol):
@@ -26,4 +24,5 @@ def bokeh_plot(figure: figure, *, style: Style | None = None):
     style: An optional Style object that defines the visual styling for the
       plot component. If None, default styling (e.g. height, width) is used.
   """
+  from bokeh.embed import file_html
   html(file_html(figure, "inline"), style=style)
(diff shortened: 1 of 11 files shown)


mesop/components/bokeh_plot/bokeh_plot.py
```diff
@@ -0,0 +1,29 @@
+from io import BytesIO
+from typing import Protocol
+from bokeh.plotting import figure
```
wwwillchen, line 3: can we remove this? it looks like it's not needed because figure is redefined below.

mesop/components/bokeh_plot/bokeh_plot.py
```diff
@@ -0,0 +1,29 @@
...
+from mesop.component_helpers import Style, component
+from mesop.components.html.html import html
+from bokeh.embed import file_html
+
+
+class figure(Protocol):
```
wwwillchen, line 9: figure -> Figure

mesop/components/bokeh_plot/bokeh_plot.py
```diff
@@ -0,0 +1,29 @@
+from io import BytesIO
+from typing import Protocol
+from bokeh.plotting import figure
+from mesop.component_helpers import Style, component
+from mesop.components.html.html import html
+from bokeh.embed import file_html
```
wwwillchen, line 6: Rather than importing here, which will make everyone using Mesop depend on bokeh, could you import bokeh  inside the bokeh_plot function? (i.e. L29).

wwwillchen (commented):
Thanks for contributing!

Could you add an E2E test? 
https://google.github.io/mesop/internal/testing/#e2e-tests

You can see some other tests here:
https://github.com/google/mesop/tree/main/mesop/tests/e2e

mesop/components/bokeh_plot/bokeh_plot.py
techie-subhadeep: The class name in bokeh library starts with small case [figure class github](https://github.com/bokeh/bokeh/blob/ce4ab47a86197e3742d82cbe7cb4565f840a7210/src/bokeh/plotting/_figure.py#L94)
do you still recommends the change?

wwwillchen: ah, I missed that. in that case let's keep it as-is, thanks.

wwwillchen:
Looks like some formatting failures, could you follow: https://google.github.io/mesop/internal/development/#commit-hooks - the pre-commit hooks should fix the formatting issues. thanks.

techie-subhadeep:
Hello @wwwillchen 
I have fixed pre-commit lint errors
image

Also added e2e tests, however I am not able to run pywright tests it failed with error `error Your lockfile needs to be updated, but yarn was run with --frozen-lockfile.`

Can you please run CI again?

wwwillchen:
> Hello @wwwillchen I have fixed pre-commit lint errors image
> 
> Also added e2e tests, however I am not able to run pywright tests it failed with error `error Your lockfile needs to be updated, but yarn was run with --frozen-lockfile.`
> 
> Can you please run CI again?

Re-running CI. Sometimes the npm modules gets in a weird state - try running: `bazel clean --expunge && rm -rf node_modules` and see if that fixes the yarn error.

wwwillchen:
I've been thinking about this some more and I think it may be better to just have bokeh be an example in the demo gallery rather than its own component because the implementation is fairly simple:

```py
  from bokeh.embed import file_html

  html(file_html(figure, "inline"), style=style)
```

Arguably the existing `plot` component (for matplotlib) shouldn't be a component and I think it was kind of borderline then, but matplotlib is the most popular Python package for plotting: https://pypistats.org/packages/matplotlib, so that was my original rationale for including it in Mesop.

ChemicalFC:
This approach has not been working for me, also not with plotly's own generated html file.
I do not get anything displayed (see image below).

Screenshot 2024-07-04 113341

**Would that issue be solved by the (already done in code) addition of "sandboxed" mode in the html component (hopefully coming soon :))?**

> I've been thinking about this some more and I think it may be better to just have bokeh be an example in the demo gallery rather than its own component because the implementation is fairly simple:
> 
> ```python
>   from bokeh.embed import file_html
> 
>   html(file_html(figure, "inline"), style=style)
> ```
> 
> Arguably the existing `plot` component (for matplotlib) shouldn't be a component and I think it was kind of borderline then, but matplotlib is the most popular Python package for plotting: https://pypistats.org/packages/matplotlib, so that was my original rationale for including it in Mesop.

wwwillchen:
> This approach has not been working for me, also not with plotly's own generated html file. Both make heavy use of Javascript. I do not get anything displayed (see image below). I am assuming that this has to do with the javascript since i cannot see the javascript when looking at the page using dev tools in chrome.
> 
> Screenshot 2024-07-04 113341
> 
> This is the code that ends up being displayed:
> 
> ```
> <div style="background: lightgrey; border-width: 0px; border-style: initial; border-color: initial; height: 33%; width: 100%;">Bokeh Application
> <div></div>
> </div>
> ```
> 
> **Would that issue be solved by the (already done in code) addition of "sandboxed" mode in the html component (hopefully coming soon :))?**
> 
> > I've been thinking about this some more and I think it may be better to just have bokeh be an example in the demo gallery rather than its own component because the implementation is fairly simple:
> > ```python
> >   from bokeh.embed import file_html
> > 
> >   html(file_html(figure, "inline"), style=style)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Arguably the existing `plot` component (for matplotlib) shouldn't be a component and I think it was kind of borderline then, but matplotlib is the most popular Python package for plotting: https://pypistats.org/packages/matplotlib, so that was my original rationale for including it in Mesop.

hey @ChemicalFC we released [v0.9.4](https://github.com/google/mesop/releases/tag/v0.9.4) yesterday which includes the HTML sandboxed mode - please give that a shot! If you run into more issues, please feel free to open a new issue. I'll close this PR for now.

ChemicalFC:
Tested and worked with bokeh and plotly (via html code)

PR #432, opened 2024-06-13
use re to convert UpperCamelCase to snake_case

djm93dev:
Hi, I've made a slight change to use re for converting UpperCamelCase to snake_case.

### DIFF
--- a/mesop/utils/str_utils.py
+++ b/mesop/utils/str_utils.py
@@ -1,10 +1,7 @@
+import re
+
 def snake_case(str: str) -> str:
   """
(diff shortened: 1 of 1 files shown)


wwwillchen:
Personally, I find this harder to read even though it's terser.

As a general suggestion, I'd recommend contributing to an existing issue, especially those marked as https://github.com/google/mesop/labels/help%20wanted. Although there's plenty of code health issues, I think it'll be more impactful to work on fixing bugs / adding more features / improving docs, rather than rewriting small chunks of code.

PR #457, opened 2024-06-18
Unship editing functionality from dev tools

### LINKED ISSUE #442: Can we upgrade pydantic to a more recent version?
I received the following error due to the fact that mesop seems to rely on a old version of pydantic. Is it possible to upgrade it to a more recent version.

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
mesop 0.8.0 requires pydantic==1.10.13, but you have pydantic 2.7.4 which is incompatible.

Thanks,

wwwillchen:
Closes #442.


mesop/pip_package/requirements.txt
```diff
@@ -1,9 +1,7 @@
 absl-py
-deepdiff==6.7.1
+deepdiff
```
richard-to, line 2: [question] is there a reason to remove the version here? Mainly pinning to downstream. Right now, the latest is 7.0.1 which I'm worried could have breaking changes.

mesop/pip_package/requirements.txt
```diff
@@ -1,9 +1,7 @@
...
 flask
-libcst==1.1.0
 markdown
 protobuf
-pydantic==1.10.13 # --no-binary pydantic
```
richard-to, line 7: Do we still need pydantic in Mesop? In the earlier requirements.txt you left it in. I guess I'm unsure how what's the difference from  requirements.txt in pip_package and the other one.

richard-to (approved):

mesop/pip_package/requirements.txt
wwwillchen: i'll undo this. since there's already a lot of changes in this PR. I think it's generally not a good thing to pin to an exact version in Python because then people will run into version dependency conflicts and not be able to install mesop in their python environment (the ideal solution is to use virt)

mesop/pip_package/requirements.txt
wwwillchen: good catch! I shouldn't have deleted it.

wwwillchen: Re: the difference. This requirement.txt is what's actually used when people do `pip install mesop`. The one in build_defs is what we actually use when compiling Mesop w/ Bazel.

i've filed an issue: https://github.com/google/mesop/issues/467 to refactor this to avoid this duplication.

mesop/utils/validate.py
```diff
@@ -11,19 +11,37 @@
 
 F = TypeVar("F", bound=Callable[..., Any])
 
+pydantic_major_version = int(pydantic.VERSION.split(".")[0])
+
 
 def validate(fn: F) -> F:
-  validated_fn = validate_arguments(fn)
+  # from pydantic import ValidationError, validate_arguments
```
wwwillchen, line 18: remove

PR #469, opened 2024-06-18
Update markdown_editor.py

MeDott29:


richard-to (approved):
Thanks for catching that typo.

PR #470, opened 2024-06-19
Add auth example: firebase auth - Google sign-in

wwwillchen:


richard-to (approved):
Nice. That's a cool use of the web components feature. LGTM.

PR #476, opened 2024-06-19
Prototype tailwind support

wwwillchen:
Few issues:

- This loads a pretty large file from CDN (300kb+) and isn't recommended by Tailwind for production use cases
- Can't load it via ES6 `import` due to CORS issue, this seems like something they could potentially fix
- Not sure whether every component should accept classes
- Potential class name conflicts?

### DIFF
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -322,6 +322,7 @@ class Style:
   bottom: int | str | None = None
   box_shadow: str | None = None
   box_sizing: str | None = None
+  classes: list[str] | None = None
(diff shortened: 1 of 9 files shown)


ducan-ne:
@wwwillchen 
> This loads a pretty large file from CDN (300kb+) and isn't recommended by Tailwind for production use cases
This is fine to me as the page is intended to use in internal services
> Not sure whether every component should accept classes
I only need for Box now, but every component seems cool
>  Potential class name conflicts?
We also have this problem in FE project if not using `tailwind-merge` lib

I've answered some questions

Issue #447, opened 2024-06-17
Select component: support multi select


ghchinoy:
It would be great to support multiple option selections within a Select element.

https://material.angular.io/components/select/examples#select-multiple

**Describe the solution you'd like**

Provide a multiselect option for the component that stores the options as a list, and convenience methods for handling adding, removing of items to the select element's selected options.

**Describe alternatives you've considered**
Checkboxes, but the HTML element supports this so would prefer to use Select

PR #481, opened 2024-06-19
Allow multiple selections on select box

richard-to:
Ref #447

### DIFF
--- a/mesop/components/select/select.py
+++ b/mesop/components/select/select.py
@@ -42,7 +42,7 @@ class SelectSelectionChangeEvent(MesopEvent):
       key (str): Key of the component that emitted this event.
   """
 
-  values: str
+  values: list[str]
 
   @property
   def value(self):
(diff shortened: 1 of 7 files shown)


mesop/components/select/select.py
```diff
@@ -35,23 +35,33 @@ class SelectOpenedChangeEvent(MesopEvent):
 
 @dataclass(kw_only=True)
 class SelectSelectionChangeEvent(MesopEvent):
-  """Event representing a change in the select component's value.
+  """Event representing a change in the select component's value(s).
 
   Attributes:
```
wwwillchen, line 40: Just double checking: are the pydocs generated correctly for `value`?

wwwillchen (approved):
awesome - thanks!

mesop/components/select/select.py
richard-to: Good call. I didn't consider how the property would be rendered. Yeah looks like it.

Screenshot 2024-06-19 at 5 10 59 PM

Issue #468, opened 2024-06-18
Unpin deepdiff


wwwillchen:
Try unpinning deepdiff and see if the tests still pass (or make it a more flexible range <7)

richard-to:
Tests look like they are passing. And looking at the changelog, I'm not seeing any mentions of breaking changes.

https://zepworks.com/deepdiff/current/changelog.html

PR #482, opened 2024-06-19
Unpin deepdiff

richard-to:
Does not look like there any breaking changes: https://zepworks.com/deepdiff/current/changelog.html

Ref #468

### DIFF
--- a/build_defs/requirements.txt
+++ b/build_defs/requirements.txt
@@ -1,10 +1,10 @@
 flask
 absl-py
+deepdiff
 markdown
 protobuf
 pydantic
 pygments
-deepdiff==6.7.1
 
 # Deps for dev/testing:
 mypy-protobuf
--- a/build_defs/requirements.txt
+++ b/build_defs/requirements.txt
@@ -1,6 +1,6 @@
 flask
 absl-py
-deepdiff
+deepdiff==6.*
 markdown
 protobuf
 pydantic
--- a/mesop/pip_package/requirements.txt
+++ b/mesop/pip_package/requirements.txt
@@ -1,5 +1,5 @@
 absl-py
-deepdiff==6.7.1
+deepdiff
 flask
 markdown
 protobuf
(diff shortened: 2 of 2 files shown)


wwwillchen (commented):
You need to run:

 $ bazel run //build_defs:pip_requirements.update

I forget this too :(

Otherwise, it's just not actually changing the dependency version used (which is from requirements_lock.txt)

https://google.github.io/mesop/internal/development/#third-party-packages-pip

richard-to:
> You need to run:
> 
> $ bazel run //build_defs:pip_requirements.update
> 
> I forget this too :(
> 
> Otherwise, it's just not actually changing the dependency version used (which is from requirements_lock.txt)
> 
> https://google.github.io/mesop/internal/development/#third-party-packages-pip

Good call there. The interesting is that I had run the command. So I ran it again. The locked requirements didn't update for some reason.  I had to manually remove the entry from the lock file.

Turns out version 7 does break something. I do remember now that the diff output is a bit different. I remember it being more verbose. Though it's possible if we need to we could reconcile the formats if absolutely necessary.

But for now I just pinned to any version of 6 where 6.7.1 is the last version of 6.

wwwillchen:
> > You need to run:
> > $ bazel run //build_defs:pip_requirements.update
> > I forget this too :(
> > Otherwise, it's just not actually changing the dependency version used (which is from requirements_lock.txt)
> > https://google.github.io/mesop/internal/development/#third-party-packages-pip
> 
> Good call there. The interesting thing is that I had run the command. So I ran it again. The locked requirements didn't update for some reason. I had to manually remove the entry from the lock file.
> 
> Turns out version 7 does break something. I do remember now that the diff output is a bit different. I remember it being more verbose. Though it's possible if we need to we could reconcile the formats if absolutely necessary.
> 
> But for now I just pinned to any version of 6 where 6.7.1 is the last version of 6.

yeah, i've noticed that too. I'm unclear whether that's WAI or a bug.

wwwillchen (approved):

PR #487, opened 2024-06-20
chore: fix typo in docs/index.md

simkimsia:
The link to components is broken.

It should be `./components` instead of `./guides/components`


richard-to:
Thanks for the link fix, KimSia.

PR #495, opened 2024-06-21
me.navigate fixed

### LINKED ISSUE #403: `me.navigate` doesn't work for absolute URLs
**Describe the bug**
`me.navigate` only works for relative paths (e.g. other pages of the Mesop app), but not absolute paths.

**To Reproduce**

```py
@me.page(path="/navigate_advanced/page1")
def page1():
  me.text("page1")
  me.button("navigate", on_click=navigate)

def navigate(e: me.ClickEvent):
  me.navigate("https://github.com/google/mesop")
```

Mesop shows an error saying that the path "https://github.com/google/mesop" does not exist.

**Expected behavior**
Mesop should navigate.

**Additional context**
The fix should be updating https://github.com/google/mesop/blob/d41a05857110a92e9bb01fb97f72a5739284978e/mesop/web/src/shell/shell.ts#L87C25-L87C33 so that it doesn't use the Angular router for absolute path navigations.

kulkamKakarot:
Fixes #403 

- added a new private method navigate(url: string) to the Shell class. This method checks if the URL is absolute and uses the appropriate navigation method.
- added a private method isAbsoluteUrl(url: string) to check if a given URL is absolute.
- In the onCommand method, we've replaced the direct call to this.router.navigateByUrl() with a call to our new navigate() method.


wwwillchen:
Looks like the presubmit failed. Please setup your dev env, particularly the pre-commit hooks: https://google.github.io/mesop/internal/development/#commit-hooks - thanks.

kulkamKakarot:
i have addded pre-commit hook now!

wwwillchen:
I patched this locally and I ran into an error:
image

Based on a quick look, I think the problem is in server.py we run the render loop and assume that the path is a registered mesop path. If we change L158, so that it doesn't change the path if command.navigate.url is an absolute url (similar logic what you have in TS), then it should work:
https://github.com/google/mesop/blob/c6b2323a9dc98f6c21894dc863b0e6a0abb327da/mesop/server/server.py#L158

Can you also add an e2e test? https://google.github.io/mesop/internal/testing/#e2e-tests It'll prevent regressions. Thanks.

kulkamKakarot:
Okay

RogueJin:
any updates on this PR? we are facing the same issue and waiting for this PR as well.

kulkamKakarot:
> any updates on this PR? we are facing the same issue and waiting for this PR as well.

I have fixed the error in server.py but didn't generate e2e test, will generate it.

wwwillchen:
Thanks for opening this PR, I've used it and written some tests here: https://github.com/google/mesop/pull/748
I'll go ahead and close this. Thanks.

PR #497, opened 2024-06-22
Use ipv4 host if ipv6 is not available

### LINKED ISSUE #491: Support ipv4 & ipv6
Looks like my change in #91 has broken users in ipv4 environments, see: https://github.com/google/mesop/pull/91#issuecomment-2179577803

wwwillchen:
Fixes #491.

### DIFF
--- a/mesop/utils/host_util.py
+++ b/mesop/utils/host_util.py
@@ -3,9 +3,38 @@ import socket
 
 def get_default_host() -> str:
   """
-  This function returns the default host (which is externally accessible)
+  Returns the default host (which is externally accessible)
   based on the availability of IPv6.
+
+  See: https://github.com/urllib3/urllib3/pull/611
   """
-  if socket.has_ipv6:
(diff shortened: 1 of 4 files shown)


mesop/utils/host_util.py
```diff
@@ -0,0 +1,11 @@
+import socket
+
+
+def get_default_host() -> str:
+  """
+  This function returns the default host (which is externally accessible)
+  based on the availability of IPv6.
+  """
+  if socket.has_ipv6:
```
richard-to, line 9: I was reading through https://stackoverflow.com/questions/66246308/detect-if-ipv6-is-supported-os-agnostic-no-external-program and also this link https://github.com/urllib3/urllib3/pull/611 from the stackoverflow.

Is this an issue for our use case?

wwwillchen: good catch. fixed. PTAL.

richard-to (approved):

PR #502, opened 2024-06-22
chore: fix typo in posts/why_mesop.md

simkimsia:
similar to #487

The link to components is broken.

It should be `./components` instead of `./guides/components`


richard-to (approved):
Thanks again for finding these broken links. LGTM.

PR #504, opened 2024-06-23
add more border styles to BorderSide style element

seekerquest:
Add dashed", "dotted", "double", "groove", "ridge", "inset", "outset" to BorderSide class

### DIFF
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -73,12 +73,14 @@ class BorderSide:
                                 a string with a unit (e.g., '2em'), or None for no width.
       color: The color of the border, represented as a string. This can be any valid CSS color value,
                           or None for no color.
-      style: The style of the border, which can be 'none' for no border, 'solid' for a solid line.
+      style: The style of the border, which can be 'none' for no border, 'solid' for a solid line, 'dashed' for dashed line,
+              'dotted' for dotted line, 'double' for double line, 'groove' for 3D grooved border, 'ridge' for 3D ridged border,
+              'inset' for 3D inset border, 'outset' for 3D outset border.
   """
 
   width: int | str | None = None
   color: str | None = None
-  style: Literal["none", "solid"] | None = None
+  style: Literal["none", "solid", "dashed", "dotted", "double", "groove", "ridge", "inset", "outset"] | None = None
 
 
 @dataclass(kw_only=True)
--- a/mesop/component_helpers/style.py
+++ b/mesop/component_helpers/style.py
@@ -80,18 +80,21 @@ class BorderSide:
 
   width: int | str | None = None
   color: str | None = None
-  style: Literal[
-    "none", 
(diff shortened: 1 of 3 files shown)


mesop/component_helpers/style.py
```diff
@@ -73,12 +73,14 @@ class BorderSide:
                                 a string with a unit (e.g., '2em'), or None for no width.
       color: The color of the border, represented as a string. This can be any valid CSS color value,
                           or None for no color.
-      style: The style of the border, which can be 'none' for no border, 'solid' for a solid line.
+      style: The style of the border, which can be 'none' for no border, 'solid' for a solid line, 'dashed' for dashed line,
```
richard-to, line 76: Maybe you can link to this as well: https://developer.mozilla.org/en-US/docs/Web/CSS/border-style

mesop/component_helpers/style.py
```diff
@@ -73,12 +73,14 @@ class BorderSide:
...
+              'dotted' for dotted line, 'double' for double line, 'groove' for 3D grooved border, 'ridge' for 3D ridged border,
+              'inset' for 3D inset border, 'outset' for 3D outset border.
   """
 
   width: int | str | None = None
   color: str | None = None
-  style: Literal["none", "solid"] | None = None
+  style: Literal["none", "solid", "dashed", "dotted", "double", "groove", "ridge", "inset", "outset"] | None = None
```
richard-to, line 83: Seems like there is also a "hidden" style, which seems slightly different than none. So we may want to add it for completeness.

richard-to (changes requested):
Can you also update "box" demo example to use some of the added styles? That's just to make sure we have an example of it working. See: https://github.com/google/mesop/blob/main/demo/box.py

--------

Also looks like someone else also posted a pull request for the border side change after you: https://github.com/google/mesop/pull/509. Looks like what happened is that they didn't see your message. And I had referred that person to this issue from another issue.

Since you were first, I will give your pull request precedence.

richard-to (approved):
LGTM - Thanks

richard-to:
Looks like the pre-submit checks failed. Just some linting stuff. The best option is to set up the pre-commits described here: https://google.github.io/mesop/internal/development/

The precommits should manually run the ruff formatter to fix the formatting.

For a manual run, I think the command is something like:

```
pre-commit run --files FILES
```

seekerquest:
Thanks for sharing the steps to fix the pre-checks. I think this has now passed the pre-checks.

richard-to:
Hmm looks like there is another type checking issue on the demo page. It's an interesting one. I'm not quite sure how best to resolve that type checking issue.

For simplicity, you may just want to remove the loop. Then for the styles, just hardcode to different styles.

richard-to:
Closing this one. We ended up merging a similar pull request in https://github.com/google/mesop/pull/509

PR #517, opened 2024-06-24
Add `on_blur` for input / textarea

### LINKED ISSUE #516: Provide `on_blur` for input/textarea components
Mesop developers oftentimes run into some kind of performance issue with using `on_input` which is triggered very frequently (on almost every key input, with a small debounce timer). By providing `on_blur`, which is only triggered when the input loses focus, they can have a more performant way of getting the input value.

wwwillchen:
Fixes #516.

This allows people to avoid using `on_input` in the common case.

### DIFF
--- a/mesop/components/input/e2e/input_blur_app.py
+++ b/mesop/components/input/e2e/input_blur_app.py
@@ -1,0 +1,32 @@
+import mesop as me
+
+
+@me.stateclass
+class State:
+  input: str = ""
+  input_value_when_button_clicked: str
+
+
+def on_blur(e: me.InputBlurEvent):
+  state = me.state(State)
+  state.input = e.value
+
+
+def click_button(e: me.ClickEvent):
+  state = me.state(State)
+  state.input_value_when_button_clicked = state.input
+
+
+@me.page(path="/components/input/e2e/input_blur_app")
+def app():
+  s = me.state(State)
+  me.input(label="Input", on_blur=on_blur)
+  me.button("button", on_click=click_button)
+  me.text("Input: " + s.input)
+  me.text(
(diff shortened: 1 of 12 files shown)


mesop/components/input/e2e/input_test.ts
```diff
@@ -7,3 +7,34 @@ test('test input interactivity works', async ({page}) => {
...
+  // Same with textarea:
+  await page.getByLabel('Regular textarea').click();
+  await page.getByLabel('Regular textarea').fill('123');
+  await page.getByRole('button', {name: 'button'}).click();
+  await expect(page.getByText('Input: 123')).toBeVisible();
+  await expect(
+    page.getByText('input_value_when_button_clicked: 123'),
+  ).toBeVisible();
+
+  // Same with native tectarea:
```
richard-to, line 32: [typo] textarea

mesop/components/input/e2e/input_blur_app.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def on_input(e: me.InputEvent):
+  state = me.state(State)
+  state.input = e.value
+
+
+def on_enter(e: me.EnterEvent):
+  state = me.state(State)
+  state.input = "boo"
```
richard-to, line 17: [question] What's this boo about? 

The on blur won't work with on enter right?

richard-to (approved):
LGTM

mesop/components/input/e2e/input_blur_app.py
wwwillchen: oops, this was a remnant from input_app.py, removed.

yeah, on_blur and on_enter don't work well, I'm thinking about changing on_enter to also include the input value

Issue #330, opened 2024-05-30
Setting a complex nested data structure in State fails in a mysterious way


wwwillchen:
A downstream user had something like this: list[dict[dict[]]]. This resulted in a hard-to-debug error with state where the state wasn't being properly populated.

We should either show an error right away OR properly support it.

PR #520, opened 2024-06-24
Fix deserializing dict for stateclass

### LINKED ISSUE #518: Dict state not persisting correctly
# Example 

```
import mesop as me


@me.stateclass
class State:
  checks: dict[str, bool]


def on_user_checkbox_change(event: me.CheckboxChangeEvent):
  state = me.state(State)
  print(state.checks)
  state.checks[event.key] = event.checked
  print(state.checks)


@me.page()
def app():
  state = me.state(State)

  me.checkbox(
      label="box1",
      on_change=on_user_checkbox_change,
      key="box1",
  )

  me.checkbox(
      label="box2",
      on_change=on_user_checkbox_change,
      key="box2",
  )

  me.text("Checks: " + str(state.checks.items()))
```

# Current behavior

state.checks keeps resetting to the last clicked check box value 

# Expected behavior

state.checks should accumulate new check box clicks when new keys are added

# Things tested

- Manually disabled state diff (same issue)

# Suspected issue

Most likely a serialization/deserialization issue with dicts. Think it was mentioned that it tries to convert dicts to dataclasses??

wwwillchen:
Fixes #518, #330

### DIFF
--- a/mesop/dataclass_utils/dataclass_utils.py
+++ b/mesop/dataclass_utils/dataclass_utils.py
@@ -103,6 +103,13 @@ def _recursive_update_dataclass_from_json_obj(instance: Any, json_dict: Any):
       else:
         # For other types, set the value directly.
         setattr(instance, key, value)
+    else:
+      if isinstance(instance, dict):
+        instance[key] = value
+      else:
+        raise MesopException(
+          f"Unhandled stateclass deserialization where key={key}, value={value}, instance={instance}"
+        )
   return instance
 
 
--- a/mesop/dataclass_utils/dataclass_utils.py
+++ b/mesop/dataclass_utils/dataclass_utils.py
@@ -89,8 +89,9 @@ def _recursive_update_dataclass_from_json_obj(instance: Any, json_dict: Any):
         updated_list: list[Any] = []
         for item in cast(list[Any], value):
           if isinstance(item, dict):
-            # If the json item value is an instance of dict,
-            # we assume it should be converted into a dataclass
+            # If the json item value is an instance of dict
+            # and the instance has an attribute with a matching name,
+            # we assume the dict should be converted into a dataclass.
             attr = getattr(instance, key)
             item_instance = instance.__annotations__[key].__args__[0]()
             updated_list.append(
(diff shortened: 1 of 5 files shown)


mesop/examples/dict_state.py
```diff
@@ -0,0 +1,30 @@
+import mesop as me
```
richard-to, line 1: [question] Should this be some kind of e2e test rather than an example?

mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -103,6 +103,13 @@ def _recursive_update_dataclass_from_json_obj(instance: Any, json_dict: Any):
       else:
         # For other types, set the value directly.
         setattr(instance, key, value)
+    else:
```
richard-to, line 106: It won't let me comment on line 88, so commenting here.

It's not related to this PR.  But what happens in the nested list case?

mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -103,6 +103,13 @@ def _recursive_update_dataclass_from_json_obj(instance: Any, json_dict: Any):
...
+      if isinstance(instance, dict):
```
richard-to, line 107: Comment for line 92-93 -> Is the comment there still relevant?

richard-to (approved):
LGTM

mesop/examples/dict_state.py
wwwillchen: that's fair, i was being a little lazy :) done.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: added a test case to cover nested list - it works.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: good question - i've updated the comment to clarify the assumption.

wwwillchen:
Thanks for the review.

PR #522, opened 2024-06-24
Create InputEnterEvent which includes input value

wwwillchen:


mesop/components/input/input.py
```diff
@@ -14,15 +14,28 @@
 
 @dataclass(kw_only=True)
 class EnterEvent(MesopEvent):
-  """Represents an "Enter" keyboard event."""
+  """[Deprecated] Represents an "Enter" keyboard event. Use InputEnterEvent instead."""
```
richard-to, line 17: I see that this is deprecated. But seems like you wouldn't be able to actually use it? I thought if it was deprecated, the event would still be usable?  Or maybe I'm misreading the code.

mesop/labs/chat.py
```diff
@@ -169,7 +169,7 @@ def chat(
   """
   state = me.state(State)
 
-  def on_submit(e: me.ClickEvent | me.EnterEvent):
+  def on_submit(e: me.ClickEvent | me.InputEnterEvent):
```
richard-to, line 172: Are you planning to remove the on input handler as a follow up? I think if on input is removed, then InputEnterEvent won't work as is since we'd need to set the input here.

richard-to (approved):
LGTM

mesop/components/input/input.py
wwwillchen: It's kinda confusing. You can use the EnterEvent, but there's no reason to, and you should use the InputEnterEvent instead. The main reason why I left it here is that if I removed it, it would be a breaking change. I think the hope is that people will migrate from EnterEvent to InputEnterEvent and then we can get rid of EnterEvent in a future version of Mesop (in hindsight, it was a mistake to have an event named EnterEvent).

mesop/labs/chat.py
wwwillchen: Yeah, good point, I'll take care of this in a follow-up PR.

PR #525, opened 2024-06-24
added proper example for navigate api

### LINKED ISSUE #524: Navigate api docs have wrong example
In mesop docs navigate api has the example of onload example.

Rakib-hosen:
Fixes #524 .


wwwillchen:
Thanks for fixing this - appreciate it!

PR #531, opened 2024-06-25
Add test for text_to_text

wwwillchen:


richard-to (approved):
LGTM - Were you planning to also update to use on blur? I guess this also applies to text_to_image.

wwwillchen:
> LGTM - Were you planning to also update to use on blur? I guess this also applies to text_to_image.

yup, will do in a follow-up PR.

PR #541, opened 2024-06-25
Add docs VS Code remote containers

richard-to:
Ref https://github.com/google/mesop/issues/101


docs/internal/vs_code_remote_container.md
```diff
@@ -0,0 +1,64 @@
+# VS Code Remote Container
+
+> Git credentials need to be setup in order to commit changes inside the remote
```
wwwillchen, line 3: add a link on how to do this?

wwwillchen (approved):

docs/internal/vs_code_remote_container.md
richard-to: That's a good question. I haven't decided on the best way to do this yet. There's a few ways I tried in the past.

1. Just add credentials in the Docker container at the right location (problem here is if you turn off the container, you lose the credentials

2. Add a bind mount from host to container to the ssh keys. Problem here is we'd need a predetermined folder to make this work inside the mesop repository (which would be git ignored)

3. Use a Docker volume. This will persist after the container is shut down. Then people can generate the ssh key and add the public key to Github.

I think the best option is probably number 3, but I need to update the docker-compose file to include the volume.

I think there could be some other options described in the VS Code docs that I haven't explored before.

I'll make a note to fix this soon.

PR #542, opened 2024-06-26
Create python FE fatigue blog post

wwwillchen:
@richard-to WDYT? :) I'm thinking about publishing this blog post and sharing it on social media 😃

### DIFF
--- a/docs/blog/posts/web_components.md
+++ b/docs/blog/posts/web_components.md
@@ -12,7 +12,7 @@ At first, I thought of it as joke, but now that Mesop has launched experimental 
 
 ## What is Mesop?
 
-Before we dive in, let me explain what Mesop is. Mesop is a Python UI framework focused on rapidly building AI apps. You can write a lot of kinds of apps all in Python as you can see from the examples on our [home page](https://google.github.io/mesop/), but when you need to, Mesop provide the flexibility of dropping down into web components so you can use existing JS libraries or have fine-grained UI control.
+Before we dive in, let me explain what Mesop is. Mesop is a Python UI framework focused on rapidly building AI apps. You can write a lot of kinds of apps all in Python as you can see from the examples on our [home page](https://google.github.io/mesop/), but when you need to, Mesop provides the flexibility of dropping down into web components so you can  have fine-grained UI control and use existing JS libraries.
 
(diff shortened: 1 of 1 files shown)


richard-to (approved):
I think it looks fine for a blog post. It's a good way to get in that we have web component support.

A few thoughts:

1. May want to say that it is still an experimental feature
2. May also want to say that one can write a Mesop app without writing a web component and digging into javascript/html/css
3. I think the web component feature is still a bit early, so we haven't explored enough use cases yet to truly say if we can avoid bundling use cases
  - It does seem useful for cases where there's a javascript library out there that can be imported through a script tag, then the web component is mainly a wrapper over the library, like the Plotly case and firebase auth case we've seen so far
 4. With Web components, there is now the need to learn javascript and html. And most likely they need to learn Lit. So there is still something to learn on the frontend. So can't get away with just Python. But as you say, at least it seems like something that isn't the latest fad. And it's also good that the code is self-contained. Which I think is fine for more advanced cases. Ideally our ecosystem of web components that get shared and reused will grow into a healthy one, thus making it less common for people to reach into the javascript stuff.

wwwillchen:
> I think it looks fine for a blog post. It's a good way to get in that we have web component support.
> 
> A few thoughts:
> 
> 1. May want to say that it is still an experimental feature

Yep, it's mentioned.

> 2. May also want to say that one can write a Mesop app without writing a web component and digging into javascript/html/css
> 3. I think the web component feature is still a bit early, so we haven't explored enough use cases yet to truly say if we can avoid bundling use cases

Softened the phrasing around bundling.
> 
> * It does seem useful for cases where there's a javascript library out there that can be imported through a script tag, then the web component is mainly a wrapper over the library, like the Plotly case and firebase auth case we've seen so far
> 
> 4. With Web components, there is now the need to learn javascript and html. And most likely they need to learn Lit. So there is still something to learn on the frontend. So can't get away with just Python. But as you say, at least it seems like something that isn't the latest fad. And it's also good that the code is self-contained. Which I think is fine for more advanced cases. Ideally our ecosystem of web components that get shared and reused will grow into a healthy one, thus making it less common for people to reach into the javascript stuff.

PR #544, opened 2024-06-26
Add sandboxed mode to HTML component

### LINKED ISSUE #156: Add a component to directly embed html

wwwillchen:
**Approach:** This iframes HTML content (potentially from end-user input) in a sandboxed iframe _without_ "allow-same-origin" to ensure the content does not run on the same origin as the main Mesop app.

This uses a double-iframing approach because we cannot directly use `<iframe srcdoc="${html}">` from the main Mesop app because Mesop by default applies a stringent CSP (e.g. trusted types) and iframe documents using srcdoc automatically inherit their parent's CSP and this makes the iframe basically unusable for many use cases.

Instead we iframe `sandboxed_iframe.html` which is served with a minimal CSP that only prevents being iframed by other origins besides the Mesop app and then within `sandboxed_iframe.html` we render the (untrusted) HTML content inside a sandboxed iframe.

Fixes #156.


mesop/components/html/html.py
```diff
@@ -6,23 +8,31 @@
...
+  This function renders custom HTML in a secure way.
 
   Args:
       html: The HTML content to be rendered.
+      mode: The mode of the iframe. Can be either "sanitized" or "sandboxed". If "sanitized" then potentially dangerous content like `<script>` and `<style>` are
+          stripped out. If "sandboxed", then all content is allowed, but rendered in an iframe for isolation.
       style: The style to apply to the embed, such as width and height.
       key: The component [key](../components/index.md#component-key).
   """
+  if mode != "sandboxed" and ("<script>" in html or "<style>" in html):
```
wwwillchen, line 32: @richard-to WDYT about having a default mode (where mode is not specified) where it uses sandboxed if it detects a <script> or <style> and then sanitized otherwise? I think this would produce the right behavior in _most_ circumstances, although it's definitely possible it'll over-trigger for sandboxed (which isn't the worst thing). Or do you think that is too magical?

mesop/components/html/html.ts
```diff
@@ -16,12 +17,52 @@ export class HtmlComponent {
...
+      if (this.isSandboxed) {
+        console.warn(
+          'iframe element in Mesop html component unexpectedly not found',
+        );
+      }
+      return;
+    }
+    const iframe = this.iframe.nativeElement as HTMLIFrameElement;
+    setIframeSrc(iframe, '/sandbox_iframe.html', {allowSameOrigin: false});
+    iframe.onload = () => {
```
wwwillchen, line 60: @richard-to wondering whether if I should automatically resize the iframe if the style height is not set (vs. the browser default which is 400px for iframe height, I think). I think most people would want/expect the html component to display all the contents and not have a scrollbar. WDYT?

mesop/components/html/html.ts
```diff
@@ -16,12 +17,52 @@ export class HtmlComponent {
...
+    if (!this.iframe) {
+      if (this.isSandboxed) {
+        console.warn(
+          'iframe element in Mesop html component unexpectedly not found',
+        );
+      }
+      return;
+    }
+    const iframe = this.iframe.nativeElement as HTMLIFrameElement;
+    setIframeSrc(iframe, '/sandbox_iframe.html', {allowSameOrigin: false});
```
richard-to, line 59: [question] What's the reason for setting `allowSameOrigin` false in this case but not for embed? My assumption is that the sandbox file is on the Mesop server so we don't want it accessing things like cookies especially if the content will be dynamically generated based on user input. And I guess with the embed case, we assume the embed will typically be from a different origin and that site may need access to the cookies of their origin?

[question] I remember you mentioning before about using src doc attribute. But here I see you're loading this sandbox html file and then updating the HTML through the message event. I'm assuming this is a safer approach?

mesop/web/src/safe_iframe/safe_iframe.ts
```diff
@@ -5,28 +5,38 @@
...
-  setIframeSrcImpl(iframe, src);
+  setIframeSrcImpl(iframe, src, options);
 }
 
 // copybara:strip_begin(external-only)
-function setIframeSrcImpl(iframe: HTMLIFrameElement, src: string) {
+function setIframeSrcImpl(
+  iframe: HTMLIFrameElement,
+  src: string,
+  options: {allowSameOrigin: boolean},
```
richard-to, line 22: Maybe add a comment about allowSameOrigin being an option. It was not immediately clear to me the reasoning.

richard-to (approved):
LGTM - I like the api here. I think better than the alternative of having the embed element accept html.

mesop/components/html/html.ts
richard-to: Yeah, if you're able to render the iframe to the correct size of the content without a scrollbar, I would agree that makes a lot of sense since one would expect it to work as if they were rendering HTML. So would be slightly annoying to see it rendered to 400px.

mesop/components/html/html.py
richard-to: Yeah that's a good question. Personally I like the more explicit modes here since it requires the developer to make the choice for what they intend. So hopefully they would know the pros/cons of their choice.

But I also agree that there is the edge case where someone has like some text that says "<script>" somewhere that could cause the sandbox to trigger. Seems rare and could be preventable with a more robust check for script and style tags.

In the end, I don't feel strongly one way or the other here. Both seem fine, especially if the iframe can be rendered to the correct height automatically.

mesop/components/html/html.ts
wwwillchen: done - I've automatically resized the iframe

mesop/components/html/html.py
wwwillchen: OK, I've done the implicit mode since I've automatically resized the iframe.

mesop/components/html/html.ts
wwwillchen: Good questions. It's pretty complicated and I left detailed comments both here and in embed.

For this case (sandbox_iframe.html) it's important that we don't allow it to be on the same origin because user inputs are being executed inside that document with no sanitization, e.g. imagine a callsite like:

`me.html(me.state(State).input, mode="sandboxed")` in this case, we don't want to execute end-user input text which could be something like (go send an XHR to some other server with sensitive data), esp. in the case where an end-user could be tricked to put something malicious.

Right, for the embed case, it's usually a different origin, and many sites won't behave properly if you don't let it run on its original origin.

---

I initially tried using srcdoc however the problem is that it inherits the CSP of the main document (index.html) which has a very stringent CSP and would basically make this feature unusable. By using this sandbox_iframe.html shim layer, I can set a much more lax CSP (with the assumption that this page is carefully sandboxed when iframed).

mesop/web/src/safe_iframe/safe_iframe.ts
wwwillchen: yep , i've added comments in this file and at each callsite.

wwwillchen:
Need to hold off, there's a security issue with this approach.

mesop/web/src/safe_iframe/safe_iframe.ts
wwwillchen: always allowing same origin since sandbox_iframe.html now renders all content itself in its own child sandboxed iframe that does _not_ have allow-same-origin

wwwillchen:
> Need to hold off, there's a security issue with this approach.

Fixed now.

PR #548, opened 2024-06-27
Update docs with info about state sessions

richard-to:
- Adds config section to describe how to set memory backend
- Update state management section to reference memory backend
- Update deployment section to include session affinity note

Ref #62


wwwillchen (approved):
awesome ,thanks!

PR #567, opened 2024-07-01
feat: adds svg_icon fixes #529

### LINKED ISSUE #529: Provide support for SVG xml in icon
**Is your feature request related to a problem? Please describe.**
It would be great to support custom svg xml as an input for the icon component

**Describe the solution you'd like**
extend icon to allow in-lining (or file path input reference) for svg xml

https://material.angular.io/components/icon/overview#svg-icons

**Describe alternatives you've considered**
Haven't tried any other way to add svg to mesop

ghchinoy:
Adds an example of a web component `svg_icon` that takes an SVG XML and renders it


wwwillchen (commented):
Thanks for contributing!

Could you create an "svg_icon_app.py" to demonstrate how to use this web component? e.g. we do this for our other web component examples: https://github.com/google/mesop/blob/main/mesop/examples/web_component/quickstart/counter_component_app.py

And then you can import it here: https://github.com/google/mesop/blob/main/mesop/examples/web_component/__init__.py and then this will load the app into the main Mesop example app, e.g. when you run ./scripts/cli.sh

wwwillchen:
I think there's some formatting issues which could be solved by installing our pre-commit hooks: https://google.github.io/mesop/internal/development/#commit-hooks

wwwillchen:
Thanks for contributing!

Issue #546, opened 2024-06-26
The `me.markdown()` component can't render the output of `pd.DataFrame().to_markdown()`


aperepel:
**Describe the bug**
The dataframe markdown is not recognized by the renderer, simply prints the markup (unreadable, no pre-formatting)

**To Reproduce**
```python
import pandas as pd
 
df = pd.DataFrame(
data={
    "NA": [pd.NA, pd.NA, pd.NA],
    "Index": [3, 2, 1],
    "Bools": [True, False, (True)],
    "Ints": [101, 90, -55],
    "Floats": [2.3, 4.5, -3.000000003],
    "Strings": ["Hello", "World", "!"],
    "Date Times": [
      pd.Timestamp("20180310"),
      pd.Timestamp("20230310"),
      pd.Timestamp("20230310"),
    ],
  }
)

me.markdown(df.to_markdown())
```

**Expected behavior**
Renders a visually appealing and formatted data table.

richard-to:
This looks like a simple bug to fix.

We just need to enable the extension: https://python-markdown.github.io/extensions/tables/

Just need to add the extension here I think: https://github.com/google/mesop/blob/412cd8e21493cc5ec5656ea81592e4255483c018/mesop/components/markdown/markdown.py#L30

richard-to:
Ok, the fix is in for the markdown formatting. The big caveat here is that the default styling of the table isn't that nice. 

There is a hacky way around this by including a [custom stylesheet](https://google.github.io/mesop/api/page/#mesop.features.page.page) and using the markdown [attr_list](https://python-markdown.github.io/extensions/attr_list/)

Just note that the custom stylesheet would need to be hosted outside of the Mesop app:

- CDN
- GCS / S3
- You could do the same server / domain if you had something like nginx to route to the files and to mesop, but that's a more advanced set up

PR #569, opened 2024-07-02
Enable markdown tables + attribute lists

richard-to:
This change allows markdown tables to be rendered correctly. One caveat is that custom styling is not possible unless someone includes an custom stylesheet via CDN or external URL.

This means that the default rendering is very basic.

If a user wants to use render markdown from a pandas data frame, they will need to install tabulate separately.

Attribute lists have also been enabled to make it easier to style the the tables via CSS.
[
Screenshot 2024-07-01 at 5 26 28 PM
](url)

Ref #546

### DIFF
--- a/mesop/components/markdown/markdown.py
+++ b/mesop/components/markdown/markdown.py
@@ -28,6 +28,8 @@ def markdown(
     html = markdown_lib.markdown(
       text,
       extensions=[
+        "attr_list",
+        "tables",
         CodeHiliteExtension(css_class="highlight"),
         "fenced_code",
       ],
--- a/demo/markdown_demo.py
+++ b/demo/markdown_demo.py
@@ -50,8 +50,8 @@ Inline `code`
 
 First Header  | Second Header
 ------------- | -------------
-Content Cell  | Content Cell
-Content Cell  | Content Cell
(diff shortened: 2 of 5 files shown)


mesop/components/markdown/markdown.py
```diff
@@ -28,6 +28,8 @@ def markdown(
     html = markdown_lib.markdown(
       text,
       extensions=[
+        "attr_list",
```
wwwillchen, line 31: Can you add an example of where you set an attribute? specifically, i'm wondering if you set the style attribute whether Angular will sanitize it out (which is OK, but then I'm not sure how helpful it is to have an attribute list here).

richard-to: Good point. I tested it out just now. Seems to work. Updated the code with an example.

wwwillchen: I guess there's no way to do inline style (e.g. `style="background: red"`)? If users set CSS class names, it's not very straight forward to define CSS class rules (I think you would need to [inject a stylesheet](https://google.github.io/mesop/api/page/#mesop.features.page.page:~:text=DEFAULT%3A%20None-,stylesheets,-List%20of%20stylesheet))

wwwillchen (approved):
LGTM

Issue #568, opened 2024-07-01
Uploader Component Parameters - Style (and others)


wwwillchen:
### Discussed in https://github.com/google/mesop/discussions/559

<div type='discussions-op-text'>

<sup>Originally posted by **lnarimatsu** June 28, 2024</sup>
“Could the Uploader Button be enhanced by adding a **style parameter**? This would allow us to customize its appearance. 

Additionally, could we consider incorporating all parameters that are currently available in the button component? This would provide greater flexibility and consistency across our components.

_Func Uploder_
![image](https://github.com/google/mesop/assets/11890782/68823a87-2557-4045-91f3-ed328695ad28)

_Func Button_
![image](https://github.com/google/mesop/assets/11890782/58e1e74b-efae-4714-b20e-6a2598464368)
</div>

---

@richard-to any thoughts?

richard-to:
Yeah I think that's a good idea, especially being able to style the button properly. I think the main thing we can leave off is on click. Should be fairly straightforward.

PR #570, opened 2024-07-02
Add button attributes to uploader

richard-to:
To make the experience more consistent, this change copies the same parameters from button to uploader. This will allow users to customize the button colors / type / style and disable it.

Ref #568

### DIFF
--- a/mesop/components/uploader/uploader.py
+++ b/mesop/components/uploader/uploader.py
@@ -1,9 +1,10 @@
 import io
 from dataclasses import dataclass
-from typing import Any, Callable, Sequence
+from typing import Any, Callable, Literal, Sequence
 
 import mesop.components.uploader.uploader_pb2 as uploader_pb
 from mesop.component_helpers import (
+  Style,
   insert_component,
   register_event_handler,
   register_event_mapper,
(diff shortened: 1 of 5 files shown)


mesop/components/uploader/uploader.py
```diff
@@ -92,5 +104,27 @@ def uploader(
...
+      color=color,
+      disable_ripple=disable_ripple,
+      disabled=disabled,
     ),
+    style=style,
   )
+
+
+def _get_type_index(
+  type: Literal["raised", "flat", "stroked", "icon"] | None,
```
wwwillchen, line 118: Drop icon here and  on L128. You removed it in L76, which I think is right because icon button needs to have a child component, and this seems overtly complex for uploader right now.

wwwillchen (approved):

mesop/components/uploader/uploader.py
richard-to: Good catch. Updated. 

I think I didn't do it on purpose. I just copied what was on button. So button is also the same where the interface does not include "icon" but the underlying code still has the icon option.

Issue #556, opened 2024-06-29
Add Support for Additional State Session Backends


richard-to:
Right now the state session backend feature only supports memory. Memory is not the best choice most of the time.

Completed:

- File (DONE)
- GCP Firestore (has free tier) (DONE)
- SQLAlchemy (Postgres, SQlite, etc, GCP Cloud Cloud SQL) (DONE)

Other considerations if there is interest:

- Redis (Valkey fork of Redis, GCP Memorystore)
- Memcached (GCP Memorystore)
- GCP Cloud Storage
- Other popular key/value store?

Ref: https://google.github.io/mesop/api/config/#mesop_state_session_backend

wwwillchen:
@richard-to I think we need some kind of guide. Technically all the information is in https://google.github.io/mesop/api/config/ but it's not obvious to look there. I think having some kind of "Performance" guide would be very helpful.

Something like:
- How do I diagnose my performance problem? e.g. is it because state is too large or some other reason
- If state is a problem, how do I pick which state session backend is suitable?

richard-to:
Going to close this for now. We can create new issues for additional backends if there is interest.

PR #574, opened 2024-07-03
Add file backend for state sessions

richard-to:
This change also adds a new dependency with msgpack. For now we use it to serialize/deserialize the JSON strings of the Mesop state. In the future we will want to serialize everything with msgpack.

The reason we can't do this now is because we will need to add msgpack encoding/decoding rules to support non-standard data types.

This change also adds a new config variable called MESOP_STATE_SESSION_BACKEND_FILE_BASE_DIR which allows the user to specify the location where the state sessions get written.

Ref #556

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -23,7 +23,7 @@ own RAM, which means cache misses will be common if each server has multiple pro
 and there is no session affinity. In addition, the amount of RAM must be carefully
 specified per instance in accordance with the expected user traffic and state size.
 
-The safest option for using the `memory` backend is to use a single a process with a
+The safest option for using the `memory` backend is to use a single process with a
 good amount of RAM. Python is not the most memory efficient, especially when saving data
 structures such as dicts.
 
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -14,25 +14,52 @@ Sets the backend to use for caching state data server-side. This makes it so sta
 not have to be sent to the server on every request, reducing bandwidth, especially if
 you have large state objects.
 
-The only backend option available at the moment is `memory`.
+The backend options available at the moment are `memory` and `file`.
+
+#### memory
 
 Users should be careful when using the `memory` backend. Each Mesop process has their
-own RAM, which means cache misses will be common if there is no session affinity. In
-addition, the amount of RAM must be carefully specified per machine in accordance with
-the expected user traffic and state size.
+own RAM, which means cache misses will be common if each server has multiple processes
+and there is no session affinity. In addition, the amount of RAM must be carefully
+specified per instance in accordance with the expected user traffic and state size.
 
 The safest option for using the `memory` backend is to use a single a process with a
-good amount of RAM. Python is not the most memory efficient, especially data structures
-such as dicts.
+good amount of RAM. Python is not the most memory efficient, especially when saving data
+structures such as dicts.
 
 The drawback of being limited to a single process is that requests will take longer to
 process since only one request can be handled at a time. This is especially problematic
 if your application contains long running API calls.
 
-If session affinity is available, you can scale up multiple machines, each running
+If session affinity is available, you can scale up multiple instances, each running
 single processes.
 
+#### file
+
+Users should be careful when using the `file` backend. Each Mesop instance has their
+own disk, which can be shared among multiple processes. This means cache misses will be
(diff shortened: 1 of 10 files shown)


docs/api/config.md
```diff
@@ -14,26 +14,53 @@ Sets the backend to use for caching state data server-side. This makes it so sta
...
 
 Users should be careful when using the `memory` backend. Each Mesop process has their
-own RAM, which means cache misses will be common if there is no session affinity. In
-addition, the amount of RAM must be carefully specified per machine in accordance with
-the expected user traffic and state size.
+own RAM, which means cache misses will be common if each server has multiple processes
+and there is no session affinity. In addition, the amount of RAM must be carefully
+specified per instance in accordance with the expected user traffic and state size.
 
 The safest option for using the `memory` backend is to use a single a process with a
```
wwwillchen, line 26: remove  extra "a" between single and process

mesop/server/state_session.py
```diff
@@ -80,12 +95,100 @@ def clear_stale_sessions(self):
...
+        logging.warning("Failed to delete file after restore - " + str(e))
+
+  def save(self, token: str, states: States):
+    """Saves state to the backend with the given token."""
+    with open(self._make_file_path(token), "wb") as f:
+      f.write(msgpack.packb(_serialize_state(states)))  # type: ignore
+
+  def clear_stale_sessions(self):
+    """Deletes unused state data."""
+    if random.random() > self._SESSION_CLEAR_RATE:
```
wwwillchen, line 152: It seems a little weird to use a probabilistic way of cleaning up. Why not use a time elapsed approach (e.g. if the last cleanup was more than 10 minutes ago, i.e. _SESSION_TTL_MINUTES, then do the cleanup)?

mesop/server/state_session.py
```diff
@@ -80,12 +95,100 @@ def clear_stale_sessions(self):
...
+    if base_dir is None:
+      raise RuntimeError(
+        "Base dir must be set when using `FileStateSessionBackend`"
+      )
+    self.base_dir = base_dir
+    self.prefix = "session"
+
+  def _make_file_path(self, token: str) -> Path:
+    return self.base_dir / (self.prefix + token)
+
```
wwwillchen, line 129: woah, never knew this was a thing!

wwwillchen (approved):
Another exciting change!

I think we could use this w/ Cloud Run+GCS using https://cloud.google.com/run/docs/configuring/services/cloud-storage-volume-mounts

This could become a very nice option that's: a) scalable (you will basically not run out of GCS space), b) [cheap](https://cloud.google.com/storage/pricing) and c) fast - according to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/benchmarks.md#results

mesop/server/state_session.py
richard-to: Yes, I mainly did this in the case that there are multiple mesop processes running, so we'd need some kind of global state across the processes. So it was just easier to randomly clear the data.

The alternative I considered was writing a sort of lock file. So basically if the file exists and it's not X minutes old, then don't delete anything.

Let me know if you'd prefer the lock file approach or if you have a better idea.

richard-to:
> This could become a very nice option that's: a) scalable (you will basically not run out of GCS space), b) [cheap](https://cloud.google.com/storage/pricing) and c) fast - according to https://github.com/GoogleCloudPlatform/gcsfuse/blob/master/docs/benchmarks.md#results

Ah nice! That's a great find.

mesop/server/state_session.py
wwwillchen: I like your idea of doing a quasi-lock file - it seems fairly straightforward and even if there's some weird race conditions (e.g. two processes see the file is > X minute and tries to do the cleanup at the same time, it doesn't seem that bad to me, it could cause contention in the filesystem / raise exceptions, but it should basically still do the clean-up).

I was thinking about you could do a proper lockfile but you would need to use the fcntl module for file locking on Unix-based systems, or the msvcrt module on Windows; and this could have tricky edge cases (e.g. a process creates a lock-file and then dies, i.e. a stale lock).

richard-to: Ok. I implemented the quasi-lock file for now. We can update to something more robust if necessary.

Issue #338, opened 2024-06-01
Create patterns examples


wwwillchen:
_This is a good issue for anyone who wants to contribute to Mesop as there's many different patterns we could add to the demo gallery!_

I'm basically calling "pattern" a multi-component examples which demonstrates a common use case.

Ideas:
- Submitting feedback (thumbs up / thumbs down + reasons)
- SxS output rating
- Nav menu
- Header (text)
- Card
- Tile
- Grid tile
- Carousel

richard-to:
Also things like Modal, Accordion, Tab, and Toasts could be useful (until we have the Angular Component equivalents created - Dialog, Tab, Snackbar, etc)

iiakshat:
I was working on the ```submitting feedback``` feature and am unsure if this is how we want to implement it.


![image](https://github.com/google/mesop/assets/92530735/9c3384c0-69a5-4bf3-9dc5-cb8d24d78535)

 If so, I am getting this `error`,
![image](https://github.com/google/mesop/assets/92530735/857f7152-178c-454f-8cd3-e2e16612d658)

Can anyone explain where did I go wrong? 

Here's the colab notebook for reference:
 https://colab.research.google.com/drive/1xma-cG7hcU3vgFjHkZGZK8E2hmGEdn21#scrollTo=hL11Su8HQm9G

pranshu-raj-211:
@iiakshat The issue is in the buttons for feedback
on_click is assigned the lambda function for the func call on_feedback(value), adding a variable made it work for me.

Use this in on_click -  on_click=lambda _: on_feedback(True)

richard-to:
Looks like you resolved your error. When I ran the colab, it seemed ran for me.

-------


One thing to note is that Mesop event handler functions have a non-intuitive behavior if you try to reuse the same function. You should read: https://google.github.io/mesop/guides/interactivity/#avoid-using-closure-variables-in-event-handler

srivats22:
Can anyone contribute to this and add those components??

If so I would also like to have a go at it...

richard-to:
@srivats22, yes, you're right free to contribute any patterns you'd feel would be helpful.

srivats22:
Nice, let me clone the repo and see what I can work on... and will try and get it done

Issue #320, opened 2024-05-28
Be able to show snackbar


wwwillchen:

wwwillchen:
Closed since it's done in #575

Issue #278, opened 2024-05-17
Add dialog component


wwwillchen:
https://material.angular.io/components/dialog/overview

wwwillchen:
Closed - example in #575

PR #575, opened 2024-07-04
Add example patterns for modal and toast

richard-to:
This change includes two example components for modals and toasts. These are two common UI widgets that Mesop does not currently have. 

Unfortunately the Angular Components for Dialog and Snackbar is not easy to add to Mesop. This is because these two components takes in a Component or Template. So it does not seem possible to use projection into ng-content (like we do with badge or button).

The good news is that those two components aren't that complicated and it is relatively easy to create similar implementations using pure Mesop.

Ref: https://github.com/google/mesop/issues/320
Ref: https://github.com/google/mesop/issues/278

# Example modal

Screenshot 2024-07-03 at 6 37 57 PM

# Example toast

Screenshot 2024-07-03 at 6 37 50 PM


demo/modal.py
```diff
@@ -0,0 +1,113 @@
...
+      position="fixed",
+      width="100%",
+      z_index=1000,
+    )
+  ):
+    with me.box(
+      style=me.Style(
+        align_items="center",
+        display="grid",
+        height="100vh",
```
wwwillchen, line 72: I'm a little surprised this is 100vh and the parent box is 100%, I would have expected the opposite

demo/toast.py
```diff
@@ -0,0 +1,182 @@
...
+from typing import Callable, Literal
+
+import mesop as me
+
+
+@me.stateclass
+class State:
+  is_visible: bool = False
+  duration: int = 0
+  horizontal_position: str = "start"
```
wwwillchen, line 13: another nitpicky comment: I'd pick as default

horizontal = "center"
vertical = "end"

Although the user can select it from the dropdown, I think it's nice to use these values as the default (and you already use it as the default in the function parameters)

demo/toast.py
```diff
@@ -0,0 +1,182 @@
+"""Simple toast component that is similar to Angular Component Snackbar."""
```
wwwillchen, line 1: I'd probably just call this snackbar to be consistent with what material calls it (i.e. https://m3.material.io/components/snackbar/overview), I think toast is an old android UI/API which looks a little different (there's no action/button) than snackbar, which is more modern - https://www.geeksforgeeks.org/differnce-between-android-toast-and-snackbar/

wwwillchen (approved):
This is great! Sorry, for the nitpicky comments :)

I think these kinds of patterns are very helpful

demo/modal.py
```diff
@@ -0,0 +1,113 @@
+"""Simple modal that looks similar to Angular Component Dialog."""
```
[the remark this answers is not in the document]
richard-to: Sure. Updated to dialog.

demo/modal.py
```diff
@@ -0,0 +1,113 @@
...
+  security_policy=me.SecurityPolicy(
+    allowed_iframe_parents=["https://google.github.io"]
+  ),
+  path="/modal",
+)
+def app():
+  state = me.state(State)
+
+  with modal(state.is_open):
+    with modal_container():
```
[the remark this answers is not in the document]
richard-to: That's fair. Since this is a copyable example I think your suggestion makes sense. I had wanted to keep it separate in case someone wanted a different design for the modal. So if they didn't want the rounded corners. But they can just edit the code in this case.

demo/toast.py
richard-to: Sure. No problem. Updated.

demo/toast.py
richard-to: Lol, thanks. I didn't realize they were different UI things. Updated.

Issue #160, opened 2024-04-26
Allow styling of table component rows/columns


richard-to:
Need to determine a way to enable styling of table rows/columns.

Some use cases from user feedback:

1. Highlight selected row / cell
2. Alternate coloring of rows

There are are also other use cases for being able to control color, such as if a cell coloring based on the value.

zacharias1219:
Can we use state-driven styling for this??

aperepel:
Additional use cases:
+ Highlight a row on hover
+ Tooltip (i.e. abbreviated text shown in full)

Most interested in eventually having the 'expandable row' feature from https://material.angular.io/components/table/examples

This should help tremendously with the master/detail workflows.

ghchinoy:
Table rows are currently centered horizontally, it would be good to be able to change the alignment, too, to top or bottom

PR #577, opened 2024-07-04
Add grid table demo

richard-to:
One thing I'm beginning to notice is that some of the Angular Components aren't really that special. And that we can probably write pure Mesop versions that would be more amenable to the Mesop event flow.

So one thing that occurred to me is that we don't need to render tables using table mark up. We can just use flex box or grid. In this case I used grid.

By doing this, it allows us to implement more advanced functionality quicker than if we to work with the Angular Table component. And I know some people have been asking for more table functionality.

So as an example, I tried to build a a moderately featured table component. One major thing that is missing is pagination, but I ran out of gas. But otherwise we have custom styling, flexible custom cell formatting, sorting, filtering.

Another area to explore I want to explore is a custom web component that integrates some other grid library, such as AG Grid.

Ref: https://github.com/google/mesop/issues/160

# Grid Table With Sort

Screenshot 2024-07-04 at 2 58 56 PM

# Grid Table Expanded

Screenshot 2024-07-04 at 2 59 06 PM

# Grid Table With Filter

Screenshot 2024-07-04 at 2 59 19 PM

### DIFF
--- a/demo/grid_table.py
+++ b/demo/grid_table.py
@@ -210,16 +210,16 @@ class GridTableTheme(Protocol):
   """Interface for theming the grid table"""
 
   def header(self, sortable: bool = False) -> me.Style:
-    pass
+    return me.Style()
 
   def sort_icon(self, current_column: str, sort_column: str) -> me.Style:
-    pass
+    return me.Style()
 
-  def row(self, cell_meta: GridTableCellMeta) -> me.Style:
-    pass
+  def cell(self, cell_meta: GridTableCellMeta) -> me.Style:
+    return me.Style()
 
-  def expander(self, cell_meta: GridTableCellMeta) -> me.Style:
-    pass
+  def expander(self, df_row_index: int) -> me.Style:
+    return me.Style()
 
 
 class GridTableThemeDark(GridTableTheme):
@@ -264,7 +264,7 @@ class GridTableThemeDark(GridTableTheme):
       border=self._BORDER,
     )
 
-  def expander(self, cell_meta: GridTableCellMeta) -> me.Style:
+  def expander(self, df_row_index: int) -> me.Style:
     return me.Style(
       background=self._CELL_BG,
       color=self._COLOR,
@@ -319,7 +319,7 @@ class GridTableThemeLight(GridTableTheme):
       border=self._CELL_BORDER,
     )
 
-  def expander(self, cell_meta: GridTableCellMeta) -> me.Style:
+  def expander(self, df_row_index: int) -> me.Style:
     return me.Style(
       background=self._CELL_BG,
       color=self._COLOR,
@@ -386,7 +386,7 @@ def on_table_sort(e: me.ClickEvent):
   state.sort_column = column
 
 
-def expander(cell_meta: GridTableCellMeta):
+def expander(df_row_index: int):
   """Rendering logic for expanded row.
 
   Here we just display the row data in two columns as text inputs.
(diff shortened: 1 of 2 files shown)


demo/grid_table.py
```diff
@@ -0,0 +1,558 @@
...
+          col_index_name_map[col_index], GridTableColumn()
+        )
+        cell_meta = GridTableCellMeta(
+          row_index=row[0],
+          col_index=col_index,
+          name=col_index_name_map[col_index],
+          value=col,
+        )
+        with me.box(
+          # Story the row index and col index for the cell click event so we know which
```
wwwillchen, line 436: typo: story -> store

demo/grid_table.py
```diff
@@ -0,0 +1,558 @@
...
+      ),
+      sort_column=state.sort_column,
+      sort_direction=state.sort_direction,
+    )
+
+    # Used for demonstrating "table button" click example.
+    if state.string_output:
+      with me.box(
+        style=me.Style(
+          background="#141d2c",
```
wwwillchen, line 108: not the most important thing, but I think it would be nice to use "light theme" colors or have a toggle-able option between light and dark theme. (most people are going to copy this directly, and it's going to be kinda hard for them to adjust all the colors in a reasonably visually appealling way).

wwwillchen (approved):
Overall, I like this direction! I had a similar thought as you that people should just use CSS grid directly for these advanced table use cases :)

---

Not an actionable comment, but I do wonder if we should eventually graduate grid_table into mesop.labs. For example, with your other PR for dialog + snackbar, I was kind of wondering the same thing. One potential inspiration is shadcn which is a "copy-and-pastable" UI library - they do a really nice job of having demo code + impl code separated out, and then make it easy to add/copy the impl code into your project, e.g. https://ui.shadcn.com/docs/components/table

I filed an issue here: https://github.com/google/mesop/issues/579 to track this. I guess in the same way of how I was thinking about `pip install git+...`, maybe what I want is something more like `mesop add github.com/google/mesop/tree/main/demo/grid_table.py` which copies the file(s) over
into my project.

demo/grid_table.py
richard-to: That's a good point. I added a theme class for controlling the main theme. And added examples for light and dark.

richard-to:
> Not an actionable comment, but I do wonder if we should eventually graduate grid_table into mesop.labs. For example, with your other PR for dialog + snackbar, I was kind of wondering the same thing. One potential inspiration is shadcn which is a "copy-and-pastable" UI library - they do a really nice job of having demo code + impl code separated out, and then make it easy to add/copy the impl code into your project, e.g. https://ui.shadcn.com/docs/components/table

I filed an issue here: https://github.com/google/mesop/issues/579 to track this. I guess in the same way of how I was thinking about pip install git+..., maybe what I want is something more like mesop add github.com/google/mesop/tree/main/demo/grid_table.py which copies the file(s) over
into my project.

Yeah, that's a great question. I was also wondering how we could make it easier for people to use these user definied components.

Yeah I think labs is definitely a possibility especially if we don't feel we can integrate Dialog / Snackbar in the near future. But yeah I think for now, the implementations still need polishing and improvements for them to make it to labs.

I also considered adding them to a separate repository, but I think labs would make more sense. Plus probably easier for downstream.

I like the idea of making it easier to copy and paste the component part of the example.

Yeah, I think for copy/paste, then I do like the `mesop add github.com/google/mesop/tree/main/demo/grid_table.py` I think maybe want to make it easier for people to find the path to the file somehow, but initially I think that would be a good start. Some caveats being how do we prevent people from just doing add for random python files.

PR #589, opened 2024-07-09
Add Firestore state session backend

richard-to:
- Adds `firestore` option to `MESOP_STATE_SESSION_BACKEND`
- New config: `MESOP_STATE_SESSION_BACKEND_FIRESTORE_COLLECTION`
- Updates config.md
- Notes on implementation
  - User needs to manually set the TTL policy - Though the policy seems a bit slow to delete stuff
  - Uses the `(default)` database only
  - Uses application default credentials to pick project

Ref #556

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -12,7 +12,7 @@ Sets the backend to use for caching state data server-side. This makes it so sta
 not have to be sent to the server on every request, reducing bandwidth, especially if
 you have large state objects.
 
-The backend options available at the moment are `memory` and `file`.
+The backend options available at the moment are `memory`, `file`, and `firestore`.
 
 #### memory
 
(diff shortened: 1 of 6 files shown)


docs/api/config.md
```diff
@@ -49,6 +49,51 @@ traffic and state size.
...
+GCP_PROJECT=gcp-project
+gcloud config set project $GCP_PROJECT
+gcloud auth application-default set-quota-project $GCP_PROJECT
+```
+
+Mesop leverages Firestore's [TTL policies](https://firebase.google.com/docs/firestore/ttl)
+to delete stale state sessions. This needs to be set up using the following command,
+otherwise old data will accumulate unnecessarily.
+
+```sh
```
wwwillchen, line 88: I'm a little concerned people aren't going to read this far and forget to set this, and then rack up a hefty bill with firestore. In the past when I've used firestore, I never used TTL and it's generally not too bad since you're keying things off a stable id like a user id.

I think it would be good to either: a) print some kind of warning when server starts up and firestore is configured as the state session backend to remind the developer to set TTL or b) detect that TTL policy has _not_ been set and remind the user. I think b) is more work and not sure if it's doable.

richard-to: Yes, that's a great point. I'll add a warning.

The other thing I noticed just now is that using the TTL requires that billing is enabled. The deletion is $0.01 per 100,000 documents.

The other thing is that the TTL deletion takes a while. I have some example records that still haven't been deleted. In the cases where you don't have really large state objects. That's probably fine. Most likely they'd stay under the 1GB daily storage free allowance. But if some's storing a bunch of even 20MB, that could potentially fill up fast.

So one option I think is to clear stale sessions by tracking them per session. So if a process saves a session, it would also save the document ID and timestamp. Then it could try to delete any sessions it has created.

Since the session may be deleted from a different process, we'd have to live with trying to delete sessions that were already deleted. This would lead to extra reads, but should be OK for reasonable demo traffic.

Unfortunately this solution wouldn't be able to delete all stale sessions since the Cloud Run process shutdown with stale sessions still around. So the TTL policy would still be need.

----------

If we want a solution that does not depend on the TTL policy, it can definitely be done. I think the main consideration here is how many extra reads / writes / deletes do we want to incur.

At the moment, I'd vote for enabling requiring the TTL policy.

----------

In terms of billing getting out of control, I think in Firestore, there's a way to set a billing cap. So someone could set it to $5 dollars. But I guess they'd have to be aware of that too.

wwwillchen (approved):

Issue #588, opened 2024-07-08
Unintentional on_input_enter trigger when confirming IME composition in chat component


kyohei3:
**Describe the bug**

This is an issue in the chat component.

When using an IME that requires composition, such as Japanese or Chinese, the on_input_enter event is unintentionally triggered when pressing Enter to confirm the composition.

**To Reproduce**

1. Go to the [demo gallery](https://google.github.io/mesop/demo/) and open Chat demo.
2. Focus on the input element below and type something using a Japanese/Chinese IME (or any other IME that requires composition).
3. Press "Enter" to confirm the IME conversion.

**Expected behavior**

The composition should be confirmed, but the text should not be passed to the `on_input_enter` event.

**Screenshots**

**Desktop System Info**

- OS: macOS Sonoma 14.5
- Browser: Google Chrome
- Version:  126.0.6478.127（Official Build） （arm64）

**Additional context**

richard-to:
Thanks for the catch. That's a tricky one.

As a temporary fix, you can make a copy of https://github.com/google/mesop/blob/main/mesop/labs/chat.py and remove the on enter event on line 13. This means you will only be able to send the chat input by pressing the button. But probably better than the current behavior.

Basically, copy the chat.py file to your own chat.py file. Then import your custom chat.py into your main mesop application.

kyohei3:
@richard-to Thank you for the advice. 

I tried to change code like this, but it does not work with my environment (Mac and Chrome).

```typescript
const keyboardEvent = event as KeyboardEvent;
if (keyboardEvent.key === 'Enter' && !keyboardEvent.isComposing) {
  // ...
}
```

As far as I observed key events on my environment and read some Japanese articles, `isComposing` is set to `true` on keydown event but set to `false` on keyup event. I locally changed here to `(keydown)=...` and found it worked as expected.

https://github.com/google/mesop/blob/273c65d26f710f19504525b7295e101430f284d5/mesop/components/input/input.ng.html#L54

Do you think it makes scene to replace keyup event with keydown?

richard-to:
Thanks for looking into this. That's very interesting to know.

Key up would be preferable since it won't fire that many events to Mesop server. If necessary, it's possible we can debounce the keydown if someone decides to hold down the enter key for whatever reason.

---------

I wonder if we can have both a key up and key down event. So the key down event would check if isComposing is True (this would be stored as a variable on the component. Then on key up we could check the stored variable.

[...]

PR #593, opened 2024-07-09
Fix false on enter trigger when using IME.

richard-to:
The isComposing field tells us if we are using an input method editor (IME) which
will display a menu of characters that can't be represented on a standard QWERTY
keyboard. The user can select from this menu by pressing "enter" or clicking the
menu item. The user will then need to press "enter" to confirm the selection (at
least on MacOS).

If we do not check isComposing on key up, this will trigger the Mesop on enter
event.

If we check isComposing on key up, the enter confirmation will trigger the Mesop
on enter event since isComposing becomes false on key up.

We need to track isComposing with the key down event since this will trigger the
on enter event when "enter" is pressed again (when the IME is inactive and
a selection has been made and applied).

Safari has a bug where the order of operations for isComposing is in a different
order. See https://bugs.webkit.org/show_bug.cgi?id=165004.

In order to work around this issue, we need to count expectedIsComposingCount
equals false twice on Safari rather than just once.

Ref: https://github.com/google/mesop/issues/588

### DIFF
--- a/mesop/components/input/input.ts
+++ b/mesop/components/input/input.ts
@@ -25,7 +25,8 @@ export class InputComponent {
   @Input() style!: Style;
   private _config!: InputType;
   private inputSubject = new Subject<Event>();
-
+  private isComposingCount: number = 1;
+  private expectedIsComposingCount: number = 1;
   constructor(private readonly channel: Channel) {
     this.inputSubject
       .pipe(
@@ -35,6 +36,14 @@ export class InputComponent {
         debounceTime(150),
       )
       .subscribe((event) => this.onInputDebounced(event));
+    // See keyDown event for explanation for why we have a special case for Safari.
+    if (
+      navigator.userAgent.indexOf('Safari') > -1 &&
+      navigator.userAgent.indexOf('Chrome') === -1
+    ) {
+      this.expectedIsComposingCount = 2;
+      this.isComposingCount = 2;
+    }
   }
 
   ngOnDestroy(): void {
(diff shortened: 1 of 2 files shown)


richard-to:
Hmm, I tested on Chrome and Firefox and the behavior was the same. However, Safari didn't work.

mesop/components/input/input.ts
```diff
@@ -35,6 +36,14 @@ export class InputComponent {
         debounceTime(150),
       )
       .subscribe((event) => this.onInputDebounced(event));
+    // See keyDown event for explanation for why we have a special case for Safari.
+    if (
+      navigator.userAgent.indexOf('Safari') > -1 &&
```
wwwillchen, line 41: minor: you can use String.prototype.includes for a bit more straightforward check: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

mesop/components/input/input.ts
```diff
@@ -35,6 +36,14 @@ export class InputComponent {
...
+      navigator.userAgent.indexOf('Chrome') === -1
```
wwwillchen, line 42: It's probably worth an inline comment to explain why you're checking for the absence of chrome (i.e. Chrome's user-agent basically spoofs other browsers and includes Safari, so we need to explicitly look for the absence of it).

wwwillchen (approved):
Could we add a test for this? at least for the chrome case.

richard-to:
> Could we add a test for this? at least for the chrome case.

I think that's a bit tricky to do since the IME stuff is OS specific. The user also has to enable a language that would trigger the IME.

So I think it could be possible to do if the VM we're using in playwright for CI had Japanese enabled. But I think in that scenario, we'd skip it for local tests since that means the user needs to enable Japanese on their OS. And the key combination for switching languages may be different in each OS.

So I think for now, I'll skip the Playwright test. I think it may be possible to test some of the behavior in a unit test. But I don't think we currently have examples of testing angular components via unit test. So will skip for now. Maybe can come back to it later.

PR #597, opened 2024-07-11
Add SQL state session backend.

richard-to:
This state session backend uses SQL Alchemy to support different SQL databases.

I tested using SQlite3 and PostgreSQL. Both seemed to work fine.

One slight workflow issue is that for non-SQLite3 databases, people will need to manually run a script to create the database table used for storing state sessions. The commands are mentioned in the config.md file. That's not the most user friendly approach. So probably need a better way to approach it. But I think for the initial version of this backend, it should be OK, especially considering we don't know how many people will actually use it.

Ref: https://github.com/google/mesop/issues/556


mesop/server/state_session.py
```diff
@@ -269,6 +269,98 @@ def clear_stale_sessions(self):
...
+
+  def clear_stale_sessions(self):
+    """Clears stale sessions from database.
+
+    Only query the database after every X requests to avoid fruitlessly querying the
+    database unnecessarily.
+    """
+    from sqlalchemy import delete
+
+    if self.request_count < self._SESSION_CLEAR_N_REQUESTS:
```
wwwillchen, line 349: This doesn't seem right to me. For services like Cloud Run, it's very possible that the process only serves one request (or a few), before it's disposed.

I think we can keep this simple for now and just clear stale sessions every time. It's possible for some SQL databases, they can have a TTL policy, but I think we can wait before optimizing this.

build_defs/requirements.txt
```diff
@@ -20,3 +20,10 @@ mkdocstrings[python]
 firebase-admin
 matplotlib
 pandas
+sqlalchemy
```
wwwillchen, line 23: I don't think we should mark these as "Deps for examples" because you are actually loading them in Mesop code, but you're lazy loading them. I'd add a section above L11 like "Optional (lazily-loaded) deps:"

docs/api/config.md
```diff
@@ -49,6 +49,62 @@ traffic and state size.
...
+  MetaData,
+  String,
+  Table,
+  create_engine,
+)
+
+CONNECTION_URI = "your-database-connection-uri"
+TABLE_NAME = "your-table-name" # or the default "mesop_state_session"
+
+self.db = create_engine(CONNECTION_URI)
```
wwwillchen, line 89: a little confusing why there's self even though this isn't in a class __init__, I would just remove all the self.

docs/api/config.md
```diff
@@ -49,6 +49,62 @@ traffic and state size.
...
+  DateTime,
+  LargeBinary,
+  MetaData,
+  String,
+  Table,
+  create_engine,
+)
+
+CONNECTION_URI = "your-database-connection-uri"
+TABLE_NAME = "your-table-name" # or the default "mesop_state_session"
```
wwwillchen, line 87: I would reverse this; set the value to "mesop_state_session" and have the comment be, you can configure to "your-table-name", but then you need to override MESOP_STATE_SESSION_BACKEND_SQL_TABLE env var

mesop/server/config.py
```diff
@@ -11,9 +11,13 @@
...
   """Mesop app configuration."""
 
-  state_session_backend: Literal["memory", "file", "firestore", "none"] = "none"
+  state_session_backend: Literal[
+    "memory", "file", "firestore", "sql", "none"
+  ] = "none"
   state_session_backend_file_base_dir: Path = Path("/tmp")
   state_session_backend_firestore_collection: str = "mesop_state_sessions"
+  state_session_backend_sql_connection_uri: str = ""
+  state_session_backend_sql_table: str = "mesop_state_session"
```
wwwillchen, line 20: I would use the same str as you do for firestore "mesop_state_sessions" for consistency (I think it's fine to pluralize SQL table names b/c I think conventions vary widely for table naming)

mesop/server/config.py
```diff
@@ -11,9 +11,13 @@
 class Config(BaseModel):
   """Mesop app configuration."""
 
-  state_session_backend: Literal["memory", "file", "firestore", "none"] = "none"
+  state_session_backend: Literal[
+    "memory", "file", "firestore", "sql", "none"
+  ] = "none"
```
wwwillchen, line 16: I meant to ask this earlier; it's not clear why you need to set default values for these attributes since it looks like you're always setting a value for each attribute in CreateConfigFromEnv

mesop/server/state_session.py
```diff
@@ -269,6 +269,98 @@ def clear_stale_sessions(self):
...
+
+    with self.db.connect() as conn:
+      result = conn.execute(
+        select(self.table).where(self.table.c.token == token)
+      ).one_or_none()
+      if result:
+        states_data = msgpack.unpackb(result.states)  # type: ignore
+        _deserialize_state(states, states_data)
+      else:
+        raise Exception("Token not found in state session backend.")
```
wwwillchen, line 324: raise MesopException

mesop/server/state_session_test.py
```diff
@@ -251,6 +261,78 @@ def test_file_backend_clear_stale_sessions_skipped(tmp_path):
     assert empty_states == {type(StateA): StateA()}
 
 
+def test_sql_backend_clear_stale_sessions_not_stale(sqlite_backend):
+  # GIVEN
+  empty_states: States = {
```
wwwillchen, line 266: not very clear why empty_states is initialized to a non-empty value

wwwillchen (changes requested):

build_defs/requirements.txt
richard-to: Sure sounds good.

docs/api/config.md
richard-to: Good idea. That makes more sense.

mesop/server/config.py
richard-to: Yeah, I think initially I was considering the case where someone didn't use the factory function, but since internally the only usage is from the Factory, I've updated to set the defaults from factory function under the assumption that the config class generally wouldn't be instantiated without a factory function.

The other consideration was updating the factory function to only pass in the env vars that were set, thus allowing the defaults to be set by the config class.

mesop/server/state_session.py
richard-to: That's a good point. I think it would still be good to run every X requests, but just at a lower number. I've changed that to 10 for now.

I think Cloud Run instances get disposed if they are inactive for 15 minutes.

It's definitely possible that the server never hits 10 requests, but in that case, there's probably very little data there. It definitely could slowly accumulate more and more stale sessions but it would take a while to do so while never hitting 10 requests.

In future, this possible could be made configurable so the user can control based on their traffic and server.

mesop/server/config.py
richard-to: Heh, that's true. I was on the fence about this and ended up going for my preference non-plural table names. However, I agree that it would be better to be consistent with how we name the firestore collection.

mesop/server/config.py
richard-to: Actually, changed my mind. Switched to setting defaults in the config class. Then specifying parameter in the factory function only if the env var was set. Forgot I had some tests state session that were depending on this behavior.

mesop/server/state_session.py
wwwillchen: Good point about the 15 minute waiting period - I didn't realize it was that long. I think this is good enough for now.

I think longer term, something like [Cloud Run scheduled jobs](https://cloud.google.com/run/docs/execute/jobs-on-schedule) could be a more robust solution as it would enable cleanup on a regular cadence without contention, but it's a more complex deployment.

but I think, overall, we can wait for more feedback on our state session backends, before making more improvements.

wwwillchen (approved):

mesop/server/state_session.py
richard-to: Nice point about Cloud Run Scheduled Jobs. I totally forgot about that feature. Another option is just using plain Cloud Scheduler (or Cron Job) and hitting an endpoint exposed by the Mesop server.

I'd agree that if I were developing this for myself, I'd set up some separate task to do the clean up rather than doing after the server request. But I  also agree about potentially more complex development. Cloud Scheduler / Cron Jobs I think are relatively easy to use so long as we could safely expose an endpoint that cleaned up the sessions (though the issue with that is that it would only work for certain types of backends (i.e. probably wouldn't work for memory and with file, we'd need to hit all the servers).

So agree about waiting for more feedback.

Issue #580, opened 2024-07-05
Support multi-file demos in demo gallery


wwwillchen:
Needed for #535 and if we do this, it would be good to have #579  solved, which will make it easy for people to copy these multi-file demos.

tpcgold:
This seems to be the easiest way to do it (disclaimer: we currently do so as there doesn't seem another option):

```python
# chat_multimodal.py
import mesop as me

from .base_chat_app import BaseChatApp
from .config import _BOT_USER_DEFAULT


class ChatMultimodal(BaseChatApp):
    def render_input_box(self):
        super().render_input_box()
        with me.box(
            style=me.Style(flex_grow=1, display="flex", flex_direction="column")
        ):
            me.uploader(
                label="Upload File",
                accepted_file_types=["image/jpeg", "image/png", "application/pdf"],
                on_upload=lambda e: self.on_click_upload(e, file=1),
                type="flat",
                color="primary",
                style=me.Style(font_weight="bold", margin=me.Margin(bottom=1)),
            )
            if self.state.file1:
                me.uploader(
                    label="Upload File",
                    accepted_file_types=["image/jpeg", "image/png", "application/pdf"],
                    on_upload=lambda e: self.on_click_upload(e, file=2),
                    type="flat",
                    color="primary",
                    style=me.Style(font_weight="bold", margin=me.Margin(bottom=1)),
                )
            if self.state.file2:
                me.uploader(
                    label="Upload File",
                    accepted_file_types=["image/jpeg", "image/png", "application/pdf"],
                    on_upload=lambda e: self.on_click_upload(e, file=3),
                    type="flat",
                    color="primary",
                    style=me.Style(font_weight="bold", margin=me.Margin(bottom=1)),
                )


def chat_multimodal(transform, *, state, title=None, bot_user=_BOT_USER_DEFAULT):
    app = ChatMultimodal(transform, state, title, bot_user)
    app.render()
```

Issue #579, opened 2024-07-05
Create a way to easily install patterns / components


wwwillchen:
**Motivation:** Make it easier for people to use existing components that people (both Mesop core team & broader community) create, which may be pure Python components or web components.

See: https://github.com/google/mesop/pull/577

Prior art: https://ui.shadcn.com/docs/components/table

PR #600, opened 2024-07-11
Update demo gallery: chat_inputs & visual polish

wwwillchen:


richard-to (approved):
Nice. These look pretty good. Much better than the default chat input. Tempted to use one of these in chat.py. But probably better to keep it as the default for simplicity.

Maybe one suggestion I have it to wrap the examples in a `@me.component` to show how these could be reusable. But I think that will be more useful once you add the copy code stuff in #579 and #580, so I think fine for now.

wwwillchen:
> Nice. These look pretty good. Much better than the default chat input. Tempted to use one of these in chat.py. But probably better to keep it as the default for simplicity.
> 
> Maybe one suggestion I have it to wrap the examples in a `@me.component` to show how these could be reusable. But I think that will be more useful once you add the copy code stuff in #579 and #580, so I think fine for now.

good idea - done.

PR #614, opened 2024-07-12
Set LRU cache for compute_fn_id

### LINKED ISSUE #410: Mesop can be very slow if there are a lot of child components being rendered with click event handlers
I had a use case a few weeks ago where I was splitting up about 200 lines in box/text components and making them clickable. I noticed Mesop got very sluggish in this scenario so I removed it. 

Didn't look into the root cause for this yet. My hunch is tree diffing with a lot of child components. But not sure.

wwwillchen:
Analysis: https://colab.research.google.com/drive/1KcAG0iQJ-vor3KQw7A-CFKW85hPb1_Bg?usp=sharing
Fixes #410

### DIFF
--- a/mesop/component_helpers/helper.py
+++ b/mesop/component_helpers/helper.py
@@ -1,7 +1,7 @@
 import hashlib
 import inspect
 import json
-from functools import wraps
+from functools import lru_cache, wraps
 from typing import (
   Any,
   Callable,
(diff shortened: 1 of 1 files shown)


richard-to (approved):
Awesome. Great find. Good old lru_cache.

Issue #608, opened 2024-07-12
Adding a Authenticator demo in the demo use cases section 


itsBaivab:
**Is your feature request related to a problem? Please describe.**
A saw that there is no demo related to how authentication works in mesop. 

**Describe the solution you'd like**
It is better to have a username, password filed and a login button that will welcome a user after logging. If the password and username fields are empty then it will through an error to the user. 

**Additional context**

this is how it should look like 
![image](https://github.com/user-attachments/assets/a6ac86ec-802d-4675-a820-2ac30799b9b2)

can i send a pr?

wwwillchen:
Feel free to send a PR - here's our contributing guide:  https://google.github.io/mesop/internal/contributing/

itsBaivab:
Submitted a pr let me know if I need to perform any changes.

PR #625, opened 2024-07-16
ADD demo/authentication

itsBaivab:
FIX : CI test fails


itsBaivab:
PR for #608

demo/authenticator.py
```diff
@@ -0,0 +1,105 @@
...
+    if state.username and state.password:
+        state.welcome_message = f"Welcome, {state.username}!"
+        state.error_message = ""  # Clear any previous error message
+        # Navigate to the welcome page after successful login
+        me.navigate("/welcome")
+    else:
+        state.error_message = "Incorrect username or password."
+    
+    # Clear the username and password fields
+    state.username = ""
```
richard-to, line 36: Good idea to clear these fields, especially password. 

Sorry for the delay on this review. I've just been thinking about if it's safe in Mesop to store the password in state. Ideally, we'd only send the password for the log in check (but never send it back). But I don't think that's currently possible in Mesop. The blur event helps in that at least we're not repeatedly sending parts of the password back and forth via the on input event.

I'm still not sure at the moment.

demo/authenticator.py
```diff
@@ -0,0 +1,105 @@
...
+                me.button(label="Login", on_click=on_login)
+
+@me.page(
+    security_policy=me.SecurityPolicy(
+    allowed_iframe_parents=["https://google.github.io"]
+  ),
+    path="/welcome"
+)
+def welcome_page():
+    s = me.state(State)
```
richard-to, line 102: [suggestion] Maybe add an example of redirecting the user to the log in page if not logged in. We may also want to simulate a log out flow as well.

I think this is another area where Mesop may be lacking in terms of the API. For example, if a user is not allowed to view the page, we may want to return a 401 http code and error page. But again, not something I think Mesop is able to do at the moment.

tools/defaults.bzl
```diff
@@ -364,17 +364,17 @@ def ng_web_test_suite(browsers = [], deps = [], static_css = [], exclude_init_sc
...
-        # Escape all double-quotes so that the content can be safely inlined into the
-        # JS template. Note that it needs to be escaped a second time because the string
-        # will be evaluated first in Bash and will then be stored in the JS output.
-        css_content=$$(cat $${files[0]} | sed 's/"/\\\\"/g')
-        js_template='var cssElement = document.createElement("style"); \
-                    cssElement.type = "text/css"; \
-                    cssElement.innerHTML = "'"$$css_content"'"; \
-                    document.head.appendChild(cssElement);'
-         echo "$$js_template" > $@
+            cmd = """\r
```
richard-to, line 367: [nit] can you revert this?

richard-to (changes requested):
Overall, it looks fine. Thanks for the contribution. I think it's illustrated some missing functionality in Mesop as a framework around how it handles sensitive information like passwords and also being able to control access to pages. So these may be some things we need to figure for Mesop, either patterns or framework improvements, before we can merge this example in.

richard-to:
Going to close this pull request since I don't think we have a clear way forward for authentication interfaces yet.

PR #630, opened 2024-07-18
Allow iframing any other sites

### LINKED ISSUE #628: Not able to iframe a localhost site from a localhost site (or http site)
I believe this requires frame-src to be updated to include `localhost:*` or `http:`

Currently it looks like:

```
"frame-src": "'self' https:"
```

I think this is a good default, but would be nice to have the option to allow localhost iframe embeds if the current host is localhost. Or allow the user to add additional options to frame-src.

wwwillchen:
Closes #628.

`frame-src` (i.e. iframing other sites) is fairly low-risk and we may as well allow iframing any site. This is different than `frame-ancestors` which is much more security sensitive and has a strict allowlist.

Background: https://cheatsheetseries.owasp.org/cheatsheets/Content_Security_Policy_Cheat_Sheet.html#preventing-framing-attacks-clickjacking-cross-site-leaks

### DIFF
--- a/mesop/server/static_file_serving.py
+++ b/mesop/server/static_file_serving.py
@@ -148,7 +148,7 @@ def configure_static_file_serving(
         "default-src": "'self'",
         "font-src": "fonts.gstatic.com",
         # Mesop app developers should be able to iframe other sites.
-        "frame-src": "'self' https:",
+        "frame-src": "*",
         # Mesop app developers should be able to load images and media from various origins.
         "img-src": "'self' data: https: http:",
         "media-src": "'self' data: https:",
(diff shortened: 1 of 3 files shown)


richard-to (approved):

richard-to:
Does iframing an http site from say an https pose an concerns? In this case, it's definitely the developer's choice. But I did kind of like the current setting that prevented the https -> http iframing case as a kind of safe guard for people who were not thinking about that.

wwwillchen:
> Does iframing an http site from say an https pose an concerns? In this case, it's definitely the developer's choice. But I did kind of like the current setting that prevented the https -> http iframing case as a kind of safe guard for people who were not thinking about that.

I don't think so. Even if you had a weird situation where the parent frame was https://foo.com and it iframed http://foo.com (this would be pretty unusual because usually if a site supports https, then the http would redirect to https), there's not really an additional risk because they would be considered separation origins (which is basically subdomain + protocol).

PR #631, opened 2024-07-18
Create starter kit template

wwwillchen:
- Creates a chat-like UI as a starter template which demonstrates core Mesop functionality , particularly with styling.
- Will be followed up by creating a `mesop init` CLI command.

image

### DIFF
--- a/mesop/examples/BUILD
+++ b/mesop/examples/BUILD
@@ -44,6 +44,7 @@ py_library(
         "//mesop/examples/docs",
         "//mesop/examples/integrations",
         "//mesop/examples/shared",
+        "//mesop/examples/starter_kit",
         "//mesop/examples/testing",
     ],
 )
--- a/mesop/examples/__init__.py
+++ b/mesop/examples/__init__.py
@@ -27,6 +27,7 @@ from mesop.examples import playground as playground
 from mesop.examples import playground_critic as playground_critic
 from mesop.examples import readme_app as readme_app
 from mesop.examples import scroll_into_view as scroll_into_view
+from mesop.examples import starter_kit as starter_kit
 from mesop.examples import sxs as sxs
 from mesop.examples import testing as testing
 from mesop.examples import viewport_size as viewport_size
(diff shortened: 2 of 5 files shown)


richard-to (approved):
I think this is a good start. But I think the starter example will need to be more robust/polished since it is going to be a big part of the initial Mesop experience. So providing something that looks really good visually will be important.

Just minor things mostly (not in any order):

1) Footer should hopefully be sticky at the bottom
2) We really need a real link component (since that prevents us from styling the link)
3) It's not a really a chat example, though one can definitely see how it could become one.
4) May help to make it clear how a user could add their ML functionality, similar to text_to_text / chat / text_to_image. In this case either have a stub function or a comment delineating where to put their logic
5) The output isn't really styled so it looks strange

----------

But overall, I like the idea. Will be good to have other starter kit options. Even something just adds the boilerplate for text_to_text / chat / text_to_image.

wwwillchen:
> I think this is a good start. But I think the starter example will need to be more robust/polished since it is going to be a big part of the initial Mesop experience. So providing something that looks really good visually will be important.
> 
> Just minor things mostly (not in any order):
> 
> 1. Footer should hopefully be sticky at the bottom

fixed it now that there's min-height :)

> 2. We really need a real link component (since that prevents us from styling the link)
agreed - I'll try to add this later.

> 3. It's not a really a chat example, though one can definitely see how it could become one.
> 4. May help to make it clear how a user could add their ML functionality, similar to text_to_text / chat / text_to_image. In this case either have a stub function or a comment delineating where to put their logic

good point - done.

> 5. The output isn't really styled so it looks strange

yeah, i added some styling.

> 6. Also can send input with nothing entered.

fixed.

> But overall, I like the idea. Will be good to have other starter kit options. Even something just adds the boilerplate for text_to_text / chat / text_to_image.

Yeah, I think this could definitely use some more polish, but it's better than nothing :) 

Having more options will be good. I think in the future we could do something like `mesop add https://github.com/google/mesop/blob/main/demo/chat.py` and this would copy the file down into your local directory.

PR #632, opened 2024-07-18
Add {min, max}-{height, width} to Style

wwwillchen:


richard-to (approved):
Nice. These are good additions. Surprised we didn't have these already. Only one minor nit is to update the doc string

wwwillchen:
> Nice. These are good additions. Surprised we didn't have these already. Only one minor nit is to update the doc string

good catch. updated docstring.

Yeah, I'd been using stuff like `width: max(300px, 100%)`, but there's actually something where you need to use max-width or it doesn't work right

PR #633, opened 2024-07-18
Add `mesop init` CLI command

wwwillchen:
This depends on #631

### DIFF
--- a/mesop/bin/bin.py
+++ b/mesop/bin/bin.py
@@ -30,15 +30,15 @@ flags.DEFINE_bool(
 def main(argv: Sequence[str]):
   if len(argv) < 2:
     print(
-      """\u001b[31mERROR: missing command-line argument to Mesop application.\u001b[0m
+      """\u001b[31mERROR: missing command-line argument to Mesop.\u001b[0m
 
-Example command:
+Example run command:
 $\u001b[35m mesop file.py\u001b[0m"""
     )
     sys.exit(1)
 
   if argv[1] == "init":
-    if len(argv) > 2:
+    if len(argv) > 3:
       print(
         f"""\u001b[31mERROR: Too many command-line arguments for mesop init.\u001b[0m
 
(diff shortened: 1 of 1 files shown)


mesop/bin/bin.py
```diff
@@ -30,13 +30,45 @@
...
-      """\u001b[31mERROR: missing command-line argument to Mesop application.\u001b[0m
+      """\u001b[31mERROR: missing command-line argument to Mesop.\u001b[0m
 
-Example command:
+Example run command:
 $\u001b[35m mesop file.py\u001b[0m"""
     )
     sys.exit(1)
 
+  if argv[1] == "init":
```
richard-to, line 40: Question about argv and absl flag. Will absl flag automatically remove the known flags from the argv?

Also is there a way to use argparse with absl flag? Could make working with the command line slightly cleaner.

mesop/bin/bin.py
```diff
@@ -30,13 +30,45 @@
...
+
+    filename = argv[2] if len(argv) == 3 else "main.py"
+    absolute_path = make_path_absolute(filename)
+    if os.path.isfile(absolute_path):
+      print(f"\u001b[31mERROR: {filename} already exists.\u001b[0m")
+      sys.exit(1)
+    with open(
+      os.path.join(
+        os.path.dirname(os.path.abspath(__file__)),
+        "../examples/starter_kit/starter_kit.py",
```
richard-to, line 61: Ah interesting. Didn't know we included the examples in the Mesop PIP package.

mesop/bin/bin.py
```diff
@@ -30,13 +30,45 @@
...
+  if argv[1] == "init":
+    if len(argv) > 3:
+      print(
+        f"""\u001b[31mERROR: Too many command-line arguments for mesop init.\u001b[0m
+
+Actual:
+$\u001b[35m mesop {" ".join(argv[1:])}\u001b[0m
+
+Re-run with:
+$\u001b[35m mesop init file.py\u001b[0m"""
```
richard-to, line 49: [question] Do you think people would interpret file.py literally?

-----

I think this goes back to one of the Github issues about the CLI help not being to helpful

richard-to (approved):

mesop/bin/bin.py
wwwillchen: yes, argv is the non-flag arguments.

Turns out you can!
https://github.com/abseil/abseil-py/blob/main/absl/flags/argparse_flags.py

mesop/bin/bin.py
wwwillchen: yeah, i forgot we included it too. I think it's done here: https://github.com/google/mesop/blob/b47ea60130b9d082e088e276d317c08ddce08bfc/mesop/pip_package/build_pip_package.sh#L74

we could exclude the other examples, but it's probably not too important to trim our binary size (yet)

mesop/bin/bin.py
wwwillchen: I'm not sure. I thought about including the argument they provided, but it might not even be a python filename either. I think we can go with this for now and if people get confused we can update this.

Even if they literally do file.py, it's not that bad, we will tell them to run `mesop file.py` and then I think they'll get the idea.

PR #634, opened 2024-07-18
Update docs with starter kit

wwwillchen:
@richard-to do you think there should be a standalone docs page for the starter kit?

### DIFF
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -1,6 +1,6 @@
 # Components
 
-Please read [Quickstart](../getting_started/quickstart.md) before this as it explains the basics of components. This page provides an overview of the different types of components in Mesop.
+Please read [Core Concepts](../getting_started/core_concepts.md) before this as it explains the basics of components. This page provides an overview of the different types of components in Mesop.
 
 ## Types of components
 
--- a/docs/getting_started/core_concepts.md
+++ b/docs/getting_started/core_concepts.md
@@ -1,0 +1,80 @@
+This doc will explain the core concepts of building a Mesop app.
+
+## Hello World app
+
+Let's start by creating a simple Hello World app in Mesop:
+
+```python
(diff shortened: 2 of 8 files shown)


richard-to:
That's a good question. I'm not sure. I think for now starting with quickstart is fine. Can always adjust it later. But it does seem like the fastest way to get started, so I think it makes sense in quickstart.

But I guess maybe you're asking if the detailed explanation of the quick start should be in a separate document. In that case, I think yes. Because user does not necessarily need to know all the gritty details of how it works. The quick start should be just be boiled down to what you need to do get an app running. From there, the user could read more or they're experienced enough to get it by reading the code.

Personally, I'm a big fan of the Divio documentation system (https://docs.divio.com/documentation-system/) in terms of how they split up documentation. I see the quick start as a tutorial type documentation. Or maybe how-to. 

And I think the explanation of the code is more in line with "explanation" documentation in the divio system.

wwwillchen:
> That's a good question. I'm not sure. I think for now starting with quickstart is fine. Can always adjust it later. But it does seem like the fastest way to get started, so I think it makes sense in quickstart.
> 
> But I guess maybe you're asking if the detailed explanation of the quick start should be in a separate document. In that case, I think yes. Because user does not necessarily need to know all the gritty details of how it works. The quick start should be just be boiled down to what you need to do get an app running. From there, the user could read more or they're experienced enough to get it by reading the code.
> 
> Personally, I'm a big fan of the Divio documentation system (https://docs.divio.com/documentation-system/) in terms of how they split up documentation. I see the quick start as a tutorial type documentation. Or maybe how-to.
> 
> And I think the explanation of the code is more in line with "explanation" documentation in the divio system.

thanks for the feedback. i split out the detailed explanation from the quick start into a "core concepts" page, which feels a lot better. I've also linked the different getting started pages with a "next" link to make it a bit more coheisve.

PR #639, opened 2024-07-19
Add a framework comparison doc

### LINKED ISSUE #638: Create a comparison guide with Streamlit / Gradio / Reflex

wwwillchen:
Closes #638

### DIFF
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -42,17 +42,17 @@ Mesop, while well-suited for ML/AI use cases, is a more general-purpose framewor
 
 Gradio provides a set of pre-built components optimized for common ML inputs and outputs (e.g. image classification, text generation). This makes it fast to set up standard model interfaces. Gradio has a powerful [theming system](https://www.gradio.app/guides/theming-guide) that provides prebuilt theme options and many options for customizing the UI theme.
 
-Although Mesop offers a few AI-specific components like [chat](./components/chat.md), Mesop provides general-purpose UI components, which can be used for a variety of layout and UI designs. This makes it better suited for building custom interfaces, for example like the [demo gallery](https://google.github.io/mesop/demo/).
+Mesop provides general-purpose UI components, which can be used for a variety of layout and UI designs. Higher-level components like the [chat](./components/chat.md) component are [built](https://github.com/google/mesop/blob/main/mesop/labs/chat.py) on top of these low-level UI components. This makes it better suited for building custom interfaces, for example like the [demo gallery](https://google.github.io/mesop/demo/).
+
+### State management
+
+Gradio uses an imperative approach to [state management](https://www.gradio.app/guides/state-in-blocks), tightly coupling state with component updates. State is typically managed through function parameters and return values, which can be straightforward for simple interfaces but may become complex as applications grow.
+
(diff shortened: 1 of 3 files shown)


wwwillchen:
@richard-to finally ready for review :)

Once you've taken a look, I'll share this with some Googlers and also the framework authors before publishing.

richard-to (approved):
My perspective is from someone who hasn't really used Gradio or Streamlit.

In reading this at a high level, I wasn't quite able to tell what was a pro/con between Mesop in some cases. So I think it reads somewhat objectively.

I think what's a bit confusing for me is that the document compares different aspects of Streamlit versus Gradio. I don't think that's a problem necessarily, but it could feel like cherry picking certain things. I don't think that's what's happening here. I think the doc is just pointing out the key differentiating features.

One thing I'd like to hear more about is performance  / scalability of Streamlit / Gradio / Mesop. I think that's an important use case.

Also what types of use cases are Streamlit / Gradio / Mesop typically used for.

Another thing I'd be curious of is extensibility. So if the core stuff is not sufficient, how much work is it to find/get the functionality I want into my app.

wwwillchen:
Thanks for the review!

> My perspective is from someone who hasn't really used Gradio or Streamlit.
> 
> In reading this at a high level, I wasn't quite able to tell what was a pro/con between Mesop in some cases. So I think it reads somewhat objectively.
> 
> I think what's a bit confusing for me is that the document compares different aspects of Streamlit versus Gradio. I don't think that's a problem necessarily, but it could feel like cherry picking certain things. I don't think that's what's happening here. I think the doc is just pointing out the key differentiating features.

Yeah, I thought about this too and I think if we were to be comprehensive (and have the same criteria), it would be a much longer doc.

> 
> One thing I'd like to hear more about is performance / scalability of Streamlit / Gradio / Mesop. I think that's an important use case.

I think it's important but also requires benchmarking to be objective, speaking of which, we should do it ourselves :) Filed #706.

> 
> Also what types of use cases are Streamlit / Gradio / Mesop typically used for.

I mention this briefly - Streamlit historically has focused more on the data science use case, Gradio on ML use cases, but, tbh, everything seems to be converging as all three support LLM use cases.

> Another thing I'd be curious of is extensibility. So if the core stuff is not sufficient, how much work is it to find/get the functionality I want into my app.

Let me add this in - it's a big topic but I think the main differences are the UI frameworks that are well-supported.

docs/comparison.md
```diff
@@ -0,0 +1,73 @@
...
+
+Mesop enables creating custom [web components](https://google.github.io/mesop/web_components/) based on open web standards, facilitating interoperability with components from different frameworks like Lit. Mesop web components are rendered in the same frame as the rest of the Mesop app which provides more flexibility but less isolation.
+
+Streamlit has a more established ecosystem of community-developed components, while Mesop's community and component ecosystem are still developing.
+
+## Gradio
+
+Gradio and Mesop both enable rapid ML/AI app development but with different approaches.
+
+Gradio has a strong focus on creating demos and interfaces for machine learning models and makes it easy to build a UI for a model.
```
abidlabs, line 41:
```suggestion
Gradio has a strong focus on creating demos and interfaces for machine learning models and makes it easy to build a UI for a model. Gradio also offers a lower-level abstraction known as Blocks for more general web applications. 
```

wwwillchen:
Thanks @abidlabs for the suggestions - I've incorporated them.

PR #640, opened 2024-07-19
Add link component

### LINKED ISSUE #236: Add a link component

wwwillchen:
Fixes #236

### DIFF
--- a/mesop/components/link/link.scss
+++ b/mesop/components/link/link.scss
@@ -1,0 +1,3 @@
+a {
+  display: block;
+}
--- a/mesop/components/link/link.ts
+++ b/mesop/components/link/link.ts
@@ -1,24 +1,26 @@
 import {ChangeDetectionStrategy, Component, Input} from '@angular/core';
 import {
   Key,
+  Style,
   Type,
   UserEvent,
 } from 'mesop/mesop/protos/ui_jspb_proto_pb/mesop/protos/ui_pb';
 import {LinkType} from 'mesop/mesop/components/link/link_jspb_proto_pb/mesop/components/link/link_pb';
 import {Channel} from '../../web/src/services/channel';
+import {formatStyle} from '../../web/src/utils/styles';
 
 @Component({
   selector: 'mesop-link',
   templateUrl: 'link.ng.html',
+  styleUrl: 'link.css',
   standalone: true,
 })
 export class LinkComponent {
   @Input({required: true}) type!: Type;
   @Input() key!: Key;
+  @Input() style!: Style;
   private _config!: LinkType;
   isChecked = false;
-
-  constructor(private readonly channel: Channel) {}
 
   ngOnChanges() {
     this._config = LinkType.deserializeBinary(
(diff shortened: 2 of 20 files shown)


mesop/components/link/link.ts
```diff
@@ -0,0 +1,38 @@
...
+  templateUrl: 'link.ng.html',
+  styleUrl: 'link.css',
+  standalone: true,
+})
+export class LinkComponent {
+  @Input({required: true}) type!: Type;
+  @Input() key!: Key;
+  @Input() style!: Style;
+  private _config!: LinkType;
+  isChecked = false;
```
richard-to, line 23: Is this needed?

mesop/components/link/link.scss
```diff
@@ -0,0 +1,3 @@
+a {
+  display: block;
```
richard-to, line 2: Curious on the decision to make this a block element by default

richard-to (approved):
🤯 - whoa, is that a link component, I see!

A couple thoughts:

- Do we want to include some of the other a tag attributes (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) into the link API?
- Also would it be possible to wrap the link around a box, to make the box a link? Or in that case, would it be preferred to use a click event with me.navigate?

mesop/components/link/link.scss
wwwillchen: I was on the fence because link is normally inline. But because text is a block element, it seems good to be consistent between text and link since link is kind of like a text.

There's also something about writing UIs in Mesop/Python that makes it seem more natural to treat these things as blocks because it resembles the code structure, for example:

```py
me.text("before")
me.link(...)
me.text("after
```

vs. in HTML, it's common to put it all on a single line:

```html
<div>before<a>link</a>after</div>
```

This is the part I've been unsure about because things like  for example are also inline by default and I've left it as inline. WDYT?

richard-to: I think because the text component is block, I think link as block is ok since if the user wanted to to make them flow together, they'd have to change both. 

I don't feel strongly either way.

mesop/components/link/link.ts
wwwillchen: nope - it's from the boilerplate generator; turns out I left it in a bunch of other components (will clean up separately).

wwwillchen:
> 🤯 - whoa, is that a link component, I see!
> 
> A couple thoughts:
> 
> * Do we want to include some of the other a tag attributes (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) into the link API?

I think we could add more later, potentially `download`, but I think this covers the vast majority of use case.

> * Also would it be possible to wrap the link around a box, to make the box a link? Or in that case, would it be preferred to use a click event with me.navigate?

I did try to implement it as something like:

```py
with me.content_link():
  me.text("foo")
```

The problem when I did this though was styling the link didn't work as expected. I think if someone really wanted a box to be a link they could use click+me.navigate, although we still don't support absolute navigation yet.

PR #642, opened 2024-07-19
Do not use class variables in runtime modules

wwwillchen:
- It's better to do the type annotations inside `__init__` instead of annotating at the class variable.
- Initializing class variables is sketchy, particularly with mutable objects.
- We were previously initializing class variables to mutable values on Runtime, which is technically not too bad because Runtime is a singleton object, but this is still a bad practice.


richard-to (approved):
Oh wow, that's a great catch. Can't believe I never noticed that.

PR #646, opened 2024-07-19
Allow data URL for font-src

richard-to:
In the Code Mirror webcomponent, I noticed that Code Mirror uses data URLs for some fonts. This was causing some CSP errors related to fonts.

So adding `data:` to font-src section.


wwwillchen (approved):
font-src isn't very security sensitive so LGTM.

PR #652, opened 2024-07-20
Enable RUF lint rules

wwwillchen:
https://docs.astral.sh/ruff/rules/#ruff-specific-rules-ruf

In particular, it contains https://docs.astral.sh/ruff/rules/mutable-class-default/ which would catch the issue that was fixed in #642.

### DIFF
--- a/mesop/components/table/e2e/table_app.py
+++ b/mesop/components/table/e2e/table_app.py
@@ -57,5 +57,5 @@ def on_click(e: me.TableClickEvent):
   state = me.state(State)
   state.selected_cell = (
     f"Selected cell at col {e.col_index} and row {e.row_index} "
-    f"with value {str(df.iat[e.row_index, e.col_index])}"
+    f"with value {df.iat[e.row_index, e.col_index]!s}"
   )
--- a/mesop/server/state_session_test.py
+++ b/mesop/server/state_session_test.py
@@ -1,3 +1,4 @@
+# ruff: noqa: RUF013
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime, timedelta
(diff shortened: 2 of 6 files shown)


mesop/server/state_session_test.py
```diff
@@ -377,10 +377,10 @@ def __init__(self):
 
   def get(
     self,
-    field_paths: Iterable[str] = None,  # type: ignore
+    field_paths: Optional[Iterable[str]] = None,  # type: ignore
```
richard-to, line 380: Hmm, these were intentional ignores since I was matching the code for the Firebase API. So looks like RUF ignored the type ignore here. Is there a different way to add an ignore for RUF?

I guess in this case, the changes are probably ok. But then we can remove the `type: ignore`

mesop/components/table/e2e/table_app.py
```diff
@@ -57,5 +57,5 @@ def on_click(e: me.TableClickEvent):
   state = me.state(State)
   state.selected_cell = (
     f"Selected cell at col {e.col_index} and row {e.row_index} "
-    f"with value {str(df.iat[e.row_index, e.col_index])}"
+    f"with value {df.iat[e.row_index, e.col_index]!s}"
```
richard-to, line 60: Nice. I didn't know about the `!s` syntax in the formatted string

richard-to (approved):

mesop/server/state_session_test.py
wwwillchen: yup, you can do an ignore with ruff (see: https://github.com/astral-sh/ruff/discussions/3442) - I've ignored this specific error for the file

mesop/components/table/e2e/table_app.py
wwwillchen: me neither!

PR #653, opened 2024-07-20
Provide more explicit docs about mutable state default values

wwwillchen:
I think the docs before were not vocal enough about avoiding this pitfall. I'm also exploring a way of programmatically safe-guarding against mutable default values, but it's a bit tricky - see #647.

### DIFF
--- a/docs/guides/state_management.md
+++ b/docs/guides/state_management.md
@@ -21,6 +21,63 @@ def page():
     me.text(state.val)
 ```
 
+## Use immutable default values
+
+Similar to [regular dataclasses which disallow mutable default values](https://docs.python.org/3/library/dataclasses.html#mutable-default-values), you need to avoid mutable default values such as list and dict for state classes. Using mutable default values can result in leaking state across sessions which can be a serious privacy issue.
+
+You **MUST** use immutable default values _or_ use dataclasses `field` initializer _or_ not set a default value.
+
+???+ success "Good: immutable default value"
+      Setting a default value to an immutable type like str is OK.
+
+      ```py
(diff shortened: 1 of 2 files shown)


docs/guides/state_management.md
```diff
@@ -21,6 +21,71 @@ def page():
     me.text(state.val)
 ```
 
+## Use immutable default values
+
+Similar to [regular dataclasses which disallow mutable default values](https://docs.python.org/3/library/dataclasses.html#mutable-default-values), you need to avoid mutable default values such as list and dict for state classes. Using mutable default values can result in leaking state across sessions which can be a serious privacy issue.
```
richard-to, line 26: > you need to avoid mutable default values such as list and dict for state classes.

This could be accidentally misinterpreted as you can't use list/dict or any mutable data types in State classes. But I think it should be fine if they read  the whole section. And I think probably better to err on the side of caution. And I'm not sure how else you could word it in a succinct and clear way

richard-to (approved):
I think this looks better and clearer.

I did noticed in 3.11, it says dataclass will now check if the data type is hashable when determining if an object is a mutable default value or not.

wwwillchen:
> I think this looks better and clearer.
> 
> I did noticed in 3.11, it says dataclass will now check if the data type is hashable when determining if an object is a mutable default value or not.

good to know! I think we can try to port this change into dataclass_utils

docs/guides/state_management.md
wwwillchen: yeah, I could see that but with the examples, it should be clear they can use list.

PR #658, opened 2024-07-21
Catch accidental usages of mutable default state values

### LINKED ISSUE #644: Prevent accidental usage of mutable classes in @me.stateclass

wwwillchen:
Fixes #644.

Note: this will raise an exception if someone tries to use a proto type in a state class. Before this, we weren't throwing an exception but we were silently not properly serializing & deserializing the proto state (effectively not updating the proto). We'll need to call this out when we sync downstream, and maybe we should support protos properly, but this is at least more explicit.

### DIFF
--- a/mesop/dataclass_utils/dataclass_utils.py
+++ b/mesop/dataclass_utils/dataclass_utils.py
@@ -69,13 +69,15 @@ def dataclass_with_defaults(cls: Type[C]) -> Type[C]:
       elif get_origin(type_hint) == dict:
         setattr(cls, name, field(default_factory=dict))
       elif isinstance(type_hint, type):
-        if is_dataclass(cls) or has_parent(type_hint):
+        if has_parent(type_hint):
           setattr(cls, name, field(default_factory=type_hint))
         else:
           setattr(
             cls, name, field(default_factory=dataclass_with_defaults(type_hint))
           )
-
+  # If a class is already a dataclass, then don't wrap it with another dataclass decorator.
+  if is_dataclass(cls):
+    return cls
   return dataclass(cls)
 
 
--- a/mesop/dataclass_utils/dataclass_utils.py
+++ b/mesop/dataclass_utils/dataclass_utils.py
@@ -9,7 +9,7 @@ from deepdiff import DeepDiff, Delta
 from deepdiff.operator import BaseOperator
 from deepdiff.path import parse_path
 
-from mesop.exceptions import MesopException
+from mesop.exceptions import MesopDeveloperException, MesopException
 
 _PANDAS_OBJECT_KEY = "__pandas.DataFrame__"
 _DATETIME_OBJECT_KEY = "__datetime.datetime__"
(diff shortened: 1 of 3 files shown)


mesop/dataclass_utils/dataclass_utils_test.py
```diff
@@ -94,6 +97,19 @@ class RootDataclass:
...
+
+  @dataclass_with_defaults
+  class RootDataclass:
+    nested: NestedDataclass
+
+  instance = RootDataclass()
+  assert instance.nested.val == "default"
+
+
 def test_dataclass_defaults_works_with_already_annotated_nested_class_fails_without_default():
```
richard-to, line 113: [question] This is a pre-existing test name, but I'm a bit confused by the test name here. Is this testing something that is valid or that fails? Seems like something valid.

mesop/dataclass_utils/dataclass_utils_test.py
```diff
@@ -94,6 +97,19 @@ class RootDataclass:
   assert instance.nested.val == "<default>"
 
 
+def test_dataclass_defaults_works_with_already_annotated_nested_class_requires_default():
```
richard-to, line 100: [question] What is meant by requires default?

mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -54,13 +75,29 @@ def dataclass_with_defaults(cls: Type[C]) -> Type[C]:
...
-        setattr(
-          cls, name, field(default_factory=dataclass_with_defaults(type_hint))
-        )
-
+        if has_parent(type_hint):
+          # If this isn't a simple class (i.e. it inherits from another class)
+          # then we will preserve its semantics (not try to set default values
+          # because it's not a dataclass) and instantiate it with each new instance
+          # of a state class.
+          setattr(cls, name, field(default_factory=type_hint))
```
richard-to, line 83: What about the case where the constructor requires that certain parameters be set? I'm assuming an exception would be thrown in that case?

mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -54,13 +75,29 @@ def dataclass_with_defaults(cls: Type[C]) -> Type[C]:
...
+        else:
+          # If it's a simple dataclass (i.e. does not inherit from another class)
+          # then we will try to set default values and then wrap it with a dataclass
+          # decorator (if necessary).
+          setattr(
```
richard-to, line 88: Why do we try to wrap it in a dataclass if the class is not a dataclass? Does this affect serialization/deserialization?

richard-to (approved):
Looks like an improvement. But I did have a few questions about the implementation.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: yeah - an exception will be thrown - if certain parameters are required, then they will need to write a custom default factory.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: we wrap it, so 1) you can serialize/deserialize it and 2) dataclasses are generally convenient b/c it auto-generates stuff like __init__() and __repr__() - see https://docs.python.org/3/library/dataclasses.html

mesop/dataclass_utils/dataclass_utils_test.py
wwwillchen: i've removed it from the test method name b/c it's confusing. basically i was trying to convey that every field needs to have a default value - either explicitly configured by the mesop app developer or implicitly, via the framework - because we instantiate the stateclass with no parameters passed in.

mesop/dataclass_utils/dataclass_utils_test.py
wwwillchen: it's basically testing that NestedDataclass which is already annotated with @dataclass fails because it doesn't have a default value.

PR #660, opened 2024-07-22
Add example headers to patterns demos.

richard-to:
A common use case to have a header for the app. This example was extracted from the prompt tuner console app that I've been working on.

Also think it will be useful for the Mesop AI stuff.

# Full width

Screenshot 2024-07-22 at 1 25 51 PM

# Mobile
Screenshot 2024-07-22 at 1 25 59 PM

### DIFF
--- a/demo/headers.py
+++ b/demo/headers.py
@@ -1,0 +1,193 @@
+from dataclasses import fields
+
+import mesop as me
+
+
+@me.page(
+  security_policy=me.SecurityPolicy(
+    allowed_iframe_parents=["https://google.github.io"]
+  ),
+  path="/headers",
+)
+def app():
+  is_mobile = me.viewport_size().width < 640
+
+  with me.box(style=me.Style(margin=me.Margin(bottom=15))):
+    # Two section basic header with fluid width.
+    # As an example, we don't use mobile view here since the header is short enough.
+    with header(max_width=None):
+      with header_section():
+        me.text(
+          "Mesop", type="headline-6", style=me.Style(margin=me.Margin(bottom=0))
+        )
+
+      with header_section():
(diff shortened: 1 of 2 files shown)


demo/headers.py
```diff
@@ -0,0 +1,192 @@
...
+def app():
+  is_mobile = me.viewport_size().width < 640
+
+  with me.box(style=me.Style(margin=me.Margin(bottom=15))):
+    # Two section basic header with fluid width.
+    # As an example, we don't use mobile view here since the header is short enough.
+    with header(max_width=None):
+      with header_section():
+        me.text(
+          "Mesop", type="headline-6", style=me.Style(margin=me.Margin(bottom=0))
```
wwwillchen, line 21: side-note: I've wondered whether we should just reset the default margins for headings to 0. I find it quite annoying to set it to margin 0, and it would improve cross-browser consistency.

demo/headers.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@me.content_component
+def header_section():
+  """Adds a section to the header."""
+  with me.box(style=me.Style(display="flex", gap=5)):
+    me.slot()
+
+
+def merge_styles(
```
wwwillchen, line 147: this is interesting :)

wwwillchen (approved):
nice!

demo/headers.py
richard-to: Yeah that's an interesting idea. On one hand, I definitely agree that the default headline margins are kind of absurdly large.

I think having some sensible margin bottom is kind of good, so it doesn't always need to be tweaked. Since 0 probably wouldn't be great in all cases.

PR #663, opened 2024-07-23
Support dark theming

### LINKED ISSUE #461: Support dark theme in Mesop
### Discussed in https://github.com/google/mesop/discussions/446

<div type='discussions-op-text'>

<sup>Originally posted by **pranshu-raj-211** June 17, 2024</sup>
I'm making a couple of websites using mesop, and I'd like to implement a dark mode toggle, cards among other stuff. Could you point me to some example in this repo where something similar has been done, so that I can make it and submit a PR?

Thanks for the wonderful work that you're doing!</div>

---

We should expose a way to use Angular Material dark theme: https://material.angular.io/guide/theming#customizing-your-colors

wwwillchen:
This adds support for dark theming at the framework level and adds dark theme support specifically for the chat component as an example of how to use it. Closes #461.

This doesn't support theming to different colors which is a lot more work and I've filed a separate issue at #669.

Considerations:

- Right now a lot of Mesop apps may not be dark-theme compatible, so we need to be keep the default as light theme. Eventually, we'd like to have Mesop apps default to system theme mode (respect user preferences for dark or light theme)

### DIFF
--- a/demo/chat.py
+++ b/demo/chat.py
@@ -5,12 +5,19 @@ import mesop as me
 import mesop.labs as mel
 
 
+def on_load(e: me.LoadEvent):
+  me.set_theme_mode(
+    me.ThemeMode.DARK if e.prefers_dark_theme else me.ThemeMode.LIGHT
+  )
+
+
 @me.page(
   security_policy=me.SecurityPolicy(
     allowed_iframe_parents=["https://google.github.io"]
   ),
   path="/chat",
   title="Mesop Demo Chat",
+  on_load=on_load,
 )
 def page():
   mel.chat(transform, title="Mesop Demo Chat", bot_user="Mesop Bot")
--- a/demo/chat.py
+++ b/demo/chat.py
@@ -6,9 +6,7 @@ import mesop.labs as mel
 
 
 def on_load(e: me.LoadEvent):
-  me.set_theme_mode(
-    me.ThemeMode.DARK if e.prefers_dark_theme else me.ThemeMode.LIGHT
-  )
+  me.set_theme_mode("system")
 
 
 @me.page(
--- a/docs/guides/theming.md
+++ b/docs/guides/theming.md
@@ -1,0 +1,69 @@
+# Theming
+
+Mesop has early-stage support for theming so you can support light theme and dark theme in a Mesop application.
+
+## Example
+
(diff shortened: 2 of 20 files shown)


richard-to (approved):
LGTM - Nice. I'm slightly ambivalent about leveraging the material theming (but don't have anything concrete). But the general interface makes sense and seems flexible enough. I think it's one of those things where I'll have to try it out to get some impressions on how it is to actually use.

Seems like it will be fine so long the user leverages the the various exposed theme colors. This adds some slight learning curve I think. Now the user has to keep checking what the variable names mean and what color it will render (at least initially). At the same time, this could save time for people choosing colors of their own.

For cases where the user wants a very different color palette, then issue 669 does seem like it would be very welcomed. 

And I guess technically the user does not need to actually support a dark/light themes if they don't want to.

wwwillchen:
> LGTM - Nice. I'm slightly ambivalent about leveraging the material theming (but don't have anything concrete). But the general interface makes sense and seems flexible enough. I think it's one of those things where I'll have to try it out to get some impressions on how it is to actually use.
> 
> Seems like it will be fine so long the user leverages the the various exposed theme colors. This adds some slight learning curve I think. Now the user has to keep checking what the variable names mean and what color it will render (at least initially). At the same time, this could save time for people choosing colors of their own.
> 
> For cases where the user wants a very different color palette, then issue 669 does seem like it would be very welcomed.
> 
> And I guess technically the user does not need to actually support a dark/light themes if they don't want to.

I think the material theming is quite powerful and we basically need to use it because we use Angular Material components. There's definitely a learning curve, when I was working on the chat component, it took a bit of tinkering and looking up the CSS variable colors in Chrome DevTools - I think this is probably unrealistic to expect most Mesop app developers to do. Eventually, this is where I think something like Visual Editor can play a very helpful role, you really want to see the available palette of colors, and not just the variable names, to know what to pick. Also, Material provides _many_ colors, but in reality there's probably just a small handful that you'll use in most applications, so we can try to recommend those in the visual editor (or maybe docs?)

My guess is that we'll need to provide good examples / components (e.g. chat) that do proper dark/light theming and people can probably extend them (i.e. copy and paste). Otherwise doing proper dark theming is non-trivial.

Issue #665, opened 2024-07-24
Better default markdown support


adriens:
# :grey_question: About

I use mesop to interact with an AI Agent API that is able to reply complex markdown... with lists and sub-items... but when display its answers I got sub-item appearing at the wrong level : 

![image](https://github.com/user-attachments/assets/27129e16-b442-4acd-ab3a-89d894df210a)

In other words, sections are messed-up and all go to the same level

# :pray: Feature request

Make mesop better handle list and sublists.

For now, the only workaround I foud is to asj my agent to reply in markdown in raw format... causing the final display to be right but properly displayed (I have to copy/paste it into any markdown previewer to enjoy it).

richard-to:
Thanks for pointing that out. As an experiment I've been thinking of creating a web component that renders markdown on the client side, which I think could improve https://github.com/google/mesop/issues/591. 

So first I want to try it as a web component to see if it works. And if it does work, I think we could update the core markdown component to also render markdown on the front end.

So it's possible using the markdown web component could render the markdown better.

------

On the other hand, it is strange that the markdown is not being rendered correctly for you, especially a simple thing like nested lists. I'll see if I can reproduce it.

Do you have an example output for testing?

richard-to:
Yeah I can confirm that the sub list is not working. This can be seen also in the demo gallary for the markdown example. The sublist is not indented.

richard-to:
So looks like the reason is the markdown module we're using requires  four space indention. See https://github.com/Python-Markdown/markdown/issues/1378

This is expected behavior for this module. 

-----

I'd agree that this isn't ideal. As a workaround is it possible to tell the agent to render the sublist with four spaces?

adriens:
> I'd agree that this isn't ideal. As a workaround is it possible to tell the agent to render the sublist with four spaces?

In my case, I asked my agent to generate a `markdown` so I can put it in Github issue... and let him do the job.

richard-to:
We're planning to update the markdown object to render using javascript instead of python. This has the side effect of allowing two spaces for nested lists.

adriens:
great news ! We did a demo yesterday and enjoyed a lot the Dark theme toggle btw. I made an internal issue to force `markdown` output... so we'll wait for the next release.

Thanks for the very cool feedback and the resulting devex 🙏

adriens:
here is the yesterday pari-programming session : 
- https://github.com/google/mesop/issues/816

richard-to:
[...]

Issue #591, opened 2024-07-09
mel.chat gets very slow as more and more messages are added


richard-to:
See discussion in https://github.com/google/mesop/discussions/578

richard-to:
On the server, I think the bottle neck is the `runtime().run_path(path=path, trace_mode=trace_mode)` step in the render_loop. 

It's the one part that seems to get slower as more chat messages are added. It can also run multiple times per loop (if render_loop is called multiple times).

```
Run path took 2.07 ms
Run path took 23.95 ms
Run path took 13.17 ms
Run path took 2.33 ms
Run path took 1.28 ms
Run path took 2.89 ms
Run path took 1.78 ms
Run path took 2.02 ms
Run path took 2.12 ms
Run path took 4.20 ms
Run path took 9.52 ms
Run path took 4.86 ms
Run path took 4.06 ms
Run path took 6.63 ms
Run path took 7.76 ms
Run path took 12.51 ms
Run path took 7.59 ms
Run path took 6.71 ms
Run path took 10.55 ms
Run path took 10.56 ms
Run path took 16.05 ms
Run path took 10.56 ms
Run path took 9.18 ms
Run path took 12.23 ms
Run path took 13.28 ms
Run path took 11.30 ms
Run path took 12.44 ms
Run path took 12.19 ms
Run path took 13.89 ms
Run path took 15.27 ms
Run path took 22.88 ms
Run path took 14.18 ms
Run path took 13.72 ms
Run path took 17.43 ms
Run path took 17.63 ms
Run path took 24.96 ms
Run path took 16.84 ms
Run path took 15.55 ms
Run path took 22.88 ms
Run path took 18.88 ms
Run path took 29.28 ms
Run path took 18.47 ms
Run path took 19.85 ms
Run path took 24.90 ms
Run path took 21.01 ms
Run path took 34.92 ms
Run path took 20.80 ms
Run path took 21.60 ms
Run path took 28.37 ms
Run path took 23.13 ms
Run path took 33.91 ms
Run path took 21.84 ms
Run path took 23.68 ms
Run path took 29.98 ms
Run path took 25.43 ms
```

The other bottleneck is simply more data being sent to the server:


# Earlier example

<img width="350" alt="Screenshot 2024-07-11 at 3 29 03 PM" src="https://github.com/user-attachments/assets/a6d0f6a0-425c-4f13-98de-26c112961c90">

# Later example

[...]

PR #667, opened 2024-07-24
Add example webcomponent that uses markedjs, highlightjs, and mermaidjs

richard-to:
This component uses MarkedJS for markdown, Mermaid for diagrams, and HighlightJS for code highlighting support.

- One issue is that I kept running into a duplicate trusted type issue with lit-html, so I needed to allow duplicates, but not sure if that's a good thing or not.
- This is an example workaround for the markdown issue that requires four indents for nested lists: https://github.com/google/mesop/issues/665
- Also curious to test this out with mel.chat  (replacing the me.markdown with this element) (https://github.com/google/mesop/issues/591)

# Screenshot

Screenshot 2024-07-26 at 4 51 28 PM

### DIFF
--- a/mesop/server/static_file_serving.py
+++ b/mesop/server/static_file_serving.py
@@ -159,7 +159,7 @@ def configure_static_file_serving(
         "style-src": "'self' 'unsafe-inline' fonts.googleapis.com",
         "script-src": f"'self' 'nonce-{g.csp_nonce}'",
         # https://angular.io/guide/security#enforcing-trusted-types
-        "trusted-types": "angular angular#unsafe-bypass lit-html 'allow-duplicates'",
+        "trusted-types": "angular angular#unsafe-bypass lit-html",
         "require-trusted-types-for": "'script'",
       }
     )
--- a/mesop/server/static_file_serving.py
+++ b/mesop/server/static_file_serving.py
@@ -159,7 +159,7 @@ def configure_static_file_serving(
         "style-src": "'self' 'unsafe-inline' fonts.googleapis.com",
         "script-src": f"'self' 'nonce-{g.csp_nonce}'",
         # https://angular.io/guide/security#enforcing-trusted-types
-        "trusted-types": "angular angular#unsafe-bypass lit-html",
+        "trusted-types": "angular angular#unsafe-bypass lit-html 'allow-duplicates'",
         "require-trusted-types-for": "'script'",
       }
     )
--- a/mesop/examples/web_component/BUILD
+++ b/mesop/examples/web_component/BUILD
@@ -12,6 +12,7 @@ py_library(
         "//mesop",
         "//mesop/examples/web_component/code_mirror_editor",
         "//mesop/examples/web_component/firebase_auth",
+        "//mesop/examples/web_component/markedjs",
         "//mesop/examples/web_component/plotly",
         "//mesop/examples/web_component/quickstart",
         "//mesop/examples/web_component/shared_js_module",
--- a/mesop/examples/web_component/__init__.py
+++ b/mesop/examples/web_component/__init__.py
@@ -4,6 +4,9 @@ from mesop.examples.web_component.code_mirror_editor import (
 from mesop.examples.web_component.firebase_auth import (
   firebase_auth_app as firebase_auth_app,
 )
+from mesop.examples.web_component.markedjs import (
+  markedjs_app as markedjs_app,
+)
 from mesop.examples.web_component.plotly import (
   plotly_app as plotly_app,
 )
--- a/mesop/examples/web_component/markedjs/BUILD
+++ b/mesop/examples/web_component/markedjs/BUILD
@@ -1,0 +1,14 @@
+load("//build_defs:defaults.bzl", "py_library")
+
+package(
+    default_visibility = ["//build_defs:mesop_examples"],
+)
+
+py_library(
+    name = "markedjs",
+    srcs = glob(["*.py"]),
+    data = glob(["*.js"]),
+    deps = [
+        "//mesop",
+    ],
+)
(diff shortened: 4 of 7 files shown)


mesop/server/static_file_serving.py
```diff
@@ -159,7 +159,7 @@ def add_security_headers(response: Response):
         "style-src": "'self' 'unsafe-inline' fonts.googleapis.com",
         "script-src": f"'self' 'nonce-{g.csp_nonce}'",
         # https://angular.io/guide/security#enforcing-trusted-types
-        "trusted-types": "angular angular#unsafe-bypass lit-html",
+        "trusted-types": "angular angular#unsafe-bypass lit-html 'allow-duplicates'",
```
richard-to, line 162: This is one issue I ran into. For some reason lit-html was getting added twice I think because I needed to also import unsafeHTML. And so I added allow-duplicates. Not sure if that's the best idea or not.

wwwillchen: I think we can avoid the trusted types duplicates issue by importing Lit from the exact same URL across all our web components (which is a good thing in general, to avoid loading Lit multiple times).
Right now we import Lit from 'https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js'

In this PR, you're importing Lit from 'https://cdn.jsdelivr.net/npm/lit@3.1.4/+esm';

I think you could try swapping out the esm URL for the first URL which we already use and then you won't load Lit twice. I'm not sure if you had to use the +esm directive.

Alternatively, you could swap out **all** the usages of  'https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js'  with 'https://cdn.jsdelivr.net/gh/lit/dist@3/all/lit-all.min.js' which should include the unsafe directive you need.

BTW, these are the bundle URLs recommended by the Lit docs: https://lit.dev/docs/getting-started/#use-bundles - I'm not sure what the differences are w/ the esm one, but the regular ones (from the docs) seems to work fine with ES modules.

wwwillchen:
I think this is definitely a good example for a web component, but before we replace it in mel.chat, a couple considerations:

1. You'll need to do extra work to support syntax highlighting which is what motivated the initial switch from markedjs to python-markdown
2. Using marked.js as a web component makes it harder to fork mel.chat (because then you'll need to copy multiple files). 

On a related tangent, I've been thinking _maybe_ we can have two chat components in mesop.labs. We could rename the current chat component to something like "basic_chat" and basically keep the functionality as-is. The goal for basic_chat is to keep the implementation simple (e.g. single file, relatively straightforward).

Then, we'd create a new component called "chat" which has the same API (or potentially a superset) which has a more optimized UI - e.g. it could use marked.js, have a web component for the textarea to support keyboard shortcuts like "enter" vs "shift+enter", example boxes (optional), and an overall more published UX like the Gemini chat example we have internally. WDYT?

mesop/server/static_file_serving.py
richard-to: Yeah, this is one of the pain points of working with the web components in Mesop. I feel like I spend most of my time trying to figure out how to get the imports to work.

I think the issue with `https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js` this URL is that it does not include unsafeHTML, which I needed in this case. So the only alternative would be to use the `lit-all.min.js` I think. I did try to include the unsafe HTML directive directly along with lit-core.min.js, but it didn't seem to find it.

Which is how I ended up with using the esm modules. Also when I search jsdelivr, the gh URL does not show up in the results.

However, I don't know if this import https://cdn.jsdelivr.net/npm/lit@3.1.4/+esm, is the same as lit-core.min.js in terms of what it includes.

Also I think when I import https://cdn.jsdelivr.net/npm/lit@3.1.4/+esm, it does not trigger the duplicate CSP issue for lit-html. iirc only does it when I include the unsafe-html directive.

--------

But I'll give give it another try.

richard-to:
> I think this is definitely a good example for a web component, but before we replace it in mel.chat, a couple considerations:
> 
> 1. You'll need to do extra work to support syntax highlighting which is what motivated the initial switch from markedjs to python-markdown
> 2. Using marked.js as a web component makes it harder to fork mel.chat (because then you'll need to copy multiple files).
> 
> On a related tangent, I've been thinking _maybe_ we can have two chat components in mesop.labs. We could rename the current chat component to something like "basic_chat" and basically keep the functionality as-is. The goal for basic_chat is to keep the implementation simple (e.g. single file, relatively straightforward).
> 
> Then, we'd create a new component called "chat" which has the same API (or potentially a superset) which has a more optimized UI - e.g. it could use marked.js, have a web component for the textarea to support keyboard shortcuts like "enter" vs "shift+enter", example boxes (optional), and an overall more published UX like the Gemini chat example we have internally. WDYT?

Good point about the syntax highlighting. I didn't realize you started with markedjs and moved to python-markdown. That's very interesting.

In terms of mel.chat, yes, my original thinking was that if using markedjs actually did improve the performance rendering the markdown, that the core markdown component should use markedjs, but now that I know that you originally used markedjs, then maybe that won't work well. 

But yeah my main reason for adding this web component is that the Prompt Tuner app I'm working on could potentially render a lot of markdown boxes and I'm worried the Eval page will be very slow, hence the experimenting with this web component.

I like the idea of the optimized UI, especially the custom text area support

mesop/server/static_file_serving.py
richard-to: In the end, it seems I needed to set `dangerously_disable_trusted_types=True` for DomPurifier so the duplicate lit-html didn't matter anymore. But I'll keep an eye on this when I add more web components.

richard-to:
Ok, I added code highlighting. It uses highlightjs since Marked doesn't have highlighting builtin.

wwwillchen (approved):
LGTM, but don't forget to rename the files and PR to say highlightjs

PR #675, opened 2024-07-26
Prevent page or web component from being registered after traffic has been served

wwwillchen:
We want to avoid an anti-pattern where Mesop app developers are defining a page or web component dynamically after _any_ traffic has been served (page and web components are registered globally as part of the runtime singleton, and are not scoped to a specific context/request). 

For example, we want to prevent the following:

```py
@me.page()
def p():
   s = me.state(State)

  @me.page(security_policy=me.SecurityPolicy(allowed_frame_parents=[s.user_input]))
  def dynamic_page():
     pass
```

### DIFF
--- a/mesop/examples/testing/__init__.py
+++ b/mesop/examples/testing/__init__.py
@@ -8,6 +8,12 @@ from mesop.examples.testing import (
   csp_escaping as csp_escaping,
 )
 from mesop.examples.testing import (
+  error_register_page_too_late as error_register_page_too_late,
+)
+from mesop.examples.testing import (
+  error_register_web_component_too_late as error_register_web_component_too_late,
+)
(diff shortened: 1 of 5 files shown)


richard-to (approved):
Interesting. 

What is security issue with the late registration?

It's ok to late register user defined components? What is the difference?

wwwillchen:
> Interesting.
> 
> What is security issue with the late registration?
> 
> It's ok to late register user defined components? What is the difference?

The issue with late registration for web components is that it could result in a user loading a local JS file (which the mesop app developer didn't intend to serve).

I can't think of a specific issue with late-registering user defined components, but it seems pretty weird to do this, so I'm open to prohibiting it too (mostly for consistency and encouraging good practices).

PR #676, opened 2024-07-26
Support query params

### LINKED ISSUE #340: Retrieve query params from URL to use in Mesop
Right now it's not possible to get the query params from the URL. 

Even request.args from flask would not work in this case since Mesop gets called via the `/ui` endpoint so the initial query params wouldn't be available in the flask request.

wwwillchen:
Fixes #340.

This provides query param support using a dict-like API with inspiration from the [Web API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) and [Streamlit's API](https://docs.streamlit.io/develop/api-reference/caching-and-state/st.query_params#stquery_paramsget_all).

### DIFF
--- a/docs/api/query_params.md
+++ b/docs/api/query_params.md
@@ -1,0 +1,96 @@
+# Query Params API
+
+## Overview
+
+Query params, also sometimes called query string, provide a way to manage state in the URLs. They are useful for providing deep-links into your Mesop app.
+
+## Example
+
+Here's a simple working example that shows how you can read and write query params.
+
+```py
+@me.page(path="/examples/query_params/page_2")
+def page_2():
+  me.text(f"query_params={me.query_params}")
+  me.button("Add query param", on_click=add_query_param)
+  me.button("Navigate", on_click=navigate)
+
+def add_query_param(e: me.ClickEvent):
+  me.query_params["key"] = "value"
+
+def navigate(e: me.ClickEvent):
+  me.navigate("/examples/query_params", query_params=me.query_params)
+```
+
+## Usage
+
+You can use query parameters from `me.query_params`, which has a dictionary-like interface, where the key is the parameter name and value is the parameter value.
+
+### Get a query param value
+
+```py
+value: str = me.query_params['param_name']
+```
+This will raise a KeyError if the parameter doesn't exist. You can use `in` to check whether a key exists in `me.query_params`:
+
+```py
+if 'key' in me.query_params:
+    print(me.query_params['key'])
+```
+
+???+ NOTE "Repeated query params"
+    If a query param key is repeated, then you will get the _first_ value. If you want all the values use `get_all`.
+
+### Get all values
+
+To get all the values for a particular query parameter key, you can use `me.query_params.get_all`, which returns a sequence of parameter values (currently implemented as a `tuple`).
+
+```py
+all_values = me.query_params.get_all('param_name')
+```
+
+### Iterate
+
+```py
+for key in query_params:
+  value = query_params[key]
+```
+
+### Set query param
+
+```py
+query_params['new_param'] = 'value'
+```
+
+### Set repeated query param
+
+```py
+query_params['repeated_param'] = ['value1', 'value2']
+```
+
+### Delete
+
+```py
+del query_params['param_to_delete']
+```
+
+## Patterns
+
+### Navigate with existing query params
+
+Here's an example of how to navigate to a new page with query parameters:
+
+```py
+def click_navigate_button(e: me.ClickEvent):
+    me.query_params['q'] = "value"
+    me.navigate('/search', query_params=me.query_params)
+```
+
+### Navigate with only new query params
+
+You can also navigate by passing in a dictionary to `query_params` parameter for `me.navigate` if you do _not_ want to keep the existing query parameters.
+
+```py
+def click_navigate_button(e: me.ClickEvent):
+    me.navigate('/search', query_params={"q": "value})
+```
--- a/mesop/commands/navigate.py
+++ b/mesop/commands/navigate.py
@@ -1,11 +1,39 @@
+from typing import Sequence
+
+from mesop.features.query_params import (
+  QueryParams,
+)
+from mesop.features.query_params import (
+  query_params as me_query_params,
+)
 from mesop.runtime import runtime
+from mesop.utils.url_utils import remove_url_query_param
+from mesop.warn import warn
 
 
-def navigate(url: str) -> None:
+def navigate(
+  url: str,
+  *,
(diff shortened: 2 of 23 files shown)


docs/api/query_params.md
```diff
@@ -0,0 +1,96 @@
...
+### Set query param
+
+```py
+query_params['new_param'] = 'value'
+```
+
+### Set repeated query param
+
+```py
+query_params['repeated_param'] = ['value1', 'value2']
```
richard-to, line 68: For repeated params, is there a way to append to it. Or do I need to set the whole list each time? Seems like that latter, which I think is fine.

docs/api/query_params.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+### Delete
+
+```py
+del query_params['param_to_delete']
```
richard-to, line 74: Same question here. There's no way to delete one item from a repeated query param I'm assuming?

docs/api/query_params.md
```diff
@@ -0,0 +1,96 @@
...
+    me.navigate('/search', query_params=me.query_params)
+```
+
+### Navigate with only new query params
+
+You can also navigate by passing in a dictionary to `query_params` parameter for `me.navigate` if you do _not_ want to keep the existing query parameters.
+
+```py
+def click_navigate_button(e: me.ClickEvent):
+    me.navigate('/search', query_params={"q": "value})
```
richard-to, line 95: But this wouldn't support repeated fields. Could I also create the QueryParams object on my own or is tied to context object? 

-----

ah nvm, I guess the value can also be list[str]

----

ah nvm (I probably should have read further down into the code first), I guess the context is needed so the updated query params can be forwarded to the client to be updated.

mesop/commands/navigate.py
```diff
@@ -1,11 +1,37 @@
...
+      "Used me.navigate to navigate to a URL with query params. The query params have been removed. "
+      "Instead pass the query params using the keyword argument like this: "
+      "me.navigate(url, query_params={'key': 'value'})"
+    )
+  if isinstance(query_params, QueryParams):
+    query_params = {key: query_params.get_all(key) for key in query_params}
+
+  # Clear the query params because the query params will
+  # either be replaced with the new query_params or emptied.
+  me_query_params.clear()
```
richard-to, line 36: I found this slightly confusing. Took me a bit to realize these where the query params for the current page.

mesop/examples/query_params.py
```diff
@@ -0,0 +1,92 @@
+import mesop as me
+
+
+def on_load(e: me.LoadEvent):
+  me.query_params["on_load"] = "loaded"
+
+
+@me.page(path="/examples/query_params", on_load=on_load)
```
richard-to, line 8: [nit] May want to add some tests that include url encoded params.

mesop/features/query_params.py
```diff
@@ -0,0 +1,38 @@
+from typing import Callable, Iterator, MutableMapping
+
+from mesop.runtime import runtime
+from mesop.runtime.context import Context
+
+
+class QueryParams(MutableMapping[str, str]):
+  def __init__(self, get_context: Callable[[], Context]):
+    self._get_context = get_context
```
richard-to, line 9: [question] I'm very curious why this depends on context.

mesop/runtime/context.py
```diff
@@ -37,15 +37,65 @@ def __init__(
...
+        )
+        for key, value in query_params.items()
+      ]
+    # Construct the full URL with query parameters
+    full_url = url
+    if query_params:
+      query_string = "&".join(
+        f"{key}={value}"
+        if isinstance(value, str)
+        else "&".join(f"{key}={v}" for v in value)
```
richard-to, line 87: I think key/value may need to be url encoded and the key may need to be possible validated (is there a specific format the key must follow?) 

There is one case where a user may use "&" as a value.

mesop/utils/url_utils.py
```diff
@@ -18,4 +20,25 @@ def _escape_url_for_csp(url: str) -> str:
...
+    for param in query_param.split("&"):
+      if "=" in param:
+        key, value = param.split("=", 1)
+        if key in param_dict:
+          param_dict[key].append(value)
+        else:
+          param_dict[key] = [value]
+      else:
+        # Handle cases where there's a key without a value
+        param_dict[param] = []
```
richard-to, line 39: For the case of `var1=&var2=2, I noticed that that var1 would become var1='' empty string `{'var1': ['']}`. Is that expected behavior -- I guess the differentiator between `var1&` and `var1=&`?

mesop/web/src/utils/query_params.ts
```diff
@@ -0,0 +1,22 @@
+import {QueryParam} from 'mesop/mesop/protos/ui_jspb_proto_pb/mesop/protos/ui_pb';
+
+export function getQueryParams(): QueryParam[] {
```
richard-to, line 3: [optional] Maybe add a test spec for this function?

richard-to (approved):
Nice. I'm sure some people will be really happy to see this feature arrive.

docs/api/query_params.md
wwwillchen: thx for the detailed review - for these first few points, I'm wondering if I should have query_params.get_all return a tuple (immutable) instead of a list, this way developers won't be tempted to delete an element from the list (which would do nothing). WDYT?

wwwillchen: will return a tuple but type it as a sequence for future flexibility.

docs/api/query_params.md
wwwillchen: need to set the whole list (now a tuple to make it clearer it's not mutable)

mesop/commands/navigate.py
wwwillchen: mentioned this is done in server.py

mesop/examples/query_params.py
wwwillchen: good idea-done.

mesop/runtime/context.py
wwwillchen: Good callout - I added a test case, turns out it's handled by URLSearchParams ([spec](https://url.spec.whatwg.org/#urlsearchparams-stringification-behavior)) on the TS side. This is nice because the Python side doesn't need to worry about URL encoding and decoding and we use the browser APIs.

mesop/utils/url_utils.py
wwwillchen: I think so - I tested out the browser API and also get an empty string:
```js
new URLSearchParams('?a=&b=1').forEach((value, key) => {console.log('value', value, 'key', key)})
```

Issue #627, opened 2024-07-17
Add Autocomplete Component from Angular Components


richard-to:
Link: https://material.angular.io/components/autocomplete/api

Issue #610, opened 2024-07-12
Adding a Search box demo


itsBaivab:
**Is your feature request related to a problem? Please describe. **
I saw there is no demo for a search box. It's great to have one for easy intermigration. 

**Describe the solution you'd like**
The search field should be able to search from multiple results. It should be able to dynamically change its results on the basis of current text in the search field.

**Describe alternatives you've considered**

**Additional context**
![image](https://github.com/user-attachments/assets/8e88fc19-961b-4f25-8e8a-caa3fbb874bb)

richard-to:
I think for a search box, we could try adding the Angular Autocomplete component (https://material.angular.io/components/autocomplete/overview). I haven't tested if this component would be easy to add to Mesop or not.

So I think if we were to add search box demo, we'd try to make it look similar to the Angular Autocomplete component.

I do agree that this is a good use case to cover. Thanks for the suggestion.

itsBaivab:
Sure, I will try to make a pr on it.

richard-to:
Going to close this issue since we'll likely be adding the Angular Autocomplete component.

PR #677, opened 2024-07-26
Add basic autocomplete component.

richard-to:
This adds support for the autocomplete Angular Component.

Supported features:

- Option groups

It has some limitations:

1. Does not allow the custom display selections (only text is allow)
2. Key and values are text only
3. Filtering is done on client side but limited for now to "case insensitive contains"
  - Filtering can also be done on the server side by manually filtering the input on input (but we do not explicitly call this out for now)  
4. Had to remove require selection since it does not work well with filtering. There are workarounds online but seemed better to keep it simple for now unless people really want require selection.
5.  
---

Screenshot 2024-07-26 at 3 36 35 PM

Ref: #627
Ref: #610


wwwillchen:
I'm somewhat hesitant about doing the filtering on the server-side because this will run into the same performance issues we ran into w/ input, and I think users expect autocomplete interactions to be quite snappy. What if we did filtering on the client-side and initially started off with a simple default option like - does the query match any of the options (case-insensitive). And then, based on user feedback, we could expand it to more options, e.g. case-sensitive, must be a prefix match, etc. I think the default filtering option could handle a big chunk of use cases and it would be both a better developer experience and user experience. WDYT?

richard-to:
> I'm somewhat hesitant about doing the filtering on the server-side because this will run into the same performance issues we ran into w/ input, and I think users expect autocomplete interactions to be quite snappy. What if we did filtering on the client-side and initially started off with a simple default option like - does the query match any of the options (case-insensitive). And then, based on user feedback, we could expand it to more options, e.g. case-sensitive, must be a prefix match, etc. I think the default filtering option could handle a big chunk of use cases and it would be both a better developer experience and user experience. WDYT?

Yeah I agree that filtering on the client side would be more efficient for the use case where the user has all the data passed to the frontend.

I was just concerned that it didn't cover the use case where the data is too large to store on the frontend, which means needing to do the filtering on the backend.

I think we can definitely do both, have client side filtering running by default + filtering the options in the backend. I think my main concern there was confusion on how where and how the filtering was occurring, so I kept it simpler by doing backend only.

However, I agree that the trade off would be efficiency. And perhaps that is too much of a trade off. And I think the benefit of having the client side filtering it's one less thing the developer would have to worry about.

------

So I'll add the client side filtering back in. I think just need to update the docs and example to be more clear about how filtering works.

mesop/components/autocomplete/autocomplete.py
```diff
@@ -0,0 +1,233 @@
...
+    value=userEvent.string_value,
+    key=key.key,
+  ),
+)
+
+
+@register_native_component
+def autocomplete(
+  *,
+  # This should be Iterable[AutocompleteOption | AutocompleteOptionGroup], but breaks
```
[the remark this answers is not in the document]
richard-to: So the error is this:

```
mesop.components.autocomplete.autocomplete.AutocompleteOption | mesop.components.autocomplete.autocomplete.AutocompleteOptionGroup is not a callable object

Traceback
//mesop/mesop/server/server.py:68 | render_loop
 else get_component_configs(), 
//mesop/mesop/editor/component_configs.py:13 | get_component_configs
 return [ 
//mesop/mesop/editor/component_configs.py:14 | <listcomp>
 generate_component_config(component) 
//mesop/mesop/editor/component_configs.py:35 | generate_component_config
 fields=get_fields(sig.parameters.items()), 
//mesop/mesop/editor/component_configs.py:92 | get_fields
 field_type = pb.FieldType( 
/python3.10/inspect.py:3254 | signature
 return Signature.from_callable(obj, follow_wrapped=follow_wrapped, 
/python3.10/inspect.py:3002 | from_callable
 return _signature_from_callable(obj, sigcls=cls, 
/python3.10/inspect.py:2396 | _signature_from_callable
 raise TypeError('{!r} is not a callable object'.format(obj)) 
```

I took a look, but not sure the best way to handle update this for the visual editor. But also not sure how to delete the code in a minimal way (other than to delete it all -- lol)

richard-to:
Ok, updated the code to use Javascript filtering (can still do server side filtering as well, but I decided not to highlight that in demo examples). 

Also required the require selection option since it messes up the server side filtering. It seems like a known issue. There are workarounds, but decided to keep it simple.

demo/main.py
```diff
@@ -136,14 +137,15 @@ class Section:
...
+      Example(name="autocomplete"),
       Example(name="button"),
       Example(name="checkbox"),
       Example(name="input"),
-      Example(name="textarea"),
       Example(name="radio"),
       Example(name="select_demo"),
       Example(name="slide_toggle"),
       Example(name="slider"),
+      Example(name="textarea"),
```
wwwillchen, line 148: I know you're trying to alphabetize, but I intentionally put textarea next to input (same as the docs) since they are similar functionality-wise. I think it's OK to not be strictly alphabetical.

mesop/components/autocomplete/autocomplete.ng.html
```diff
@@ -0,0 +1,40 @@
...
+    (keyup)="onKeyUp($event)"
+    [formControl]="autocompleteControl"
+    [matAutocomplete]="auto"
+  />
+  <mat-autocomplete
+    (optionSelected)="onAutocompleteSelectEvent($event)"
+    #auto="matAutocomplete"
+  >
+    @for (option of filteredOptions | async; track $index) { @if
+    (option.getOptionGroup()) {
```
wwwillchen, line 25: this formatting looks a bit funky, not sure if prettier is forcing it?

mesop/components/autocomplete/autocomplete.py
```diff
@@ -0,0 +1,229 @@
...
+
+  This event should mainly be used custom filtering of autocomplate results based on
+  the given user input.
+
+  Attributes:
+    value: Input value.
+    key (str): key of the component that emitted this event.
+  """
+
+  value: str
```
wwwillchen, line 55: I think it's OK to use the InputEvent already defined. https://github.com/google/mesop/blob/c9e38cdf024e7ee175f2fe9fe2521bf938ff76b0/mesop/events/events.py#L22

If we need to add more properties for autocomplete input specifically, we could define AutocompleteInputEvent as a subclass of InputEvent

mesop/components/autocomplete/autocomplete.py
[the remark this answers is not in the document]
wwwillchen: If you're getting an exception here: https://github.com/google/mesop/blob/c9e38cdf024e7ee175f2fe9fe2521bf938ff76b0/mesop/editor/component_configs.py#L91

You can wrap it with a try.. except.. and then set el_fields to empty.

Although I do think deleting it all is not a bad option :D

mesop/components/autocomplete/autocomplete.py
```diff
@@ -0,0 +1,229 @@
...
+  style: Style | None = None,
+  key: str | None = None,
+):
+  """Creates an autocomplete component.
+
+  Args:
+    options: Selectable options from autocomplete.
+    label: Label for input.
+    on_selection_change: Event emitted when the selected value has been changed by the user.
+    on_input: [input](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) is fired whenever the input has changed (e.g. user types). Note: this can cause performance issues. Use `on_blur` instead.
```
wwwillchen, line 145: update

mesop/components/autocomplete/autocomplete.ts
```diff
@@ -0,0 +1,168 @@
...
+  >();
+  private inputSubject = new Subject<Event>();
+  autocompleteControl = new FormControl('');
+  constructor(private readonly channel: Channel) {
+    this.inputSubject
+      .pipe(
+        // Setting this to a short duration to avoid having the user trigger another event
+        // during this debounce time period:
+        // https://github.com/google/mesop/issues/171
+        debounceTime(150),
```
wwwillchen, line 57: no action-needed: I've been thinking about tweaking the input debounce time higher, but I think it's fine to keep these consistent for now.

mesop/components/autocomplete/autocomplete.ts
```diff
@@ -0,0 +1,168 @@
...
+      return options;
+    }
+    const filterValue = value.toLowerCase();
+    const filteredOptions = new Array<AutocompleteOptionSet>();
+    for (const option of options) {
+      if (option.getOptionGroup()) {
+        const filteredOptionGroup = new AutocompleteOptionGroup();
+        filteredOptionGroup.setLabel(option.getOptionGroup()?.getLabel()!);
+        for (const subOption of option.getOptionGroup()?.getOptionsList()!) {
+          if (subOption?.getValue()?.toLowerCase().includes(filterValue)) {
```
wwwillchen, line 149: here and below - I think we should be filtering based on the label and not the value. I think the value is more of an internal representation (e.g. could be a serialized JSON), and the label is what the user sees, so we should filter based on the label.

mesop/components/autocomplete/e2e/autocomplete_test.ts
```diff
@@ -0,0 +1,15 @@
...
+  await expect(page.getByRole('option', {name: 'Arizona'})).toBeAttached();
+  await expect(page.getByRole('option', {name: 'Alaska'})).toBeAttached();
+  await page.locator('//input').fill('Al');
+  await expect(page.getByRole('option', {name: 'Arizona'})).not.toBeAttached();
+  await page.getByRole('option', {name: 'Alaska'}).click();
+  await expect(page.getByText('Selected: Alaska')).toBeAttached();
+  await page.locator('//input').fill('CALI');
+  await page.getByRole('option', {name: 'California'}).click();
+  await expect(page.getByText('Selected: California')).toBeAttached();
+});
```
wwwillchen, line 15: Also, add a test case for enter?

wwwillchen (commented):
overall - looks good, I think the main thing here is whether the filtering should be done on label vs. value.

mesop/components/autocomplete/autocomplete.ts
richard-to: Doh. Good catch. Updated.

mesop/components/autocomplete/autocomplete.py
richard-to: Ah I see. Didn't realize I could reuse the events. Updated

mesop/components/autocomplete/autocomplete.ng.html
richard-to: Hmm, yeah looks like the pre-commit is updating it to look like that. I tried reformatting, but after commiting it changed to that format again.

demo/main.py
richard-to: That's fair. Updated.

mesop/components/autocomplete/autocomplete.py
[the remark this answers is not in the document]
richard-to: Heh, ok. I'll do the try/except.

mesop/components/autocomplete/autocomplete.ng.html
wwwillchen: ok, i think prettier doesn't quite understand angular's new-ish syntax @for / @if

mkdocs.yml
```diff
@@ -47,14 +47,15 @@ nav:
...
+          - Autocomplete: components/autocomplete.md
           - Button: components/button.md
           - Checkbox: components/checkbox.md
           - Input: components/input.md
-          - Textarea: components/textarea.md
           - Radio: components/radio.md
           - Select: components/select.md
           - Slide toggle: components/slide_toggle.md
           - Slider: components/slider.md
+          - Textarea: components/textarea.md
```
wwwillchen, line 58: revert

wwwillchen (approved):
LGTM % minor issue

Issue #670, opened 2024-07-25
Make UploadedFile json serializable


tpcgold:
The File cannot be written to state as it's not yet serializable (although it could be)

**Describe the solution you'd like**
def on_click_upload(e: UploadEvent):
    state = me.state(State)
    state.file = e.file

**Describe alternatives you've considered**
Making a custom class and filling the values when the event happens

**Additional context**
A feature like this would make it easier to handle multiple uploaded files and pass them to other google products like vertex api (just pseudo code):
from vertexai.generative_models import Part
Part.from_data(data=state.file.contents, mime_type=state.file.mime_type)

richard-to:
That's a good call out. Currently the class extends from `io.BytesIO` which is why it's not serializable by default. But I think we can update our serializer code to handle this specific use case much like we do with DataFrame serialization.

PR #679, opened 2024-07-27
Make UploadedFile class serializable

richard-to:
Since we're inheriting from io.Bytes, this made it harder to serialize since we couldn't just make it a data class.

So to workaround this we added custom encoder/decoder logic to handle UploadedFile specifically.

We also needed to add a custom operator for DeepDiff. For whatever reason deep diff wasn't diffing it correctly, so instead we just perform a basical equality check here.

Demos/tests have been updated to illustrate that UploadedFile is now serializable.

Ref #670


mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -241,6 +263,29 @@ def give_up_diffing(self, level, diff_instance) -> bool:
...
+  to diff using equality checks.
+  """
+
+  def match(self, level) -> bool:
+    return isinstance(level.t1, UploadedFile) and isinstance(
+      level.t2, UploadedFile
+    )
+
+  def give_up_diffing(self, level, diff_instance) -> bool:
+    if level.t1 != level.t2:
```
wwwillchen, line 282: for BytesIO is __eq__ equivalent to using `is`? If so, it may be clearer to use `is` to show that this is comparing whether the objects are the same identity.

mesop/dataclass_utils/diff_state_test.py
```diff
@@ -395,6 +396,32 @@ class C:
   ]
 
 
+def test_diff_uploaded_file():
```
wwwillchen, line 399: add a test case when the uploaded file is the same? See earlier comment, it's not clear whether the two uploaded file with the same value (but different identity) will produce a diff or not.

wwwillchen (approved):
One more thing that would be good to test in the current E2E test mesop/components/uploader/e2e/uploader_test.ts is that the actual bytes are correct (and not just the metadata), just to make sure state diffing is working as intended. For example, could we inspect the data URL of the image and make sure it's correct?

mesop/dataclass_utils/dataclass_utils.py
richard-to: I didn't realize there existed an "is" comparator. It seems to compare that it is the same object. But I think when we're doing the diff, there are cases where the object would not be the same exact object since we do a deep copy for tracking the old state.

But maybe I'm not understanding what you're getting at.

mesop/dataclass_utils/diff_state_test.py
richard-to: Sure added a test to check same file data case (different identity)

PR #680, opened 2024-07-27
Update demo screenshots

richard-to:
Noticed that we were missing some screenshots on the demo site.


wwwillchen:
I've noticed this too, that we're not really remembering to re-gen screenshots. Also, I think the screenshots will be slightly different depending on the platform (Mac vs. Windows vs. Linux) so it'll create a lot of diffs when different contributors create them.

 I'm thinking whether we should just generate the screenshots right before pushing a new demo. The only trick is that it might be a bit more difficult to run this on our work machines. WDYT?

richard-to:
> I've noticed this too, that we're not really remembering to re-gen screenshots. Also, I think the screenshots will be slightly different depending on the platform (Mac vs. Windows vs. Linux) so it'll create a lot of diffs when different contributors create them.
> 
> I'm thinking whether we should just generate the screenshots right before pushing a new demo. The only trick is that it might be a bit more difficult to run this on our work machines. WDYT?

Yeah it does seem like the steps to generate the screenshots are easy to automate via Github actions or something. Yes, my work machine is unfortunately pretty borked for running Mesop (need to fix that at some point...).

I do wonder if we could host the demo site on Hugging Face (would just need something to keep the Space from sleeping though)

wwwillchen:
> > I've noticed this too, that we're not really remembering to re-gen screenshots. Also, I think the screenshots will be slightly different depending on the platform (Mac vs. Windows vs. Linux) so it'll create a lot of diffs when different contributors create them.
> > I'm thinking whether we should just generate the screenshots right before pushing a new demo. The only trick is that it might be a bit more difficult to run this on our work machines. WDYT?
> 
> Yeah it does seem like the steps to generate the screenshots are easy to automate via Github actions or something. Yes, my work machine is unfortunately pretty borked for running Mesop (need to fix that at some point...).
> 
> I do wonder if we could host the demo site on Hugging Face (would just need something to keep the Space from sleeping though)

yeah, I think hosting on hugging face isn't a bad idea - let me give it a try. I think we have enough traffic that it wouldn't sleep.

wwwillchen:
Just opened #684 - I think we can close this.

PR #684, opened 2024-07-29
Remove screenshots from repo (run before deployment)

wwwillchen:
See discussion on #680 - it's too easy to forget to remember re-run the generate command and there's also small platform diffs that are annoying to deal with.


richard-to (approved):
Maybe we can move the generate sections screenshot up before the deploy command.

wwwillchen:
> Maybe we can move the generate sections screenshot up before the deploy command.

done

PR #685, opened 2024-07-29
Make demo app deploy-able to Hugging Face spaces

wwwillchen:


demo/README.md
```diff
@@ -30,6 +48,18 @@ gcloud run deploy mesop --source .
 
 See our Mesop deployment [docs](https://google.github.io/mesop/guides/deployment/#deploy-to-google-cloud-run) for more background.
 
+### Deploy to Hugging Face Spaces
+
+Because Hugging Face Spaces has restrictions on not having binary files (e.g. image files), we cannot push the full Mesop Git repo to Hugging Face Spaces. Instead, we copy just the `demo` directory and turn it into a standalone Git repo which we deploy.
```
richard-to, line 53: If image files aren't allowed, how are you planning to generate the screenshots? Would that be done as part of the  docker image build?

demo/README.md
```diff
@@ -1,3 +1,14 @@
+---
+title: Mesop Demo Gallery
+emoji: 🐠
```
richard-to, line 3: 👓 or maybe 🥸

richard-to (approved):

demo/README.md
wwwillchen: I just tried and it's OK to push small image files into HF repo: https://huggingface.co/spaces/wwwillchen/mesop/tree/main/screenshots - not exactly sure what's the cutoff, but the images it was complaining about were much larger.

PR #704, opened 2024-07-31
Update a few patterns for dark mode

richard-to:
Updates dialog, headers, and snackbar to be compliant with dark mode. I was updating these in prompt tuner app, so figured I'd update them here too.

Overall, using theme colors wasn't too bad.


wwwillchen (approved):
cool - good to hear it wasn't bad to use!

PR #705, opened 2024-08-01
Support dark theme: syntax highlighting + demo gallery

### LINKED ISSUE #699: Make demo gallery dark theme

wwwillchen:
Also: minor UI tweaks to demo gallery - remove GitHub logo which looks a little weird; and format the repo link more nicely.

Fixes #699.

### DIFF
--- a/demo/main.py
+++ b/demo/main.py
@@ -635,7 +635,7 @@ def nav_section(section: Section):
       example_name = format_example_name(example.name)
       path = f"/embed/{example.name}"
       with me.box(
-        style=me.Style(color="#0B57D0", cursor="pointer"),
+        style=me.Style(color=me.theme_var("primary"), cursor="pointer"),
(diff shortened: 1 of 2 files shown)


richard-to (approved):
I think it looks good. Only thing is the links color contrast is poor in dark mode. Sounds like a good opportunity to use the link component.

Although this may also be an issue for markdown rendering in dark mode in general.

wwwillchen:
> I think it looks good. Only thing is the links color contrast is poor in dark mode. Sounds like a good opportunity to use the link component.
> 
> Although this may also be an issue for markdown rendering in dark mode in general.

I'm actually using a regular text wrapped by a box with an on_click handler. Right now, you can't handle an in-app navigation with link, which is something we should probably support (opened #707).

I've updated the color for better contrast.

Issue #609, opened 2024-07-12
Focus input field of mesop.labs.Chat after hitting enter


FlorianWilk:
**Is your feature request related to a problem? Please describe.**
When using the mesop.labs.Chat component in a frontend, i always have to re-focus the input text field, after submitting a message. 

**Describe the solution you'd like**
Focussing the InputTextField after submitting a message in mesop.labs.Chat

rsenjins:
In addition to that, it would be great if it is possible to add autofocus to a component on loading of a page. So if loading a page it automatically focusses on the input component for instance.

richard-to:
I think adding focus should be doable by creating a "command" similar to what we do with scroll_into_view command: https://github.com/google/mesop/pull/315

I think it should also be possible to use this approach to focus on loading use the load hook (I'd have to verify that commands can be executed on the load event, but seems likely).

PR #718, opened 2024-08-02
Add focus command.

richard-to:
Add focus command and adds focus to mel.chat

Ref: https://github.com/google/mesop/issues/609

### DIFF
--- a/mesop/labs/chat.py
+++ b/mesop/labs/chat.py
@@ -166,7 +166,7 @@ def chat(
     state = me.state(State)
     state.input = e.value
     yield from submit()
-    me.focus_component(key=f"{len(state.output)}")
+    me.focus_component(key=f"input-{len(state.output)}")
     yield
 
   def submit():
@@ -201,8 +201,7 @@ def chat(
         start_time = time.time()
         yield
     state.in_progress = False
-    yield
-    me.focus_component(key=f"{len(state.output)}")
+    me.focus_component(key=f"input-{len(state.output)}")
     yield
 
   def toggle_theme(e: me.ClickEvent):
@@ -241,7 +240,7 @@ def chat(
           me.input(
             label=_LABEL_INPUT,
             # Workaround: update key to clear input.
-            key=f"{len(state.output)}",
+            key=f"input-{len(state.output)}",
             on_blur=on_blur,
             on_enter=on_input_enter,
             style=_STYLE_CHAT_INPUT,
--- a/mesop/labs/chat.py
+++ b/mesop/labs/chat.py
@@ -166,6 +166,8 @@ def chat(
     state = me.state(State)
     state.input = e.value
     yield from submit()
+    me.focus_component(key=f"{len(state.output)}")
+    yield
 
   def submit():
     state = me.state(State)
@@ -199,6 +201,8 @@ def chat(
         start_time = time.time()
         yield
     state.in_progress = False
+    yield
+    me.focus_component(key=f"{len(state.output)}")
     yield
 
   def toggle_theme(e: me.ClickEvent):
--- a/mesop/runtime/context.py
+++ b/mesop/runtime/context.py
@@ -52,6 +52,11 @@ class Context:
       pb.Command(scroll_into_view=pb.ScrollIntoViewCommand(key=key))
     )
 
+  def focus_component(self, key: str) -> None:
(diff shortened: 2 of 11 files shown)


mesop/tests/e2e/focus_component_test.ts
```diff
@@ -0,0 +1,92 @@
+import {test, expect} from '@playwright/test';
```
richard-to, line 1: These tests are somewhat flaky during local testing.

mesop/tests/e2e/focus_component_test.ts
```diff
@@ -0,0 +1,92 @@
...
+const isFocused = (selector: string) => {
+  const element = document.querySelector(selector);
+  return document.activeElement === element;
+};
+
+test('focus autocomplete', async ({page}) => {
+  await page.goto('/focus_component');
+  await page.locator('//mat-select[@id="mat-select-0"]').click();
+  await page.getByRole('option', {name: 'Autocomplete'}).click();
+  await expect(
```
wwwillchen, line 12: Assuming this is the part that's flaky, here's a suggestion from our friend Claude on how to deflake:

```ts
await expect(async () => {
    const isFocused = await page.evaluate(() => {
      const element = document.querySelector('input[aria-controls="mat-autocomplete-0"]');
      return document.activeElement === element;
    });
    expect(isFocused).toBe(true);
  }).toPass({ timeout: 5000 });
```

If this still doesn't work, you can do a retry for this test specifically, e.g. https://github.com/google/mesop/blob/8d0108f84902d81cda0ebe63757ba9076ae0ca38/mesop/tests/e2e/web_components/slot_test.ts#L5

mesop/runtime/context.py
```diff
@@ -52,9 +52,14 @@ def scroll_into_view(self, key: str) -> None:
...
     )
 
+  def focus_component(self, key: str) -> None:
+    self._commands.append(
+      pb.Command(focus_component=pb.FocusComponentCommand(key=key))
+
   def set_theme_density(self, density: int) -> None:
     self._commands.append(
       pb.Command(set_theme_density=pb.SetThemeDensity(density=density))
+
```
wwwillchen: remove stary line?

mesop/labs/chat.py
```diff
@@ -200,6 +202,8 @@ def submit():
         yield
     state.in_progress = False
     yield
+    me.focus_component(key=f"{len(state.output)}")
```
wwwillchen, line 205: I would add a prefix to this key like "input-" just to make it clearer for people who read this code what it's doing.

wwwillchen, line 205: I'm wondering why this needs to be in a separate yield block, i.e. can this be moved to above L204?

wwwillchen: I'm not sure if you're doing this because the focus command executes too early (e.g. before the component-to-be-focused had been rendered/updated). If that's the case, maybe in shell.ts, you can do some kind of trick like `setTimeout(() => { // focus...}, 0)`

(p.s. maybe scroll_into_view has the same issue)

wwwillchen (commented):
Neat!

Seeing this implementation (quite subtle!) makes me think it's a good thing we've implemented in the core framework :)

mesop/labs/chat.py
richard-to: Ah good catch. Looks like that after yield was not needed.

mesop/runtime/context.py
richard-to: I'm not following the comment? I guess the comment says "outdated" so maybe I fixed it in a follow up commit after I messed up the merge conflict maybe.

mesop/tests/e2e/focus_component_test.ts
richard-to: Thanks. I'll give that a try.

richard-to:
> Seeing this implementation (quite subtle!) makes me think it's a good thing we've implemented in the core framework :)

Hah, yeah. Definitely turned out a bit harder than expected. And I definitely had to elicit the help of Claude to get the full list of selectors there.

mesop/runtime/context.py
wwwillchen: ah yeah.

wwwillchen (approved):

Issue #275, opened 2024-05-17
Provide a copy code button for code snippets in markdown


wwwillchen:
Similar to #224, but it would be nice to automatically have a clipboard button for `me.markdown` (which will automatically benefit `me.code` which uses markdown under the hood)

richard-to:
Question. Do you want the copy button on the markdown block? Or do you only want it to appear for each code fenced block?

Initially was thinking just the markdown block since that's easiest, but reading the title, it occurred to me you could mean the latter.

wwwillchen:
I think it's better to do it for each code-fenced block, since that's a big reason why people want to copy and paste. similar to what Gemini / GitHub does:

<img width="627" alt="image" src="https://github.com/user-attachments/assets/580a14e3-bd21-44fc-a02d-b542c7a74e46">


```py
def foo():
```

text

```py
def foo():
```

richard-to:
Ok, yeah, that's what I was afraid of. I think it's doable. 

Seems like we'll need to get the native html element, then grab all the pre tags and update the mark up to include the copy button. Then I guess under the hood we could try to extract all the code blocks and send them over to the frontend to use with the copy button.

Another option is to write a custom extension which make it easier to extract the code blocks (since we need to account for the various ways code blocks can be rendered) Then we can added some extra formatting for the copy button in the mark up. Then we'd still need to do some DOM manipulation on the native html to add the click handlers.

Third option, which is my preference is to move the markdown component to render on the frontend using markedjs and highlightjs. Then I think it will be easier to add our hook to add the copy button. This also has the benefit of improving the performance of the markdown component. The only hesitation about it is that I don't know if there's any downstream implications for using markedjs and highlightjs. Also don't know what issues you ran into initially with using markedjs.

simonff:
+1, missing this feature

Issue #224, opened 2024-05-10
Provide a copy to clipboard command


wwwillchen:
This puts text into a clipboard:

`me.copy_to_clipboard(text)`

Potentially want the corresponding API of reading from a clipboard but that requires browser permission

ChaitanyaYeole02:
Hey, can you assign this issue to me?

tpcgold:
There are solutions for this. I currently have a workaround like this implemented: 
in main.py:
 ```
 from copy_to_clipboard import copy_to_clipboard
 
 def on_click_clipboard():
 state = me.state(State)
 copy_to_clipboard(value=state.output)
 ```
in copy_to_clipboard.py:
```
from typing import Any, Callable

import mesop.labs as mel
from mesop.events import ClickEvent


@mel.web_component(path="copy_to_clipboard.js")
def copy_to_clipboard(
    *,
    value: str,
    on_copy: Callable[[str], Any] = None,  # Type hint for on_copy argument
    key: str | None = None,
):
    async def _on_copy(event: ClickEvent):
        return

    return mel.insert_web_component(
        name="clipboard-component",
        key=key,
        events={
            "click": _on_copy,  # Directly bind the click event to the handler
        },
        properties={
            "value": value,
        },
    )

```
in copy_to_clipboard.js
```
import {html, LitElement} from 'https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js';

class ClipboardComponent extends LitElement {
    static properties = {
        value: {type: String},
        copyEvent: {type: String},
    };

    constructor() {
        super();
        this.value = "";
        this.copyEvent = "";
    }

    connectedCallback() {
        super.connectedCallback();
        this.addEventListener("click", this._onCopy);
    }

    disconnectedCallback() {
        super.disconnectedCallback();
        this.removeEventListener("click", this._onCopy);
    }

    render() {
        return html`
            <div class="container">
                <button @click="${this._onCopy}">
                    Copy to Clipboard
                </button>
            </div>
        `;
    }

    async _onCopy() {
        const textToCopy = this.value;
        try {
            await navigator.clipboard.writeText(textToCopy);
            console.log("Text copied to clipboard!");
            this.dispatchEvent(
                new MesopEvent(this.copyEvent, {
                    value: textToCopy,
                })
            );
        } catch (err) {
            console.error("Failed to copy text:", err);
        }
    }
}

customElements.define('clipboard-component', ClipboardComponent);
```

[...]

PR #720, opened 2024-08-02
Add copy to clipboard web component

richard-to:
One issue trying to implement a copy command is that it does not work in Safari. Safari requires a user action, such as a click unfortunately.

The other option here is to use web component with a slot and click event, which apparently does work in Safari.

This web component also expects the browser has the  navigator.clipboard function.

Ref: https://github.com/google/mesop/issues/224
Ref: https://github.com/google/mesop/issues/275

### DIFF
--- a/mesop/examples/web_component/BUILD
+++ b/mesop/examples/web_component/BUILD
@@ -11,6 +11,7 @@ py_library(
     deps = [
         "//mesop",
         "//mesop/examples/web_component/code_mirror_editor",
+        "//mesop/examples/web_component/copy_to_clipboard",
         "//mesop/examples/web_component/custom_font_csp_repro",
         "//mesop/examples/web_component/firebase_auth",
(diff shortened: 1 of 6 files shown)


wwwillchen:
For: #275 - WDYT about adding a copy to clipboard button directly in our markdown component (e.g. similar to what GitHub does)? I think it would be very useful and handle most of my (personal) use cases for wanting copy to clipboard.

wwwillchen (approved):

richard-to:
> For: https://github.com/google/mesop/issues/275 - WDYT about adding a copy to clipboard button directly in our markdown component (e.g. similar to what GitHub does)? I think it would be very useful and handle most of my (personal) use cases for wanting copy to clipboard.

Yeah, I agree that adding a builtin copy option to markdown should cover most uses cases. And if people more advanced functionality, they can use a web component.

PR #722, opened 2024-08-03
Change internal routes to prefix and postfix with "__"

wwwillchen:
This avoids potential naming collisions and establishes a convention that routes Mesop sets up are named: `/__{MESOP_ROUTE}__`.


richard-to (approved):
There's also `@flask_app.route("/hot-reload")` Does that one need to stay as is?

wwwillchen:
> There's also `@flask_app.route("/hot-reload")` Does that one need to stay as is?

good call. updated.

PR #723, opened 2024-08-03
Load JS modules for web components as-needed [BREAKING CHANGE]

wwwillchen:
I discovered an issue with the way we're loading JS modules for web components where _all_ the pages of the app will try to load the JS modules. This leads to either: a) unnecessarily loading JS modules or b) getting spammed with CSP errors because the other pages (which don't load the web component) will probably not have the right security policy (CSP) configuration. b) was happening on our own demo site.

This loads JS modules dynamically when a web component is called in a render loop. Note: JS module imports should be de-duped so re-importing the same module returns the same instance and is basically instant.

**Breaking change:** This PR introduces a breaking change for Mesop app developers using web components. You may  need to add sites to:  `Security_Policy(allowed_script_srcs=[..]`

Because web components are still in labs, I think it's OK to make this breaking change. I will also have a follow-up PR (#724) to make it easier to fix CSP errors.

**Background:** Based on https://github.com/w3c/webappsec/issues/544, it sounds like when you load a module via `<script src="foo.js" type="module" nonce="...">`, then the modules statically imported by foo.js are covered by the nonce.

### DIFF
--- a/mesop/examples/web_component/code_mirror_editor/code_mirror_editor_app.py
+++ b/mesop/examples/web_component/code_mirror_editor/code_mirror_editor_app.py
@@ -20,7 +20,11 @@ class State:
       "https://cdnjs.cloudflare.com",
       "*.fonts.gstatic.com",
     ],
-    allowed_script_srcs=["https://cdnjs.cloudflare.com", "*.fonts.gstatic.com"],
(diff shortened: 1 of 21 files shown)


richard-to (approved):
Ah great catch.

PR #724, opened 2024-08-03
Provide helpful terminal messages to fix web component CSP errors

### LINKED ISSUE #566: Make it easier to debug & fix CSP (e.g. script-src, Trusted Type) errors
See #549, https://github.com/google/mesop/issues/428#issuecomment-2190417459 - people get tripped up on CSP errors, which are quite frequent when dealing with web components.

We should do the following:

1. Prominently show CSP errors in Python stdout/stderr and in the Mesop UI (when in editor mode) - it's not obvious for people to find the errors in Chrome DevTools and then we can show a more actionable message, linking to our docs with examples of how to fix it.

2. Print a suggested code to fix the CSP issue.

wwwillchen:
Fixes #566. This makes it a lot easier to fix CSP errors by printing a helpful message in the terminal. Particularly helpful with #723. 

**Screenshot:**

image

### DIFF
--- a/mesop/server/static_file_serving.py
+++ b/mesop/server/static_file_serving.py
@@ -7,13 +7,15 @@ import secrets
 from collections import OrderedDict
 from io import BytesIO
 from typing import Any, Callable
+from urllib.parse import urlparse
 
-from flask import Flask, Response, g, request, send_file
+from flask import Flask, Response, g, make_response, request, send_file
 from werkzeug.security import safe_join
 
 from mesop.exceptions import MesopException
 from mesop.runtime import runtime
 from mesop.server.constants import WEB_COMPONENTS_PATH_SEGMENT
+from mesop.utils import terminal_colors as tc
 from mesop.utils.runfiles import get_runfile_location, has_runfiles
 from mesop.utils.url_utils import sanitize_url_for_csp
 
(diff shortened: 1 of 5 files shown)


richard-to (approved):
Looks like `report-uri` is deprecated (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri).

Seems like the newer way is `report-to`, however not all browsers support it yet (Firefox) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to.

--------

One thing that could be helpful is to also include the full url that got blocked since there could be multiple files that got blocked from the same domain. So I think it could be important to verify that all the urls were expected ones.

wwwillchen:
> Looks like `report-uri` is deprecated (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri).
> 
> Seems like the newer way is `report-to`, however not all browsers support it yet (Firefox) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-to.

Yeah, I looked into this a little bit, but the lack of Firefox support made me not go down the route. There's also some trickier aspects where it's unclear whether it'll work [locally](https://github.com/elastic/kibana/issues/149985#issuecomment-1456658576) and there may be a [delay](https://developer.chrome.com/docs/capabilities/web-apis/reporting-api#save_time). For now, I think report-uri is OK (I _think_ lots of Google sites use it exclusively and not report-to) and if it does break eventually, we can switch over to report-to. 

> One thing that could be helpful is to also include the full url that got blocked since there could be multiple files that got blocked from the same domain. So I think it could be important to verify that all the urls were expected ones.

Good idea. I've included the full blocked URL.

PR #732, opened 2024-08-04
Rename docs to use hyphen instead of underscore

wwwillchen:
Follows Google Search guidelines and it's more conventional for URLs, generally, to have hyphens instead of underscores: https://developers.google.com/search/docs/crawling-indexing/url-structure#:~:text=We%20recommend%20that%20you%20use,similar%20content%20on%20your%20site.


wwwillchen:
@richard-to FYI

PR #753, opened 2024-08-07
Set web component as properties, not attributes [breaking change]

### LINKED ISSUE #730: Web component properties do not seem to accept objects or arrays
Example:

```
  static properties = {
    value1: {type: Object},
    value2: {type: Array},
  };
```

I'm not sure if I'm doing something wrong here, but when I try to pass in an Array or Object, I get undefined.

wwwillchen:
Fixes #730.

**Breaking change**: Web components are now set via DOM properties and not attributes. This means you can _not_ use Lit's attribute API, e.g. 

```ts
  static properties = {
    value: {type: Number},
    incrementEventHandlerId: {attribute: 'increment-event', type: String},
  };
```

I mixed up [DOM attributes and DOM properties](https://jakearchibald.com/2024/attributes-vs-properties/#value-types:~:text=elements%2C%20not%20properties.-,Value%20types,-In%20order%20to) and only primitive types, e.g. string, number, boolean, worked when I erroneously set the web component properties as DOM attributes. [Lit's default converter](https://lit.dev/docs/components/properties/#conversion-type) worked for the primitive types.

### DIFF
--- a/mesop/web/src/component_renderer/component_renderer.ts
+++ b/mesop/web/src/component_renderer/component_renderer.ts
@@ -212,7 +212,7 @@ export class ComponentRenderer {
         // We should have checked this in Python, but just in case
         // we will check the attribute name right before using it.
         checkAttributeNameIsSafe(key);
-        customElement.setAttribute(key, (properties as any)[key]);
+        (customElement as any)[key] = (properties as any)[key];
       }
     }
 
--- a/mesop/web/src/component_renderer/component_renderer.ts
+++ b/mesop/web/src/component_renderer/component_renderer.ts
@@ -206,7 +206,7 @@ export class ComponentRenderer {
     for (const key of Object.keys(properties)) {
       const value = (properties as any)[key];
       // We should have checked this in Python, but just in case
-      // we will check the attribute name right before using it.
+      // we will check the property name right before using it.
       checkPropertyNameIsSafe(key);
       (customElement as any)[key] = value;
     }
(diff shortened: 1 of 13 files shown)


mesop/web/src/component_renderer/component_renderer.ts
```diff
@@ -210,7 +210,7 @@ export class ComponentRenderer {
         // We should have checked this in Python, but just in case
         // we will check the attribute name right before using it.
         checkAttributeNameIsSafe(key);
```
richard-to, line 212: Do we need to check attribute name still if we're using properties?

richard-to: Also above there is boolean check. Does that apply to the properties change as well?

richard-to (approved):

mesop/web/src/component_renderer/component_renderer.ts
wwwillchen: good questions - I'm not 100% sure if the attribute name check is still needed. When I experimented with it, things like `onclick` which are unsafe as an attribute (e.g. evals a string as a JS function) don't seem to have the same issue as a property, but I'm not sure if this is true across all browsers. Also the src* checks still seem valid for property. Since it's always easier to remove a check and I haven't heard anyone complain yet about these checks, I'll leave in the name check.

The boolean check, though, I think we can remove - I tested it out and I don't see an issue with always setting the boolean property.

PR #755, opened 2024-08-07
Add google.github.io to showcase allowed iframe parents

wwwillchen:


richard-to:
Don't think this is related, but when I click on the links on the hugging face iframe, I'm not able to view any of the links to the repos and stuff.

I also think the add show case button maybe should be made prominent -- maybe add a link one on the side bar as well?

wwwillchen:
> Don't think this is related, but when I click on the links on the hugging face iframe, I'm not able to view any of the links to the repos and stuff.

Yeah, I discovered this too, when you view the showcase in an iframe pretty much all links will break because those sites (e.g. github) prevent being iframed for web security reasons. I don't think there's much we can do about this which is why I link directly to the hf site (and not the embed): https://wwwillchen-mesop-showcase.hf.space/

> 
> I also think the add show case button maybe should be made prominent -- maybe add a link one on the side bar as well?

good idea - done - https://github.com/google/mesop/pull/757

wwwillchen:
Closing this since embedding doesn't work well for showcase (see above comment).

richard-to:
Would it be possible to open the links a new tab instead?

Issue #312, opened 2024-05-27
Edge case with re-rendering stateful components


wwwillchen:
Context: #311 

```py
import mesop as me

@me.page(path="/diff_edge_case")
def app():
    me.text("diff_edge_case")
    me.radio(options=[me.RadioOption(label="1", value="1",),
                      me.RadioOption(label="2", value="2",)], on_change=on_change)
    if me.state(State).value == "1":
        me.checkbox("check me")
    else: 
        me.checkbox("i will be erroneously checked")


@me.stateclass
class State:
    value: str ="1"

def on_change(e: me.RadioChangeEvent):
    me.state(State).value = e.value
```

If I check the checkbox, and then I toggle the radio option, the checkbox (the second one) should be unchecked, however because I believe the diff algorithm thinks the component instance is the same, it'll treat the second checkbox as the same instance as the first checkbox and the checkbox will be erroneously checked.

You can manually fix this error by setting different key values for each checkbox and then it behaves as expected.

I think one possible way of fixing this is by creating an implicit key which is the hash of the traceback, this way the first and second checkbox will have different keys. I'll try sending a PR for this and we can see whether it makes sense (the concern with hashing a traceback is that it could add a non-trivial runtime overhead for each component instance).

PR #762, opened 2024-08-08
Proper child reconciliation for web components with slots.

### LINKED ISSUE #449: Better child component reconciliation for web components.
Right now, we do a naive approach were all the children elements are removed and then new children elements are attached. The problem is that this blows away the DOM state and is particularly annoying for components like checkbox that idiomatically rely on the DOM state for whether it's checked. You can workaround this by explicitly setting the `value`, but this isn't ideal.

We can also improve the performance by having a more sophisticated way of reconciling child components.

wwwillchen:
Fixes #449.

Drive-by: deflake the existing slot test by adding in `await`.

### DIFF
--- a/mesop/web/src/component_renderer/component_renderer.ts
+++ b/mesop/web/src/component_renderer/component_renderer.ts
@@ -159,28 +159,11 @@ export class ComponentRenderer {
 
   ngOnChanges() {
     if (this.customElement) {
-      // This is a naive way to apply changes by removing all the children
-      // and creating new children. In the future, this can be optimized
-      // to be more performant, but this naive approach should have the
-      // correct behavior, albeit inefficiently.
-      // See: https://github.com/google/mesop/issues/449
-
-      // Clear existing children
-      for (const element of Array.from(
-        this.customElement.querySelectorAll(COMPONENT_RENDERER_ELEMENT_NAME),
-      )) {
-        this.customElement.removeChild(element);
-      }
-
-      // Update the custom element and its children
+      // Update the custom element properties and events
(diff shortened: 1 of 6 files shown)


mesop/web/src/component_renderer/component_renderer.ts
```diff
@@ -233,6 +216,34 @@ export class ComponentRenderer {
...
+      this.customElement!.querySelectorAll(COMPONENT_RENDERER_ELEMENT_NAME),
+    );
+    const newChildren = this.component.getChildrenList();
+
+    // Update or add children
+    for (let i = 0; i < newChildren.length; i++) {
+      const child = newChildren[i];
+      if (i < existingChildren.length) {
+        // Update existing child
+        (existingChildren[i] as any)['component'] = child;
```
richard-to, line 230: Are there any edge cases where this could have unexpected behavior?

For example, if the order of the components got swapped, will Angular still know the text input is the same text input and thus will not re-render it?

Let's say we start with:

text1
text2
button1

And after the state update we get:

text2
button1
text1

And now let's say we get this update:

button3
button2
text1
button1
text2

I think the components will render, but not sure if any of the text inputs would get reset in this scenario.

richard-to (approved):
Awesome. I'm looking forward to this improvement.

mesop/web/src/component_renderer/component_renderer.ts
wwwillchen: Discussed offline: this isn't unique to web components and is a longstanding issue: #312

PR #789, opened 2024-08-13
Support functools.partial for event handlers

### LINKED ISSUE #766: Support functools.partial instances as event handlers
Currently you cannot pass an instance of functools.partial as an event handler since several places in helper.py assume the handler has a `__call__` and `__module__` property and can be passed to `inspect.getsource` (in `compute_fn_id` and `wrap_hander_with_event`). 

But you *can* pass a lambda, which is conceptually similar.

If this can't be supported for whatever reason, then maybe the event handler type hints should be changed since event handlers are annotated as `Callable` and a `partial` is a `Callable`.

wwwillchen:
Fixes #766

### DIFF
--- a/mesop/component_helpers/helper.py
+++ b/mesop/component_helpers/helper.py
@@ -1,7 +1,9 @@
 import hashlib
 import inspect
 import json
-from functools import lru_cache, partial, wraps
+from dataclasses import is_dataclass
+from enum import Enum
+from functools import partial, wraps
 from typing import (
   Any,
   Callable,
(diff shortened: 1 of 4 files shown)


malmaud:
I'm not a super-expert on the best way to do this but this seems reasonable to me.

wwwillchen:
@richard-to WDYT about this feature / implementation? I think the feature is useful, but the implementation is a little bit hairy. In particular, we need to avoid getting a repr for types that return a memory address (e.g. vanilla Python classes)

```py
class Car:
  def __init__(self, name: str) -> None:
    self.name = name

car = Car(name="a") # repr -> <__main__.Car object at 0x79cfdc146080>
```

mesop/component_helpers/helper.py
```diff
@@ -374,14 +380,63 @@ def register_event_handler(
   return fn_id
 
 
-@lru_cache(maxsize=None)
+def has_stable_repr(obj: Any) -> bool:
+  """Check if an object has a stable repr."""
```
richard-to, line 384: [nit] Maybe add a bit more detail in this comment. Specifically would be good to know the potential limitations of using partial functions.

richard-to (approved):
Nice! I remember when I first started using Mesop, I had tried using functools.partial to no avail.

Looking at the code example, it does make the devex a lot nicer than having to pass in a key and then parsing the string.

Question: does this work when doing a dynamic loop? Say I wanted to add an event handler to each box using partial?

Yes, the limitation of needing a stable repr is interesting. How much of a drawback do you think that will be in practice?  I think it may be slightly annoying, but the improvement in devex is worth it I think (granted I haven't tested this out with some of the uses I have for this yet)

wwwillchen:
> Nice! I remember when I first started using Mesop, I had tried using functools.partial to no avail.
>
> Looking at the code example, it does make the devex a lot nicer than having to pass in a key and then parsing the string.
> 
> Question: does this work when doing a dynamic loop? Say I wanted to add an event handler to each box using partial?

Yup. I added an example in the functools_partial example.

> Yes, the limitation of needing a stable repr is interesting. How much of a drawback do you think that will be in practice? I think it may be slightly annoying, but the improvement in devex is worth it I think (granted I haven't tested this out with some of the uses I have for this yet)

I think it's not too bad since this should cover most of the cases. It may be a little hard to understand for Mesop app developers, but I think we can work on improving the error message + docs if this becomes a common issue.

Thanks for the feedback, I'll merge this in.

PR #807, opened 2024-08-16
Support non-generating coroutines as event handlers.

malmaud:
Allows you to elide a solitary final 'yield' from async event handlers that don't stream responses, which seems more ergonomic.

### DIFF
--- a/mesop/examples/async_await.py
+++ b/mesop/examples/async_await.py
@@ -30,4 +30,3 @@ async def click_async(e: me.ClickEvent):
   me.state(State).val1, me.state(State).val2 = await asyncio.gather(
     val1_task, val2_task
   )
-  yield
--- a/mesop/examples/async_await.py
+++ b/mesop/examples/async_await.py
@@ -8,7 +8,8 @@ def page():
   s = me.state(State)
   me.text("val1=" + s.val1)
   me.text("val2=" + s.val2)
-  me.button("async", on_click=click_async)
+  me.button("async with yield", on_click=click_async_with_yield)
+  me.button("async without yield", on_click=click_async_no_yield)
 
 
 @me.stateclass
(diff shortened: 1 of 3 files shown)


mesop/examples/async_await.py
```diff
@@ -30,4 +30,3 @@ async def click_async(e: me.ClickEvent):
   me.state(State).val1, me.state(State).val2 = await asyncio.gather(
```
wwwillchen, line 30: Please create a new function (and button) without doing the yield and keep the existing function and then update the e2e test to cover this case https://github.com/google/mesop/blob/fee3719827a1cd1ebfb419a639c1426279d69671/mesop/tests/e2e/async_await_test.ts#L4

For testing information, see: 
- https://google.github.io/mesop/internal/testing/#e2e-tests
- https://google.github.io/mesop/guides/testing/#running-and-debugging-a-test

wwwillchen (changes requested):
thx for sending the PR - overall change looks good, I think adding a test case (see comment) and then we can merge it.

mesop/examples/async_await.py
malmaud: done

malmaud:
Not sure if I can retrigger the CI on my side, but the tests pass locally for me now.

wwwillchen:
> Not sure if I can retrigger the CI on my side, but the tests pass locally for me now. 

I ran the ci and there's a test failing. Take a look at the error message

malmaud:
It looks like that CI run is from before I fixed the error on my branch via a force-push, and then didn't automatically re-run.

mesop/tests/e2e/async_await_test.ts
```diff
@@ -12,3 +12,16 @@ test('async await', async ({page}) => {
     'val2=<async_value>',
```
wwwillchen, line 12: need to update the existing test, see error message:


```
  1) [chromium] › mesop/tests/e2e/async_await_test.ts:3:5 › async await ────────────────────────────

    Error: locator.click: Error: strict mode violation: getByRole('button', { name: 'async' }) resolved to 2 elements:
        1) <button mat-button="" mat-ripple-loader-uninitialized=""…>…</button> aka getByRole('button', { name: 'async', exact: true })
        2) <button mat-button="" mat-ripple-loader-uninitialized=""…>…</button> aka getByRole('button', { name: 'async without yield' })

    Call log:
      - waiting for getByRole('button', { name: 'async' })


      3 | test('async await', async ({page}) => {
      4 |   await page.goto('/async_await');
    > 5 |   await page.getByRole('button', {name: 'async'}).click();
        |                                                   ^
      6 |   await expect(page.getByText('val1=<async_value>')).toBeVisible({
      7 |     // Intentionally set timeout at 3 seconds to make sure the
      8 |     // async work is happening in parallel.

        at /home/runner/work/mesop/mesop/mesop/tests/e2e/async_await_test.ts:5:51
```

wwwillchen:
thanks!

PR #808, opened 2024-08-16
Tighten event type in `register_event_handler`

malmaud:
pylance was complaining about the type of `handler_fn` not being compatible with `wrap_handler_with_event`.


wwwillchen:
thx

PR #817, opened 2024-08-21
Allow sets in State classes.

### LINKED ISSUE #770: Support `set` type for state serialization?

richard-to:
This change adds support for using sets in State classes.

- Closes #770 
- Ref #387


wwwillchen (approved):
overall change looks good, but i'm wondering if there's a way we could unify the code between set and iterable/array more. It seems like they are pretty similar but they have separate code paths now.

richard-to:
Sure. Yeah. I agree that those functions can be refactored. Probably can extract out a function like getLastSegment or something. I'll do that as a follow up.

PR #821, opened 2024-08-22
Create experimental visual editor toolbar & services

wwwillchen:
# Feature overview

image

- You can drag it around and expand/minimize which is persisted into localStorage.
- You can view history and revert to a previous revision
- You can save a change from the view history which it adds to `goldens/`

# Eng Design

- The visual editor toolbar is guarded behind a flag and lazily loaded with defer views because it's still in an experimental stage and we don't want to load it for other Mesop app developers.
- This also separates the LLM call into a separate service because we want to decouple it from the main Mesop package.

# Usage

## Server-side

```sh
cd ai/services/
pip install -r requirements.txt 
python service.py 
```

## Client-side 

**Flag control** (run in chrome devtools console):

Enable:
```js
window.localStorage.setItem('MESOP://SHOW_EDITOR_TOOLBAR', 'true');
```

Disable:
```js
window.localStorage.removeItem('MESOP://SHOW_EDITOR_TOOLBAR');
```


mesop/server/server.py
```diff
@@ -261,6 +265,128 @@ def teardown_clear_stale_state_sessions(error=None):
...
+    @flask_app.route("/__editor__/save-interaction", methods=["POST"])
+    def save_interaction() -> Response:
+      check_editor_access()
+
+      data = request.get_json()
+      if not data:
+        return Response("Invalid JSON data", status=400)
+
+      try:
+        req = urllib_request.Request(
```
richard-to, line 303: [optional] I know you hate using external libraries if not necessary (and in this case probably not necessary since usage is pretty basic), but `requests` is is a slightly nicer api and I believe it's probably already included from some dependency of a dependency.

mesop/server/server.py
```diff
@@ -261,6 +265,128 @@ def teardown_clear_stale_state_sessions(error=None):
...
+          "http://localhost:43234/save-interaction",
+          data=json.dumps(data).encode("utf-8"),
+          headers={"Content-Type": "application/json"},
+        )
+        with urllib_request.urlopen(req) as response:
+          if response.status == 200:
+            folder = json.loads(response.read().decode("utf-8"))["folder"]
+            response_data = {"folder": folder}
+            return Response(
```
richard-to, line 312: [nit] You can return a dict and it will automatically return json mime type

mesop/server/server.py
```diff
@@ -261,6 +265,128 @@ def teardown_clear_stale_state_sessions(error=None):
...
+              json.dumps(response_data), status=200, mimetype="application/json"
+            )
+          else:
+            print(f"Error from AI service: {response.read().decode('utf-8')}")
+            return Response(
```
richard-to, line 317: [nit] Could be wrong, but think you can return `error_message, status` But I think what's here is a bit more clear.

richard-to (approved):
LGTM

mesop/server/server.py
wwwillchen: you know me well :) 

yeah agree that `requests` is nicer, but we don't have any transitive dependencies in our pip build of Mesop (we have requests as a transitive dependency with some of our dev deps).

mesop/server/server.py
wwwillchen: ok neat

PR #823, opened 2024-08-23
Add script for add code/api snippets into docs.

richard-to:
This script will go through the markdown documentation and inserts code examples and api docstrings into the docs.

This code needs to import mesop, so your terminal needs to have mesop imported via pip.
It does not seem to work when .cli.venv is loaded.

```
python transform_markdown.py
python transform_markdown.py --cache_files_list
python transform_markdown.py --use_cached_files_list
```

You can use the cached_files list to create a block list to ignore specific files that are not relevant.

This script is not integrated with the existing workflow yet.

-----

It fails to find:

- mesop.components.select.select.SelectOpenedChangeEvent
- mesop.components.progress_bar.progress_bar.ProgressBarAnimationEndEvent

These two classes do not seem to be imported in the __init__.py file in Mesop, so I can look the inspect the classes. Not sure if that's intentional or not.


wwwillchen:
> This script will go through the markdown documentation and inserts code examples and api docstrings into the docs.
> 
> This code needs to import mesop, so your terminal needs to have mesop imported via pip. It does not seem to work when .cli.venv is loaded.
> 
> ```
> python transform_markdown.py
> python transform_markdown.py --cache_files_list
> python transform_markdown.py --use_cached_files_list
> ```
> 
> You can use the cached_files list to create a block list to ignore specific files that are not relevant.
> 
> This script is not integrated with the existing workflow yet.
> 
> It fails to find:
> 
> * mesop.components.select.select.SelectOpenedChangeEvent
> * mesop.components.progress_bar.progress_bar.ProgressBarAnimationEndEvent
> 
> These two classes do not seem to be imported in the **init**.py file in Mesop, so I can look the inspect the classes. Not sure if that's intentional or not.

We should export those two classes since they're part of our public API. I think it was just an oversight

ai/prompt_context/transform_markdown.py
```diff
@@ -0,0 +1,126 @@
...
+
+def main():
+  parser = argparse.ArgumentParser(
+    prog="Transform Markdown",
+    description="Transform markdown documents for prompt context.",
+  )
+  parser.add_argument("--use_cached_files_list", action="store_true")
+  parser.add_argument("--cache_files_list", action="store_true")
+  args = parser.parse_args()
+
```
wwwillchen, line 29: Add a check to make sure both flags are not set to True? (it seems like it would be a mistake)

ai/prompt_context/transform_markdown.py
```diff
@@ -0,0 +1,126 @@
...
+        # Handle code snippet imports
+        if line.startswith("--8<--"):
+          with open(
+            "../../" + line.removeprefix("--8<--").strip().replace('"', "")
+          ) as f2:
+            updated_lines.append(f2.read())
+        # Handle doc strings
+        elif line.startswith("::: "):
+          try:
+            function = line.removeprefix(":::").strip().split(".")[-1]
```
wwwillchen, line 88: a bit confusing to call this function since it's more broad than that, maybe call this symbol?

ai/prompt_context/transform_markdown.py
```diff
@@ -0,0 +1,126 @@
...
+            if function == "ThemeVar":
+              function = "theme_var"
```
wwwillchen, line 90: We should make sure we're including all the ThemeVar values in the prompt context as it's important for getting the LLM to generate the right colors. You can use this:

```
from typing import get_args

get_args(me.ThemeVar)
```

You might need to reach in and grab ThemeVar from `mesop.features.theme` since it's not exported right now

ai/prompt_context/transform_markdown.py
```diff
@@ -0,0 +1,126 @@
...
+            if "mesop.labs." in line:
+              updated_lines.append(
```
wwwillchen, line 92: We should also include the function signature as the docstring doesn't contain type information.

```py
import inspect

signature = inspect.signature(example_function)
print(signature)
```

wwwillchen (approved):
Overall I like this approach much better than my HTML approach - it's a lot cleaner and easier to extend. nice work!

ai/prompt_context/transform_markdown.py
richard-to: Good point. Updated.

ai/prompt_context/transform_markdown.py
richard-to: Ah good catch.

richard-to:
Ok. 

- Exported the progressbar end event and the select opened change event to the mesop __init__ object.
- Also added a check for enabling both flags at the same time (I had left it as is since I think it would be a noop in that situation. Would just end up overwriting the existing file with the same cached elements, but I think easy enough to add a check)

PR #842, opened 2024-08-26
Adds app url for mesop app maker.

richard-to:
I added Mesop App Maker to to Hugging Face. It's not ideal, since users need to create their own runner in Hugging Face Spaces. It's prettty easy to do. Just need to duplicate the Mesop App Runner space and set a "runner token". The runner token is just a basic way to block people from randomly using other people's runner instances.

Also added a link to the Hugging Face Jeopardy Mesop app.

### DIFF
--- a/showcase/main.py
+++ b/showcase/main.py
@@ -59,6 +59,7 @@ SECTIONS = [
       Resource(
         title="Mesop App Maker",
         description="Generate apps with Mesop using LLMs",
+        app_url="https://huggingface.co/spaces/richard-to/mesop-app-maker",
         img_url="https://github.com/user-attachments/assets/1a826d44-c87b-4c79-aeaf-29bc8da3b1c0",
         github_url="https://github.com/richard-to/mesop-app-maker",
         github_username="richard-to",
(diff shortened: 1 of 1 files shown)


wwwillchen:
Just opened #849 which I'll try to do soon.

I think this is a good improvement, but it still feels a bit cumbersome.

Just to put an idea out there, what if we executed arbitrary user code in the same HF space and used something like https://github.com/zopefoundation/RestrictedPython to execute the user code with a restricted subset (e.g. you can call me.* and not much else).

Assuming the HF space doesn't have sensitive content (the only thing sensitive might be the API keys), but I could see this being useful even outside of LLM generation (e.g. user can give a URL like TypeScript playground: https://www.typescriptlang.org/play/?#code/G4QwTgBCC8EMwCgQVgJgQoA, to reproduce an issue or simply for sharing an example). WDYT?

wwwillchen (approved):

richard-to:
Yes. Agreed not ideal. But figured it was better than nothing at the moment.

Ah, nice find. Yes, that could be useful, especially if each execution is itself sandboxed.

Definitely agree about the Mesop playground thing. That did cross my mind when I was trying to get viewer to work in a cheap and secure way.

Originally what I had wanted to do was use Docker in Docker. Then we could have an endpoint that could spawn docker containers and execute the code within a single container. But Hugging Face containers are more restricted in usage since Docker in Docker does require a bit more privileges to run.

richard-to:
I looked into restricted python. I was able to use it fine. But it has a big blocker for now (can't handle type annotations yet...) https://github.com/zopefoundation/RestrictedPython/issues/219

wwwillchen:
yeah, the workaround in the comment looks pretty promising

richard-to:
> yeah, the workaround in the comment looks pretty promising

Yeah, I tried it out and it worked. Actually their code was very helpful in getting things set up. Still running into some errors, but I think I'm fairly close to getting it to work.

PR #846, opened 2024-08-26
Add hotkeys and async action web component examples.

richard-to:
Now that the child components issue has been fixed (https://github.com/google/mesop/issues/449), the hotkeys web component that I was working on a few weeks ago is working now.

Also adding the async action web component example.

### DIFF
--- a/mesop/examples/web_component/BUILD
+++ b/mesop/examples/web_component/BUILD
@@ -10,11 +10,13 @@ py_library(
     data = glob(["*.js"]),
     deps = [
         "//mesop",
+        "//mesop/examples/web_component/async_action",
         "//mesop/examples/web_component/code_mirror_editor",
         "//mesop/examples/web_component/complex_props",
         "//mesop/examples/web_component/copy_to_clipboard",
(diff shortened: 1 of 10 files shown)


wwwillchen (approved):
These are some pretty interesting web components. I'm wondering if should: a) support some of these things natively in mesop or b) make it easier to reuse web components across apps (e.g. `mesop add foo-component`) - WDYT?

richard-to:
Those are great questions. I definitely think the hot key thing would be nice to support natively in mesop. I think we already have a Github issue for it. And I think to do "async actions" in a more full featured way, we'd probably need native support. I think the FastAPI middleware integration could potentially help with this since long running API calls could potentially be sent to a FastAPI endpoint. Something like celery integration would be nice (or whatever is better than celery nowadays). Does GCP have anyway to implement LRO APIs? Maybe revisit Websockets.

But I do like starting off with a web components since it's a cheap way to test out the behavior and usage in the short term. I definitely think the `mesop add foo-component` can be useful as well in general to improve sharing of components (granted we don't have much of an ecosystem yet).

So I think both ideas are useful.

PR #853, opened 2024-08-26
Add ~20 fine-tuning examples and switch to new ft model

wwwillchen:
This includes more examples, including ones where it's prompting a specific component


richard-to (approved):
Awesome. Looking forward to testing out this new version, especially in regard to the layout examples you added.

wwwillchen:
> Awesome. Looking forward to testing out this new version, especially in regard to the layout examples you added.

Yes, please do! I find it's working not too bad on smaller apps, but it's kinda falling apart for bigger aparts (e.g. the diff doesn't applying). If you can contribute some fine-tuning examples as well, that'll be very helpful.

richard-to:
> Yes, please do! I find it's working not too bad on smaller apps, but it's kinda falling apart for bigger aparts (e.g. the diff doesn't applying). If you can contribute some fine-tuning examples as well, that'll be very helpful.

Yeah, I'm hoping to add some examples at some point. For app maker, I've been starting to focus on layout and chat UI components specifically, so hopefully I can add some of the examples I'm creating there. 

One thing I'm trying to experiment with is regular HTML/CSS translation to me.box/me.Style (kind of similar to what you were saying about Tailwind to Mesop translation before). Seems like if we can get the model to recognize that pattern, then maybe it can do a better job of layouts without having to provide too many examples beyond a basic set.

PR #863, opened 2024-08-27
Fix dark theme for code mirror

wwwillchen:


wwwillchen:
@richard-to FYI, after you sync, you'll need to run `bazel clean --expunge && rm -rf node_modules` - there's some kind of bug with npm and bazel and you need to clear the cache.

PR #868, opened 2024-08-28
Support `MESOP_CONCURRENT_UPDATES_ENABLED` env var 

wwwillchen:
This also fixes the issue where our playwright artifacts weren't being properly saved/uploaded on GitHub Actions.

Addresses #623.


richard-to (approved):
Nice. Not able to test this out yet, so looking forward to seeing how far this can be pushed.

I think changes look good. I would have liked to see some more test cases, but since it is behind a flag I think it's OK and we can probably gather feedback from users on edge cases.

wwwillchen:
> Nice. Not able to test this out yet, so looking forward to seeing how far this can be pushed.
> 
> I think changes look good. I would have liked to see some more test cases, but since it is behind a flag I think it's OK and we can probably gather feedback from users on edge cases.

+1. I'm sure there's some funky stuff with this, for example, the 1 second is fairly arbitrary which could be too long or too short depending on the use case, but this should address the core use case in #623

PR #879, opened 2024-08-29
Plumb experimental editor toolbar env var to client-side

wwwillchen:


mesop/web/src/editor/editor.ts
```diff
@@ -163,7 +165,7 @@ class Editor {
   }
 
   showEditorToolbar(): boolean {
-    return Boolean(window.localStorage.getItem('MESOP://SHOW_EDITOR_TOOLBAR'));
+    return this.experimentService.experimentalEditorToolbarEnabled;
```
richard-to, line 168: Nice. Does this mean we don't have to enable the visual editor in Chrome dev tools anymore along with the env var?

richard-to (approved):

mesop/web/src/editor/editor.ts
wwwillchen: yup!

PR #892, opened 2024-08-29
Add docs for mounting Mesop app

### LINKED ISSUE #803: Support mounting Mesop with other Python servers
# Motivation

This could satisfy a few advanced use cases:

- Allows client-side code (web components) to directly call long-running APIs which could support: 1) not blocking state updates and 2) cancelling an LLM call (see #623)
- Support serving custom files (e.g. download links #453)
- Allow forwarding requests which can be needed in certain auth integrations (see: #713)

# Considerations

- We'd like to eventually migrate from Flask to an async-friendly framework like Quart or Starlette so we shouldn't directly expose the framework so we have flexibility to do this migration. However, we should try to provide an API that's similar so the learning curve is reasonably small.

# API proposal

## Simple request

```py
@me.route(methods=["GET", "POST"], path="/foo")
def route():
  return output
```

## Async

```py
@me.route(path="/books/{book_id}")
async def route(book_id: int):
  output = await call_api()
  return output
```

## Path parameters

```py
@me.route(methods=["GET", "POST"], path="/books/{book_id}")
def route(book_id: int):
  return output
```

## Query params
```py
@me.route(methods=["GET", "POST"], path="/books")
def route():
  qp = me.query_params()
  return do_work(qp["foo"])
```

output could be:

- dict
- json
- binary file

# Questions

- Should we try to expose Fast API directly? Or, should we implement this ourselves?
- Should we allow path-dependent ordering? See [FastAPI](https://fastapi.tiangolo.com/tutorial/path-params/#order-matters)
- Should we support multiple HTTP verbs in a single HTTP handler?
- Should you be able to access `me.state(State)`? It would be nice to do this, otherwise it creates an island but you could update the state indirectly from the client-side code.
- https://fastapi.tiangolo.com/tutorial/body/#request-body-path-query-parameters:~:text=The%20function%20parameters%20will%20be%20recognized%20as%20follows%3A

wwwillchen:
The FastAPI example seems to be working well:
https://github.com/wwwillchen/mesop-fastapi

However, I'm having trouble with the Flask example: https://github.com/wwwillchen/mesop-flask - when I try to run Mesop in debug mode, it's not loading.

All things considered though, I think encouraging people to use FastAPI is better as it reduces the likelihood of a version dependency conflict with Mesop's internal Flask usage and other people's Flask usage.

Fixes #803.

### DIFF
--- a/docs/guides/server-integration.md
+++ b/docs/guides/server-integration.md
@@ -1,0 +1,21 @@
+# Server integration
+
+Mesop allows you to integrate Mesop with other Python web servers like FastAPI or Flask by mounting the Mesop app which is a [WSGI](https://wsgi.readthedocs.io/en/latest/what.html) app.
+
+This enables you to do things like:
+
+- Serve local files (e.g. images)
+- Provide API endpoints (which can be called by the web component, etc.)
+
+## API
+
+The main API for doing this integration is the `create_wsgi_app` function.
+
+::: mesop.server.wsgi_app.create_wsgi_app
+
+## FastAPI example
+
+For a working example of using Mesop with FastAPI, please take a look at this repo:
+[https://github.com/wwwillchen/mesop-fastapi](https://github.com/wwwillchen/mesop-fastapi)
+
+> Note: you can apply similar steps to use any other web framework that allows you to mount a WSGI app.
--- a/mesop/server/wsgi_app.py
+++ b/mesop/server/wsgi_app.py
@@ -51,7 +51,10 @@ def create_app(
 
 def create_wsgi_app(*, debug_mode: bool = False):
   """
-  Creates a WSGI app that can be used to run Mesop in a WSGI server.
+  Creates a WSGI app that can be used to run Mesop in a WSGI server like gunicorn.
+
+  Args:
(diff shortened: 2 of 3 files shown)


richard-to (approved):
Nice. This is exciting. It could be good to provide examples of how one could interact with Fast API in your example repo.

Also could be good to know what limitations there are at the moment.

wwwillchen:
> Nice. This is exciting. It could be good to provide examples of how one could interact with Fast API in your example repo.
> 

I updated the readme to be clearer about this - there's an HTTP interaction w/ web component, let me know if you're thinking of something else.

> Also could be good to know what limitations there are at the moment.

good idea, added one around hot reloading.

richard-to:
> I updated the readme to be clearer about this - there's an HTTP interaction w/ web component, let me know if you're thinking of something else.

Ah, yeah I totally missed that (didn't end up looking at web_component.js).

I guess I was  wondering if there was a way to access state from Fast API.  One use case I was thinking about was the download use case, where we have some data in state that needs to be included in the download.

It's not a huge deal if we can't access state from from FastAPI. We'd just have to use the web component as a proxy to send certain state info to Fast API.

PR #893, opened 2024-08-29
Run e2e tests in debug/editor mode (in addition to prod mode)

wwwillchen:
Long overdue, but we should run our e2e tests in both prod and debug/editor mode to catch regressions, especially with all the editor toolbar changes.

This skips some of the e2e tests in debug mode, particularly those related to web security, because we have more lax web security policies when running in debug mode. We could run those tests in debug mode and maintain separate goldens (e.g. for CSPs), but this is a bit of work and doesn't seem too pressing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,8 +40,16 @@ jobs:
       - name: Bazel tests
         # NOTE: bazel test //... doesn't work (due to node_modules)
         run: bazel test //mesop/...
-      - name: Run playwright test
+      - name: Run playwright test (prod mode)
         run: yarn playwright test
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -40,14 +40,22 @@ jobs:
...
         with:
           name: playwright-report
           path: playwright-report/
           retention-days: 30
+      - name: Run playwright test (debug/editor mode)
+        run: MESOP_DEBUG_MODE=true yarn playwright test
+      - uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        if: always()
+        with:
+          name: playwright-report
```
richard-to, line 56: Does it matter what name/path that the report gets written to? I'm not familiar with the format.

richard-to (approved):
LGTM

PR #902, opened 2024-08-29
Allow scroll_into_view to work for components rendered in the same tick

### LINKED ISSUE #804: Make `scroll_into_view` more intuitive by executing at the end of a render loop
We could do a setTimeout around the scroll_into_view command so then you don't need to do the yield trick: https://github.com/google/mesop/blob/fee3719827a1cd1ebfb419a639c1426279d69671/mesop/labs/chat.py#L187C30-L187C39

wwwillchen:
Fixes #804


richard-to (approved):
Nice. That's a very interesting solution.

PR #913, opened 2024-08-31
Add is_target to Click event

### LINKED ISSUE #268: Dealing with event handler bubbling
The scenario that brought this up was the use of a Modal where I wanted to be able to close the modal when clicking on the background.

In this scenario, if I click on the modal content to say copy and paste some text, the Modal will still close since the Modal background click event gets triggered.

This seems like expected behavior. I was wondering how we would prevent this from happening? Would we need to add a noop click event at the modal container/content. Would that prevent the event on the background from being triggered, or would both events get propagated? I can't quite remember what the behavior is but I'll test it out when I get the chance.

richard-to:
Usage of is_target mainly works for me.box.

For me.button / me.content_button it will almost always be false since the buttons are composed of other sub elements. There may be cases where it is true.

Updated dialog usages to utilize `is_target`.

Closes #268

### DIFF
--- a/mesop/events/events.py
+++ b/mesop/events/events.py
@@ -13,9 +13,10 @@ class ClickEvent(MesopEvent):
 
   Attributes:
       key (str): key of the component that emitted this event.
+      is_target (bool): Whether the clicked target contains the event handler.
   """
 
-  pass
+  is_target: bool
 
 
 @dataclass(kw_only=True)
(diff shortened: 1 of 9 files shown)


mesop/events/events.py
```diff
@@ -13,9 +13,10 @@ class ClickEvent(MesopEvent):
 
   Attributes:
       key (str): key of the component that emitted this event.
+      is_target (bool): Whether the clicked target contains the event handler.
```
wwwillchen, line 16: Nit: it's more like the clicked target is the *component* which attached the event handler. (Technically an event handler can be attached to multiple components)

wwwillchen (approved):

PR #915, opened 2024-08-31
Update some usages of scroll-to

richard-to:
We can remove the intentional delay on scroll-to now.

There is one case where this does not work as expected though. When the chat message is not streamed, the scroll to with hidden box strategy does not work since the output is too fast. This occurred on LLM Rewriter example. But I updated it to fake the streaming.

Also decreases the size of the hidden box, so it is less jarring when the box disappears.


wwwillchen (approved):
The LLM rewriter case is interesting, but seems pretty unusual (e.g. doing short-running work across multiple yields). We could separate each yield into a separate JS task so that scroll into view could take effect, but this seems pretty rare so we can hold off on that. Thanks for fixing these!

PR #919, opened 2024-09-01
Add some goldens for writing chat UIs

richard-to:
- Add simple chat example
- Add fancy chat example
- Add other examples for enhancing simple chat

# Starting UI 

Screenshot 2024-09-01 at 8 32 46 AM

# Ending UI

Screenshot 2024-09-01 at 8 31 56 AM

---------

In order to make the ending UI, I did have to do a lot of manual adjustments. I think the only one it did correctly with no adjustments was adding the header. 

Hopefully, I added the goldens correctly since I had to manually edit the diffs.

Also none of the buttons work since I couldn't really add event handlers to them without making the diffs larger


wwwillchen (commented):
You can check if the golden diffs are valid by running:
https://github.com/google/mesop/blob/d22087d37042b515fe1f068b37506e1fa5a2962c/ai/README.md#process-goldens

You can also use the viewer once you processed goldens, to see what it looks like.

Thanks for adding these! BTW, what kind of adjustments did you have to make? Was it giving you invalid syntax / type failures or was the issue that the styling / appearance was off?

Feel free to merge these once you've checked it.

Also, one suggestion for the future is to try splitting up the UI into smaller/fine-grained components. I think this would help make the diffs be finer-grained. Also, we should be able to support multiple diffs (e.g. search/replace blocks) in a single generation, which could help address the handler issue), but I don't think we have enough examples of these that the model does this reliably (or often enough).

Issue #712, opened 2024-08-01
Create a fancier chat example


wwwillchen:
- Includes example boxes
- Nicer textarea styling
- Dark theme

adriens:
Yup, that would definitively help a lot and avoid lazy dev questions 😅

adriens:
:clap:

PR #922, opened 2024-09-05
Add hotkey shortcuts for text areas

### LINKED ISSUE #854: Submit text box on Shift-Enter
It would be great to make the input box submit on Shift-Enter.

As said in chat: "You should be able to implement it with web component, but not super straightforward: https://google.github.io/mesop/web-components/"

richard-to:
Also updates fancy_chat and textarea demos to show usage of shortcuts

Closes #854
Ref #712


mesop/components/input/input.py
```diff
@@ -40,6 +40,47 @@ class InputEnterEvent(MesopEvent):
 )
 
 
+@dataclass(kw_only=True, frozen=True)
+class Shortcut:
+  """Represents a keyboard shortcut combination."""
+
+  key: str
```
wwwillchen, line 47: I think it would be good to provide some documentation and (probably) validation on allowable keys. For example, "enter" is a common one. What about letters and numbers? It seems like we should only accept lowercase letters, e.g. "a", because "A" should really be "a" + shift=True

demo/fancy_chat.py
```diff
@@ -550,6 +554,13 @@ def on_click_regenerate(e: me.ClickEvent):
   yield
 
 
+def on_submit_chat_msg(e: me.TextareaShortcutEvent):
+  state = me.state(State)
+  state.input = e.value
+  yield
+  yield from on_click_submit_chat_msg(me.ClickEvent(key=e.key, is_target=False))
```
wwwillchen, line 561: this works, but it's a little weird of a pattern (for humans and LLM :), I think it's cleaner to have a new function "submit_chat_msg" and have both this function and the on_click function yield from that common function. This way you don't need to construct a click event.

wwwillchen (approved):
Awesome! (there's a test failing you'll need to look into)

mesop/components/input/input.py
richard-to: Yeah that's a good point. Maybe we can just normalize input to lowercase. Since the key for enter is "Enter" so would be easier to just normalize that to "enter".

Looks like there's a table for all the potential key values here: https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values

I'll update the documentation to link to that.

PR #928, opened 2024-09-05
Add get_path_to_page_configs

richard-to:
Adding this method for downstream usage integration.

### DIFF
--- a/mesop/runtime/runtime.py
+++ b/mesop/runtime/runtime.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Callable, Generator, Type, TypeVar, cast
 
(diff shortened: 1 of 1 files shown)


mesop/runtime/runtime.py
```diff
@@ -123,6 +124,9 @@ def register_page(self, *, path: str, page_config: PageConfig) -> None:
   def get_page_config(self, *, path: str) -> PageConfig | None:
     return self._path_to_page_config.get(path)
```
wwwillchen, line 125: I'm wondering if we should also add a deepcopy to this method, but that's a bit orthogonal to this PR.

wwwillchen (approved):

mesop/runtime/runtime.py
richard-to: I think the page config object is small enough that the deepcopy probably wouldn't be noticeable performance-wise. And doesn't seem like it's called a lot. I can add it as a follow up.

Issue #926, opened 2024-09-05
Inconsistent styling between me.text and me.markdown


richard-to:
The default text and default markdown fonts differ quite drastically. It would be great to have default sizes, fonts and weights stay consistent between the two. This causes UI to be all over the place when flipping between text and md in the UI.

PR #930, opened 2024-09-07
Update me.markdown to render on frontend + copy code

### LINKED ISSUE #275: Provide a copy code button for code snippets in markdown

richard-to:
- Updates me.markdown to markedjs and highlightjs for markdown and code highlighting.
- Adds copy code button
- Also update markdown rendering to be a bit closer to me.text rendering
  - Although I didn't want to match the headline types since the defaults are very large
- Update markdown rendering to work better with dark themes (mainly links)
- I do have some concerns about downstream compatibility, but will try to make adjustments as needed.
- I also considered adding a copy all markdown output button, but that seems like it could be better decoupled from the actual markdown place to give users freedom to place it where they want. The question is mainly how to do it since using the command approach does not work with Safari. Right now we can use a web component to do this, but maybe nice to be able to provide native support since this is a fairly common ask.

Closes #275

Ref #926

Ref #591

Ref #224

# Screenshots

Screenshot 2024-09-07 at 11 19 54 AM

Screenshot 2024-09-07 at 11 20 48 AM


mesop/components/markdown/e2e/markdown_test.ts
```diff
@@ -15,14 +15,41 @@ test('renders table markup', async ({page}) => {
...
+  await page.locator('.code-block').nth(1).hover();
+  await page.locator('//button').nth(2).click();
+  clipboardContent = await page.evaluate(async () => {
+    return await navigator.clipboard.readText();
+  });
+  expect(clipboardContent).toContain('print("Hello, World 1!")');
+
+  // Can copy text if markdown is updated
+  await page.getByText('Updated markdown').click();
+  await page.waitForTimeout(1000);
```
wwwillchen, line 48: can you check the markdown is re-rendered instead of a timeout?

mesop/components/markdown/markdown.ts
```diff
@@ -17,14 +27,79 @@ export class MarkdownComponent {
...
+    // Use the custom renderer in Marked.js
+    marked.use({renderer});
+
+    const output = await marked(this._config!.getText()!, {gfm: true});
+
+    // Sanitize the generated HTML from MarkedJS first.
+    const html = this.sanitizer.sanitize(SecurityContext.HTML, output) || '';
+    // Now update the placeholder div with our button once we know the output is
+    // sanitized.
+    this.htmlOutput = this.sanitizer.bypassSecurityTrustHtml(
```
wwwillchen, line 78: you might not be able to call this method downstream b/c of security checks and I don't think you need it. You should be able to directly replace the HTML from L72 with the `<button>` and then have this.htmlOutput be a string. When you bind htmlOutput to innerHTML, Angular should implicitly apply the sanitizer and it should leave your button code alone since it doesn't actually contain JS.

mesop/components/markdown/markdown.ts
```diff
@@ -17,14 +27,79 @@ export class MarkdownComponent {
...
+    const buttons = this.el.nativeElement.querySelectorAll('button');
+    // Use Renderer2 to safely bind a click event to the button
+    for (const button of buttons) {
+      this.renderer.listen(button, 'click', (e: Event) => {
+        const target = e.currentTarget as HTMLElement;
+        const dataText = target.getAttribute('data-index');
+        if (dataText !== null) {
+          navigator.clipboard.writeText(this._codeBlocks[parseInt(dataText)]);
+        }
+      });
```
wwwillchen, line 101: can you do event delegation instead and have a listener on the component root element? then, you don't need the settimeout.

mesop/components/markdown/markdown.ts
```diff
@@ -1,4 +1,14 @@
...
+import {
+  ChangeDetectorRef,
+  Component,
+  Input,
+  ElementRef,
+  Renderer2,
+  SecurityContext,
+} from '@angular/core';
+import {DomSanitizer, SafeHtml} from '@angular/platform-browser';
+import {marked} from 'marked';
```
wwwillchen, line 10: i think you need this shim: https://github.com/google/mesop/commit/8a4f375d56c498e95a47bc9ef8282d7b7fa39f69#diff-0bfb67474f7fa139128300039a775d666a921f24b2dee795d8d725cb2299300b to make it work downstream

wwwillchen (commented):
awesome!! Main feedback around the sanitizer usage. I think getting this working downstream will be a little tricky, not sure if you figured it out yet (you can use copybara to point to your branch if you want to test things out)

mesop/components/markdown/markdown.ts
richard-to: Ah thanks. I remember you mentioning this before. But was not able to find it. Updated.

mesop/components/markdown/markdown.ts
richard-to: Yes, event delegation worked. That's much better.

mesop/components/markdown/markdown.ts
richard-to: Unfortunately, Angular will sanitize the button, but I just changed it to an "a" tag and that worked.

mesop/components/markdown/e2e/markdown_test.ts
richard-to: I ended up doing toBePass, though I think checking for the updated title would have worked as well.

mesop/components/markdown/markdown.ts
```diff
@@ -17,14 +26,62 @@ export class MarkdownComponent {
   @Input() key!: Key;
   @Input() style!: Style;
 
+  htmlOutput: SafeHtml = '';
```
wwwillchen, line 29: I think you can remove the SafeHtml type now

richard-to:
Yeah looks like there's an error uploading the artifacts:

```
[test](https://github.com/google/mesop/actions/runs/10763828978/job/29845813732#step:13:8)
No files were found with the provided path: playwright-report-debug-mode/. No artifacts will be uploaded.
[test](https://github.com/google/mesop/actions/runs/10763828978/job/29845813732#step:11:8)
No files were found with the provided path: playwright-report-prod-mode/. No artifacts will be uploaded.
[test](https://github.com/google/mesop/actions/runs/10763828978/job/29845813732#step:17:8)
No files were found with the provided path: playwright-report-with-memory-state-session/. No artifacts will be uploaded.
[test](https://github.com/google/mesop/actions/runs/10763828978/job/29845813732#step:15:8)
No files were found with the provided path: playwright-report-concurrency/. No artifacts will be uploaded.
```

PR #931, opened 2024-09-07
Attempt to fix flaky test

richard-to:


mesop/components/input/e2e/input_test.ts
```diff
@@ -58,6 +58,8 @@ test('test textarea shortcuts', async ({page}) => {
 
   await textbox.pressSequentially('hi');
   await page.keyboard.press('Enter');
+  await page.waitForTimeout(2000);
+  // Extend timeout a bit due to flakiness.
```
wwwillchen, line 62: I think doing something like https://github.com/google/mesop/blob/0b0b9d76aa16a746e7d8dcd8b51bdbd5886f4c37/mesop/tests/e2e/focus_component_test.ts#L12 will be more resilient

wwwillchen (approved):

PR #933, opened 2024-09-08
Various updates to AI Editor stuff

richard-to:
Probably the best way to review this is to look at the individual commits (is there an easy way to chain pull requests?)

Changes:

1. Added some new goldens
2. Also renamed golden directories a bit. Added a timestamp suffix (used for roughly determining which goldens have not been fine tuned
3. Added .env file support
4. Added a way to include un-fine-tuned goldens to the prompt (quick way to sort of test the new goldens without doing a fine tuning run)
5. Some minor refactoring (not really for ideal code structure, but the minimal amount of changes to include the adjustments added here
6. Also add support for using Gemini client instead of OpenAI client (not using a fine tuned version yet, but bundling the goldens into the prompt for now)


richard-to:
> LGTM - btw, I'm prototyping a more structured way to manage evals and goldens, hope to send it out soon.

Nice. I'm looking forward to see that change. Definitely a lot of room for improvement there.

PR #942, opened 2024-09-11
Fix web component static file serving

wwwillchen:


richard-to (approved):
<redacted>

wwwillchen:
> Ah nice catch. So previously people could use things like '../..' to view files from outside the project directory? Was that the issue?

Yeah. See the link I sent you.

PR #947, opened 2024-09-12
Remove require_selection from autocomplete

richard-to:
Forgot to remove this when creating the autocomplete component initially.

The require_selection property overall is not too helpful for Mesop. The require_selection can be implemented in Mesop by just using the AutocompleteSelectionChangeEvent when populating valid option values.


wwwillchen:
It's a pretty new component, but may be worth a quick check within Google to see if this parameter is being used since removing it is technically a breaking change (worst case, I'd just call it out the next time you sync into google3).

Looking at GitHub, I don't think this parameter is used: https://github.com/search?q=%22+me.autocomplete%28%22+language%3APython&type=code

wwwillchen (approved):

richard-to:
Ah yeah, good point. Yeah, I'll check.

PR #949, opened 2024-09-13
Set default value on autocomplete

### LINKED ISSUE #948: Autocomplete initial value isn't set on the input
https://github.com/google/mesop/blob/8cbdb49a4b7b4d32b1f939cbc3c5411bce12cabf/mesop/components/autocomplete/autocomplete.ng.html#L18

Similar to input: https://github.com/google/mesop/blob/8cbdb49a4b7b4d32b1f939cbc3c5411bce12cabf/mesop/components/input/input.ng.html#L52 - I think we need to set the initial value

richard-to:
Closes #948

### DIFF
--- a/demo/autocomplete.py
+++ b/demo/autocomplete.py
@@ -4,7 +4,7 @@ import mesop as me
 @me.stateclass
 class State:
   raw_value: str
-  selected_value: str
+  selected_value: str = "California"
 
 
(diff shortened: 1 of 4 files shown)


wwwillchen (approved):
Thanks! This will help out a few pages on AI console :)

PR #950, opened 2024-09-13
Add additional validation for line_number_target

richard-to:
Starting to play around with the AI console crud app. Overall, pretty cool. But also some rough edges in the UX. This PR is mainly improves the pydantic validation for the line number target field a bit.

### DIFF
--- a/ai/src/ai/common/example.py
+++ b/ai/src/ai/common/example.py
@@ -11,13 +11,28 @@ import os
 import shutil
 from typing import Generic, Literal, TypeVar
 
-from pydantic import BaseModel
+from pydantic import BaseModel, field_validator
 
 
 class ExampleInput(BaseModel):
   prompt: str
   input_code: str | None = None
   line_number_target: int | None = None
+
+  @field_validator("line_number_target", mode="before")
+  @classmethod
+  # Intentionally leave out annotations since the input value may be ambiguous.
+  def convert_empty_string_to_none(cls, v):
+    if isinstance(v, str) and v == "":
+      return None
+    return v
+
+  @field_validator("line_number_target", mode="after")
+  @classmethod
+  def line_number_greater_than_0(cls, v: int | None) -> int | None:
+    if isinstance(v, int) and v < 1:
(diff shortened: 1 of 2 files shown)


ai/src/ai/common/example.py
```diff
@@ -11,14 +11,36 @@
...
 
 
 class ExampleInput(BaseModel):
   prompt: str
   input_code: str | None = None
   line_number_target: int | None = None
 
+  @field_validator("prompt")
+  @classmethod
+  def is_required(cls, v):
```
richard-to, line 24: Probably could make this more generic since there are other fields that would need this validation, such as the id fields.

ai/src/ai/common/example.py
```diff
@@ -11,14 +11,36 @@
...
+  @classmethod
+  # Intentionally leave out annotations since the input value may be ambiguous.
+  def convert_empty_string_to_none(cls, v):
+    if isinstance(v, str) and v == "":
+      return None
+    return v
+
+  @field_validator("line_number_target", mode="after")
+  @classmethod
+  def line_number_greater_than_0(cls, v: int | None) -> int | None:
```
wwwillchen, line 39: you could also validate that input_code is defined and line_number doesn't go out of bounds on the input code

wwwillchen (approved):

wwwillchen:
> Starting to play around with the AI console crud app. Overall, pretty cool. But also some rough edges in the UX. This PR is mainly improves the pydantic validation for the line number target field a bit.

Cool, good to know and agree about the rough edges. Some of it was actually good in terms of figuring out some pain points in the framework itself, but feel free to file bugs/FRs on issues you see.

ai/src/ai/common/example.py
richard-to: Good idea. I tried to implement it but ran into a slight issue since I think there are times where the Example model gets initialized without input code. The input code is then later updated.

Ended up just checking out of bounds for exceeding line length only if the input code data has already been loaded in the file. Which I think handles the form validation case.

Issue #935, opened 2024-09-10
AI console: Support format golden dataset (to upload for fine-tuning) 


wwwillchen:
Support multiple diff formats:
- Diff (patch)

Support multiple providers
- Gemini

PR #951, opened 2024-09-13
AI console: udiffs, caching & improve eval UX

### LINKED ISSUE #937: AI: experiment with unified diff output

### LINKED ISSUE #940: AI: support seed parameter and caching
Seed parameter should give us determinism in LLM outputs and then we can cache LLM calls to improve iterations

### LINKED ISSUE #941: AI: support deleting evals

### LINKED ISSUE #955: AI: Eval UI can get in a bad state if you try to run Eval but forget to set the API key.
Forgot the error message that occurred but can be reproduced by trying to run a new eval but not setting the API Key for whichever model (in my case Open AI) one. 

To resolve, I think I had to delete the metadata file that was generated.

wwwillchen:
- Fixes #937 - udiff seems to work better out of the box for llama 3.1 than the current diff format
- Fixes #940 - the seed parameter can be used to avoid a cache hit and for some LLM providers it can be used to get a semi-reproducible output
- Fixes #941 
- Partially supports #935 - this only supports the openAI chat format and not Gemini (yet)
- This also supports a few quality of life improvements like fixing the eval scoring, and allowing evals to be deleted
- Fixes #955 

The best performing producer right now is  `llama-70b-udiff-temp-0`


ai/src/ai/offline_common/golden_dataset.py
```diff
@@ -1,9 +1,31 @@
...
+    get_data_path("golden_datasets"), f"{dataset_name}.jsonl"
+  )
+  with open(golden_dataset_path, "w") as f:
+    examples = golden_example_store.get_all()
+    producer_executor = ProducerExecutor(producer_id)
+    for example in examples:
+      messages = producer_executor.get_provider_executor().format_messages(
+        example.input
+      )
+      if example.output.udiff_output is None:
```
richard-to, line 21: Why is udiff output required?

ai/src/generate_udiffs.py
```diff
@@ -0,0 +1,37 @@
+"""
+Generate udiffs for golden examples.
+
+I don't recommend running this script all the time as some of the udiffs
+for the goldens are generated by hand, in particular for:
```
richard-to, line 5: What's the problem with running this against udiffs that were generated by hand? What would happen?

ai/src/ai/common/example.py
```diff
@@ -88,6 +91,10 @@ def get(self, id: str) -> T:
       if os.path.exists(raw_output_path):
         with open(raw_output_path) as f:
           entity.output.raw_output = f.read()
+      udiff_output_path = os.path.join(dir_path, "udiff.txt")
```
richard-to, line 94: Fo raw_output.txt, I'm wondering if we can rename that at some point. Since I think this one contains the git patch diff. And I guess technically also could be the full output (but don't think we have any examples of that).

Or I guess the raw_output.txt can also can udiff output?

Seems like you're leaning toward going with unified diff format, so may be moot in terms of having the flexibility. I think the unified diff format does seem a bit better especially for multiple diffs.

ai/src/ai/common/executor.py
```diff
@@ -198,15 +218,39 @@ def get_provider_executor(self) -> ProviderExecutor:
...
 
-  def execute(self, input: ExampleInput):
-    return self.get_provider_executor().execute(input)
+  def execute(self, input: ExampleInput, seed: int | None = None):
+    provider_executor = self.get_provider_executor()
+    if seed is None:
+      return provider_executor.execute(input, seed=seed)
+
+    # Create a unique cache key
+    cache_key = md5(
```
richard-to, line 227: Would there be scenarios where we want to clear the cache or subsets of the cache? What would be the steps to do so?

ai/src/ai/console/pages/eval_item_page.py
```diff
@@ -89,17 +89,27 @@ def eval_item_page():
               style=me.Style(font_family="monospace", white_space="pre"),
             )
 
-          me.text("Code")
+          me.text("Code code")
```
richard-to, line 92: [question] Why Code code?

richard-to (approved):

ai/src/ai/offline_common/golden_dataset.py
wwwillchen: we should make it configurable - I've updated: https://github.com/google/mesop/issues/935; filed https://github.com/google/mesop/issues/959 to improve our golden output story

ai/src/generate_udiffs.py
wwwillchen: It will override the udiff generated by hand. In particular for `Move_the_icon_buttons_below_the_bot_text_Muy5HQ_20240901000` the udiff would not apply cleanly (i.e. computing the full output based on the udiff) and it's also not a very ideal udiff for LLM. output where it's fragmented.

wwwillchen: Longer-term, it'll be good to customize how we generate the udiffs, but looking at Python's built-in difflib it doesn't look trivial to customize the udiff logic so that it generates larger blocks of diffs that are more coherent for models to output (with the trade-off of having larger diff sizes)

ai/src/ai/common/example.py
wwwillchen: Yeah, I've been wondering about this too. I think there could be some value in having output that's not strictly code (i.e. diff/output) where it's some kind of preamble as a planning stage (e.g. consider the different interpretations of what the user wanted in cases where it's unclear). 

I do think overall udiff is better than our existing diff format because the models, particularly llama, seem to understand it better / prefer it. After some more experimentation, we could choose to just settle on udiff.

ai/src/ai/common/executor.py
wwwillchen: I think there's two reasons to clear the cache: 1) you want to simply re-generate new samples (e.g. temp>0) in which case you could use the seed parameter (as part of the eval creation) or 2) the cache contains buggy outputs in which case you could either modify the cache key (e.g. at first I wasn't capturing all the inputs used into the cache key) or you could delete the entire folder or subsets of the `.cache` dir in the `data` dir

ai/src/ai/console/pages/eval_item_page.py
wwwillchen: oops i meant to say output code

ai/src/ai/offline_common/eval.py
```diff
@@ -60,8 +61,10 @@ def get_eval_examples(eval_id: str) -> list[EvaluatedExample]:
     return []
   examples: list[EvaluatedExample] = []
   for file in os.listdir(eval_path):
-    with open(os.path.join(eval_path, file, "evaluated_example.json")) as f:
-      examples.append(EvaluatedExample.model_validate_json(f.read()))
+    json_path = os.path.join(eval_path, file, "evaluated_example.json")
```
wwwillchen, line 64: Should fix #955

Issue #953, opened 2024-09-13
AI: Add validation to models


richard-to:
Right now, easy to accidentally get the data in a bad state when manually entering input since the pydantic models don't fully validate the data.

richard-to:
Ref #950

Issue #938, opened 2024-09-11
AI console: track created and updated timestamp


wwwillchen:
We can't use git file timestamps

PR #963, opened 2024-09-14
Add some more pydantic validation to AI console

richard-to:
Added some more basic validation to console forms.

Ref #953

### DIFF
--- a/ai/src/ai/common/prompt_fragment.py
+++ b/ai/src/ai/common/prompt_fragment.py
@@ -1,22 +1,15 @@
 from typing import Literal
 
-from pydantic import BaseModel, field_validator, model_validator
+from pydantic import model_validator
 
-from ai.common.entity_store import EntityStore
-from ai.common.model_validators import is_required_str
+from ai.common.entity_store import BaseEntity, EntityStore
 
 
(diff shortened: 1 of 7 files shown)


ai/src/ai/common/prompt_fragment.py
```diff
@@ -12,6 +13,11 @@ class PromptFragment(BaseModel):
   role: Literal["user", "assistant", "system"]
   chain_of_thought: bool = False
 
+  @field_validator("id", mode="after")
```
wwwillchen, line 16: optional: we could have a parent class like BaseEntity similar to BaseExample https://github.com/google/mesop/blob/525ccd8e2c8dc8bd9095a36407754452312b200b/ai/src/ai/common/example.py#L54 to consolidate the validation logic

wwwillchen (approved):

ai/src/ai/common/prompt_fragment.py
wwwillchen: also, it'll make it easier to add timestamps #938

richard-to: Good idea. Updated.

PR #968, opened 2024-09-16
Switch to using localhost

wwwillchen:
For the `mesop` CLI (installed via pip) we will default to using `localhost` as this is more secure and there's typically not a need to have it exposed.

For Colab and Bazel CLI, I've kept it using the externally-accessible host to avoid breaking downstream use cases.


richard-to (approved):
LGTM

Issue #878, opened 2024-08-29
Test failure with state session


wwwillchen:
Test failure:
https://github.com/google/mesop/actions/runs/10606188926/job/29396449678?pr=874

Looks like there's some kind of issue/bug with state sessions:

```
  File "/home/runner/.cache/bazel/_bazel_runner/283da4fe714571e9f73edeff59ade942/execroot/mesop/bazel-out/k8-fastbuild/bin/mesop/cli/cli.runfiles/mesop/mesop/server/state_session.py", line 93, in clear_stale_sessions
    for key, (timestamp, _) in self.cache.items():
RuntimeError: dictionary changed size during iteration
```

**Stack trace:**

```
°·····························°·················································
[WebServer] [WebServer] E0829 00:12:14.158636 139957666571840 _internal.py:97] Error on request:
Traceback (most recent call last):
  File "/home/runner/.cache/bazel/_bazel_runner/283da4fe714571e9f73edeff59ade942/execroot/mesop/bazel-out/k8-fastbuild/bin/mesop/cli/cli.runfiles/my_deps_werkzeug/site-packages/werkzeug/serving.py", line 363, in run_wsgi
    execute(self.server.app)
  File "/home/runner/.cache/bazel/_bazel_runner/283da4fe714571e9f73edeff59ade942/execroot/mesop/bazel-out/k8-fastbuild/bin/mesop/cli/cli.runfiles/my_deps_werkzeug/site-packages/werkzeug/serving.py", line 324, in execute
    application_iter = app(environ, start_response)
  File "/home/runner/.cache/bazel/_bazel_runner/283da4fe714[57](https://github.com/google/mesop/actions/runs/10606188926/job/29396449678?pr=874#step:14:58)1e9f73edeff59ade942/execroot/mesop/bazel-out/k8-fastbuild/bin/mesop/cli/cli.runfiles/my_deps_flask/site-packages/flask/app.py", line 1498, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/runner/.cache/bazel/_bazel_runner/283da4fe714571e9f73edeff[59](https://github.com/google/mesop/actions/runs/10606188926/job/29396449678?pr=874#step:14:60)ade942/execroot/mesop/bazel-out/k8-fastbuild/bin/mesop/cli/cli.runfiles/my_deps_flask/site-packages/flask/app.py", line 1489, in wsgi_app
    ctx.pop(error)
[...]

PR #977, opened 2024-09-18
Try to resolve state sessions race condition

richard-to:
I was not able to reproduce the bug via the integration tests, but this has definitely happened once for me.

This change tries to at lease the reduce the chance of the race condition where we loop through the cache and one of the items has been removed.

Changes:

1. Reduce frequency of trying to clear state cache
2. Add mutex when getting the cache keys
3. Add some additional checks in case cache key does not exist.

Closes https://github.com/google/mesop/issues/878

### DIFF
--- a/mesop/server/state_session.py
+++ b/mesop/server/state_session.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import threading
 from datetime import datetime, timedelta
 from pathlib import Path
 from typing import Any, Protocol
@@ -12,6 +13,8 @@ from mesop.dataclass_utils import (
 )
 from mesop.exceptions import MesopException
 from mesop.server.config import Config, app_config
+
+mutex = threading.Lock()
 
 States = dict[type[Any], object]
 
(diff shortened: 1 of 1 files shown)


mesop/server/state_session.py
```diff
@@ -89,15 +93,37 @@ def save(self, token: str, states: States):
...
-    for key, (timestamp, _) in self.cache.items():
+    if (
+      self.last_checked_sessions + timedelta(minutes=self._SESSION_TTL_MINUTES)
+      > current_time
+    ):
+      return
+
+    self.last_checked_sessions = current_time
+
+    # Add a mutex in the case where a cache item may be removed while looping through
```
wwwillchen, line 106: it's not clear to me how beneficial this mutex is - it seems like either it's too short (you should take a mutex out for the entire block (e.g. up until you delete the cache key in L122) or you should just skip having a mutex and accept a race because you're doing error-handling in L123. WDYT?

wwwillchen (approved):

mesop/server/state_session.py
richard-to: Hmm, yeah that's a good point. Since I'm not doing a loop anymore, probably wouldn't run into the data changed while looping runtime error now. I'll remove it.

PR #983, opened 2024-09-25
Support pydantic BaseModel classes in state

### LINKED ISSUE #979: Support pydantic models in state serialization

wwwillchen:
Fixes #979

### DIFF
--- a/mesop/dataclass_utils/dataclass_utils.py
+++ b/mesop/dataclass_utils/dataclass_utils.py
@@ -68,6 +68,14 @@ def dataclass_with_defaults(cls: Type[C]) -> Type[C]:
 
   annotations = get_type_hints(cls)
   for name, type_hint in annotations.items():
+    if (
+      isinstance(type_hint, type)
+      and has_parent(type_hint)
+      and issubclass(type_hint, BaseModel)
+    ):
+      pydantic_model_cache[(type_hint.__module__, type_hint.__qualname__)] = (
+        type_hint
+      )
     if name not in cls.__dict__:  # Skip if default already set
       if type_hint == int:
         setattr(cls, name, field(default=0))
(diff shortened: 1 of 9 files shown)


mesop/dataclass_utils/dataclass_utils.py
```diff
@@ -221,6 +238,18 @@ def decode_mesop_json_state_hook(dct):
...
+    cache_key = (
+      dct[_PYDANTIC_OBJECT_KEY]["module"],
+      dct[_PYDANTIC_OBJECT_KEY]["qualname"],
+    )
+    if cache_key not in pydantic_model_cache:
+      raise MesopException(
+        f"Tried to deserialize Pydantic model, but it's not in the cache: {cache_key}"
+      )
+    model_class = pydantic_model_cache[cache_key]
+    return model_class.model_validate_json(dct[_PYDANTIC_OBJECT_KEY]["json"])
```
richard-to, line 251: Does model_validate_json validate the model if it has validators on it -- or is it just the JSON? Could there be a case where the model has invalid data based on validator logic ?

richard-to: Also curious about the deserialization since it uses pydantic. Could it be assumed that it could handle some serialization errors that we currently run into (such as tuples, unions, etc).

mesop/dataclass_utils/dataclass_utils_test.py
```diff
@@ -49,6 +50,16 @@ class WithUploadedFile:
   data: UploadedFile = field(default_factory=UploadedFile)
 
 
+class PydanticModel(BaseModel):
```
richard-to, line 53: Can we add a more complicated example:

1. Nested PydanticModel
2. List of PydanticModels
3. More complicated data types / annotations in the PydanticModel

richard-to (approved):
Nice. This is pretty interesting and useful.

mesop/dataclass_utils/dataclass_utils_test.py
wwwillchen: good idea - done.

mesop/dataclass_utils/dataclass_utils.py
wwwillchen: model_validate_json should validate with the validators - see: https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_validate_json.

Yeah - see the test cases I added, this is much more robust than what we currently support. I'm going to try using this in my personal project and if it turns out to be solid, I think we can add a recommendation in our [docs](https://google.github.io/mesop/guides/state-management/) to use Pydantic.

Issue #960, opened 2024-09-13
Configurable static assets


wwwillchen:
More details coming.

ustas4:
And load icons from local/offline PC impossible to. Or I can't understand how to do that

Issue #957, opened 2024-09-13
AI: Dark mode toggle is not remembered per page


richard-to:
Think this is more a limitation of Mesop right now (could do it by passing query param, but kind of ugly), but could be nice if the dark mode toggle could not revert back to the default system choice every time I change the page (if I override the toggle).

wwwillchen:
Yeah, I noticed this too and I think we should consider having some kind of app-wide settings for dark theme so that the preferences are set across all pages and persisted across navigations. There are ways around it, e.g. query param, passing into state, but they're quite inconvenient.

At the same time, I think there could be usages of app-wide settings like having a `static` dir for serving assets, e.g. JS, CSS, favicon, images, etc.

PR #989, opened 2024-09-27
Add option expose static folder

richard-to:
Flask allows a static assets directory be specified. It turns out this can be an absolute path or a relative path.

I'm not sure if I'm doing the relatively path part correctly for bazel or the pip version.

This is not completed yet. Still need to add docs, tests. But mainly just wanted to see if this is reasonable.

One issue is that specifying the stylesheet from the page object in terms of CSP:

```
The source list for the Content Security Policy directive 'style-src' contains an invalid source: '/assets/tailwind.css'. It will be ignored.
```
Interestingly the CSS seems to be still read correctly.

Ref: https://github.com/google/mesop/issues/960

### DIFF
--- a/demo/static/tailwind.css
+++ b/demo/static/tailwind.css
@@ -1,0 +1,732 @@
+*,
+::before,
+::after {
+  --tw-border-spacing-x: 0;
+  --tw-border-spacing-y: 0;
+  --tw-translate-x: 0;
+  --tw-translate-y: 0;
+  --tw-rotate: 0;
+  --tw-skew-x: 0;
+  --tw-skew-y: 0;
+  --tw-scale-x: 1;
+  --tw-scale-y: 1;
+  --tw-pan-x:  ;
+  --tw-pan-y:  ;
+  --tw-pinch-zoom:  ;
+  --tw-scroll-snap-strictness: proximity;
+  --tw-gradient-from-position:  ;
+  --tw-gradient-via-position:  ;
+  --tw-gradient-to-position:  ;
+  --tw-ordinal:  ;
+  --tw-slashed-zero:  ;
+  --tw-numeric-figure:  ;
+  --tw-numeric-spacing:  ;
+  --tw-numeric-fraction:  ;
+  --tw-ring-inset:  ;
+  --tw-ring-offset-width: 0px;
+  --tw-ring-offset-color: #fff;
+  --tw-ring-color: rgb(59 130 246 / 0.5);
+  --tw-ring-offset-shadow: 0 0 #0000;
+  --tw-ring-shadow: 0 0 #0000;
+  --tw-shadow: 0 0 #0000;
+  --tw-shadow-colored: 0 0 #0000;
+  --tw-blur:  ;
+  --tw-brightness:  ;
+  --tw-contrast:  ;
+  --tw-grayscale:  ;
+  --tw-hue-rotate:  ;
+  --tw-invert:  ;
+  --tw-saturate:  ;
+  --tw-sepia:  ;
+  --tw-drop-shadow:  ;
+  --tw-backdrop-blur:  ;
+  --tw-backdrop-brightness:  ;
+  --tw-backdrop-contrast:  ;
+  --tw-backdrop-grayscale:  ;
+  --tw-backdrop-hue-rotate:  ;
+  --tw-backdrop-invert:  ;
+  --tw-backdrop-opacity:  ;
+  --tw-backdrop-saturate:  ;
+  --tw-backdrop-sepia:  ;
+  --tw-contain-size:  ;
+  --tw-contain-layout:  ;
+  --tw-contain-paint:  ;
+  --tw-contain-style:  ;
+}
+
+::backdrop {
+  --tw-border-spacing-x: 0;
+  --tw-border-spacing-y: 0;
+  --tw-translate-x: 0;
+  --tw-translate-y: 0;
+  --tw-rotate: 0;
+  --tw-skew-x: 0;
+  --tw-skew-y: 0;
+  --tw-scale-x: 1;
+  --tw-scale-y: 1;
+  --tw-pan-x:  ;
+  --tw-pan-y:  ;
+  --tw-pinch-zoom:  ;
+  --tw-scroll-snap-strictness: proximity;
+  --tw-gradient-from-position:  ;
+  --tw-gradient-via-position:  ;
+  --tw-gradient-to-position:  ;
+  --tw-ordinal:  ;
+  --tw-slashed-zero:  ;
+  --tw-numeric-figure:  ;
+  --tw-numeric-spacing:  ;
+  --tw-numeric-fraction:  ;
+  --tw-ring-inset:  ;
+  --tw-ring-offset-width: 0px;
+  --tw-ring-offset-color: #fff;
+  --tw-ring-color: rgb(59 130 246 / 0.5);
+  --tw-ring-offset-shadow: 0 0 #0000;
+  --tw-ring-shadow: 0 0 #0000;
+  --tw-shadow: 0 0 #0000;
+  --tw-shadow-colored: 0 0 #0000;
+  --tw-blur:  ;
+  --tw-brightness:  ;
+  --tw-contrast:  ;
+  --tw-grayscale:  ;
+  --tw-hue-rotate:  ;
+  --tw-invert:  ;
+  --tw-saturate:  ;
+  --tw-sepia:  ;
+  --tw-drop-shadow:  ;
+  --tw-backdrop-blur:  ;
+  --tw-backdrop-brightness:  ;
+  --tw-backdrop-contrast:  ;
+  --tw-backdrop-grayscale:  ;
+  --tw-backdrop-hue-rotate:  ;
+  --tw-backdrop-invert:  ;
+  --tw-backdrop-opacity:  ;
+  --tw-backdrop-saturate:  ;
+  --tw-backdrop-sepia:  ;
+  --tw-contain-size:  ;
+  --tw-contain-layout:  ;
+  --tw-contain-paint:  ;
+  --tw-contain-style:  ;
+}
+
+/*
+! tailwindcss v3.4.14 | MIT License | https://tailwindcss.com
+*/
+
+/*
+1. Prevent padding and border from affecting element width. (https://github.com/mozdevs/cssremedy/issues/4)
+2. Allow adding a border to an element by just adding a border-width. (https://github.com/tailwindcss/tailwindcss/pull/116)
+*/
+
+*,
+::before,
+::after {
+  box-sizing: border-box;
+  /* 1 */
+  border-width: 0;
+  /* 2 */
+  border-style: solid;
+  /* 2 */
+  border-color: #e5e7eb;
+  /* 2 */
+}
+
+::before,
+::after {
+  --tw-content: '';
+}
+
+/*
+1. Use a consistent sensible line-height in all browsers.
+2. Prevent adjustments of font size after orientation changes in iOS.
+3. Use a more readable tab size.
+4. Use the user's configured `sans` font-family by default.
+5. Use the user's configured `sans` font-feature-settings by default.
+6. Use the user's configured `sans` font-variation-settings by default.
+7. Disable tap highlights on iOS
+*/
+
+html,
+:host {
+  line-height: 1.5;
+  /* 1 */
+  -webkit-text-size-adjust: 100%;
+  /* 2 */
+  -moz-tab-size: 4;
+  /* 3 */
+  -o-tab-size: 4;
+  tab-size: 4;
+  /* 3 */
+  font-family: ui-sans-serif, system-ui, sans-serif, 'Apple Color Emoji',
+    'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
+  /* 4 */
+  font-feature-settings: normal;
+  /* 5 */
+  font-variation-settings: normal;
+  /* 6 */
+  -webkit-tap-highlight-color: transparent;
+  /* 7 */
+}
+
+/*
+1. Remove the margin in all browsers.
+2. Inherit line-height from `html` so users can set them as a class directly on the `html` element.
+*/
+
+body {
+  margin: 0;
+  /* 1 */
+  line-height: inherit;
+  /* 2 */
+}
+
+/*
+1. Add the correct height in Firefox.
+2. Correct the inheritance of border color in Firefox. (https://bugzilla.mozilla.org/show_bug.cgi?id=190655)
+3. Ensure horizontal rules are visible by default.
+*/
+
+hr {
+  height: 0;
+  /* 1 */
+  color: inherit;
+  /* 2 */
+  border-top-width: 1px;
+  /* 3 */
+}
+
+/*
+Add the correct text decoration in Chrome, Edge, and Safari.
+*/
+
+abbr:where([title]) {
+  -webkit-text-decoration: underline dotted;
+  text-decoration: underline dotted;
+}
+
+/*
+Remove the default font size and weight for headings.
+*/
+
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  font-size: inherit;
+  font-weight: inherit;
+}
+
+/*
+Reset links to optimize for opt-in styling instead of opt-out.
+*/
+
+a {
+  color: inherit;
+  text-decoration: inherit;
+}
+
+/*
+Add the correct font weight in Edge and Safari.
+*/
+
+b,
+strong {
+  font-weight: bolder;
+}
+
+/*
+1. Use the user's configured `mono` font-family by default.
+2. Use the user's configured `mono` font-feature-settings by default.
+3. Use the user's configured `mono` font-variation-settings by default.
+4. Correct the odd `em` font sizing in all browsers.
+*/
+
+code,
+kbd,
+samp,
+pre {
+  font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas,
+    'Liberation Mono', 'Courier New', monospace;
+  /* 1 */
+  font-feature-settings: normal;
+  /* 2 */
+  font-variation-settings: normal;
+  /* 3 */
+  font-size: 1em;
+  /* 4 */
+}
+
+/*
+Add the correct font size in all browsers.
+*/
+
+small {
+  font-size: 80%;
+}
+
+/*
+Prevent `sub` and `sup` elements from affecting the line height in all browsers.
+*/
+
+sub,
+sup {
+  font-size: 75%;
+  line-height: 0;
+  position: relative;
+  vertical-align: baseline;
+}
+
+sub {
+  bottom: -0.25em;
+}
+
+sup {
+  top: -0.5em;
+}
+
+/*
+1. Remove text indentation from table contents in Chrome and Safari. (https://bugs.chromium.org/p/chromium/issues/detail?id=999088, https://bugs.webkit.org/show_bug.cgi?id=201297)
+2. Correct table border color inheritance in all Chrome and Safari. (https://bugs.chromium.org/p/chromium/issues/detail?id=935729, https://bugs.webkit.org/show_bug.cgi?id=195016)
+3. Remove gaps between table borders by default.
+*/
+
+table {
+  text-indent: 0;
+  /* 1 */
+  border-color: inherit;
+  /* 2 */
+  border-collapse: collapse;
+  /* 3 */
+}
+
+/*
+1. Change the font styles in all browsers.
+2. Remove the margin in Firefox and Safari.
+3. Remove default padding in all browsers.
+*/
+
+button,
+input,
+optgroup,
+select,
+textarea {
+  font-family: inherit;
+  /* 1 */
+  font-feature-settings: inherit;
+  /* 1 */
+  font-variation-settings: inherit;
+  /* 1 */
+  font-size: 100%;
+  /* 1 */
+  font-weight: inherit;
+  /* 1 */
+  line-height: inherit;
+  /* 1 */
+  letter-spacing: inherit;
+  /* 1 */
+  color: inherit;
+  /* 1 */
+  margin: 0;
+  /* 2 */
+  padding: 0;
+  /* 3 */
+}
+
+/*
+Remove the inheritance of text transform in Edge and Firefox.
+*/
+
+button,
+select {
+  text-transform: none;
+}
+
+/*
+1. Correct the inability to style clickable types in iOS and Safari.
+2. Remove default button styles.
+*/
+
+button,
+input:where([type='button']),
+input:where([type='reset']),
+input:where([type='submit']) {
+  -webkit-appearance: button;
+  /* 1 */
+  background-color: transparent;
+  /* 2 */
+  background-image: none;
+  /* 2 */
+}
+
+/*
+Use the modern Firefox focus style for all focusable elements.
+*/
+
+:-moz-focusring {
+  outline: auto;
+}
+
+/*
+Remove the additional `:invalid` styles in Firefox. (https://github.com/mozilla/gecko-dev/blob/2f9eacd9d3d995c937b4251a5557d95d494c9be1/layout/style/res/forms.css#L728-L737)
+*/
+
+:-moz-ui-invalid {
+  box-shadow: none;
+}
+
+/*
+Add the correct vertical alignment in Chrome and Firefox.
+*/
+
+progress {
+  vertical-align: baseline;
+}
+
+/*
+Correct the cursor style of increment and decrement buttons in Safari.
+*/
+
+::-webkit-inner-spin-button,
+::-webkit-outer-spin-button {
+  height: auto;
+}
+
+/*
+1. Correct the odd appearance in Chrome and Safari.
+2. Correct the outline style in Safari.
+*/
+
+[type='search'] {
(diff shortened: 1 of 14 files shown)


mesop/server/server.py
```diff
@@ -513,3 +525,16 @@ def sse_request(
         if decoded_line.startswith(SSE_DATA_PREFIX):
           event_data = json.loads(decoded_line[len(SSE_DATA_PREFIX) :])
           yield event_data
+
+
+def get_static_assets_folder() -> str | None:
```
wwwillchen, line 530: I think we should consider making this configurable by code instead of env var, because the static assets folder is something that almost by definition would involve code changes anyway (e.g. moving where your JS / images are).

Right now we lack any kind of top-level Mesop app configuration but I think it may be reasonable to add one, both for this use case and related to #957 we can configure app-wide default theme mode. The API could look something like this:


```py
me.configure_app(
  me.AppSettings(
    static_dir="assets", 
    static_url_path="/assets", 
    default_theme_mode="system"
))
```

mesop/server/server.py
```diff
@@ -49,6 +51,8 @@
 
 STREAM_END = "data: <stream_end>\n\n"
 
+DEFAULT_ASSETS_URL_PATH = "/assets"
```
wwwillchen, line 54: I think this is a reasonable default value, but we should probably also make it configurable  (e.g. they may already have a page with the path "/assets")

mesop/server/server.py
wwwillchen: As prior art, here's how Django configures a similar feature: https://docs.djangoproject.com/en/5.1/ref/settings/#staticfiles-dirs - they also use settings and allows configuring both dir and url path (they are also much more flexible/powerful in general)

mesop/server/server.py
```diff
@@ -513,3 +525,16 @@ def sse_request(
...
           event_data = json.loads(decoded_line[len(SSE_DATA_PREFIX) :])
           yield event_data
+
+
+def get_static_assets_folder() -> str | None:
+  folder = os.environ.get("MESOP_STATIC_ASSETS_FOLDER", "")
+  if not folder:
+    return None
+
+  if not os.path.isabs(folder):
```
wwwillchen, line 535: I think we should do relative dir paths for the static assets folder (we can consider the project root the dir that contains the mesop app entrypoint file, e.g. main.py) because:

1) it covers the common use case
2) it prevents security issues (e.g. if there's an absolute dir path like "/" because somebody is confused about this, they may end up exposing sensitive files). Also, we should be careful even with a relative dir path that you cannot traverse above the app dir (e.g. "../foo" should be prohibited) because similarly this can be problematic for security reasons

wwwillchen (commented):
Cool, didn't know about this Flask API, thanks for sharing this!

A couple more considerations (sorry for the terse issue description):

1. I'd like to use this static assets dir feature to (eventually) fully remediate the issue from https://github.com/google/mesop/security/advisories/GHSA-pmv9-3xqp-8w42 - right now we will still serve any JS and CSS files within your Mesop app file subtree but eventually I think we should migrate to a stricter model (and then we can remove the security warning on https://google.github.io/mesop/guides/web-security/#static-file-serving). Eventually, we should only serve JS and CSS files which are either in the static dir _or_ somehow explicitly allowlisted, for example, because it's convenient to have the JS web component files co-located near the python file, we can automatically consider them as allow-listed. The reason why I didn't already do this stricter model is that then there isn't a way to allowlist shared JS modules ([example](https://github.com/google/mesop/blob/bb8fd9841908a956e68e43f64f9e2999b0166a75/mesop/examples/web_component/shared_js_module/shared_module.js#L1)), but now with static assets dir, we could require developers to put their shared JS modules in the static dir. We could do this over multiple versions since this would be technically a breaking change (although I think the usage of shared JS modules for our web components is probably quite small): first, we release the static dir feature, which you must explicitly opt-in (and if you opt-in, then we'll do the strict static file serving enforcement), and then in the next version, we have the static dir feature be default on (e.g. set to "/assets/" by default) and then everyone will have the strict static file serving enforcement. The other benefit with spreading this across releases is that in case for some reason people are putting sensitive files in their ./assets dir, they would have a mesop version/release to migrate (e.g either moving their sensitive files to a different dir or configuring a different static dir name)
2. This is much less important than the above point, but it would be nice to support favicons #563  :) - for example, if there's a favicon.ico inside the static dir, we will use it by default, otherwise we'll use the current default mesop favicon

wwwillchen:
Re: CSP error - I'm not 100% sure, but I think what's happening is that you're probably getting a CSP value like: `style-src 'self'  https://cdn.example.com /static/foo.css;` and the "/static/foo.css" is invalid because it's after the cdn.example.com and not `'self'`. Since we already allowlist any CSS on `self`, we should update the logic in https://github.com/google/mesop/blob/bb8fd9841908a956e68e43f64f9e2999b0166a75/mesop/server/static_file_serving.py#L228 so that any relative paths are just omitted from the CSP (since they're already implied by self)

mesop/server/server.py
richard-to: Yup. I agree that this should be configurable.

richard-to:
> A couple more considerations (sorry for the terse issue description):

Hah, no problem. Figured you had more in mind which is one reason I just threw out a quick draft to start

> I'd like to use this static assets dir feature to (eventually) fully remediate the issue from [GHSA-pmv9-3xqp-8w42](https://github.com/google/mesop/security/advisories/GHSA-pmv9-3xqp-8w42) - right now we will still serve any JS and CSS files within your Mesop app file subtree but eventually I think we should migrate to a stricter model (and then we can remove the security warning on https://google.github.io/mesop/guides/web-security/#static-file-serving). Eventually, we should only serve JS and CSS files which are either in the static dir or somehow explicitly allowlisted, for example, because it's convenient to have the JS web component files co-located near the python file, we can automatically consider them as allow-listed. The reason why I didn't already do this stricter model is that then there isn't a way to allowlist shared JS modules ([example](https://github.com/google/mesop/blob/bb8fd9841908a956e68e43f64f9e2999b0166a75/mesop/examples/web_component/shared_js_module/shared_module.js#L1)), but now with static assets dir, we could require developers to put their shared JS modules in the static dir. We could do this over multiple versions since this would be technically a breaking change (although I think the usage of shared JS modules for our web components is probably quite small): first, we release the static dir feature, which you must explicitly opt-in (and if you opt-in, then we'll do the strict static file serving enforcement), and then in the next version, we have the static dir feature be default on (e.g. set to "/assets/" by default) and then everyone will have the strict static file serving enforcement. The other benefit with spreading this across releases is that in case for some reason people are putting sensitive files in their ./assets dir, they would have a mesop version/release to migrate (e.g either moving their sensitive files to a different dir or configuring a different static dir name)

Thanks for the details here. I think we could allow multiple static directories in Flask through the use of blueprints. That is a bit hacky since we'd only use it for the static directories. And I believe this could work if we allow web components to be install as pip installed modules as well (though I'd need to verify).

> This is much less important than the above point, but it would be nice to support favicons https://github.com/google/mesop/issues/563 :) - for example, if there's a favicon.ico inside the static dir, we will use it by default, otherwise we'll use the current default mesop favicon

Agree. Seems easy enough.

mesop/server/server.py
```diff
@@ -270,3 +281,9 @@ def handle_message(message):
     flask_app.socketio = socketio  # type: ignore
 
   return flask_app
+
+
+def get_static_folder() -> str | None:
+  if not app_config.static_folder:
+    return None
+  return safe_join(os.getcwd(), app_config.static_folder)
```
richard-to, line 289: Wasn't sure how best to support runfiles for this case. I feel like using os.getcwd() worked in bazel dev by chance.

mesop/BUILD
```diff
@@ -15,7 +15,10 @@ py_library(
     srcs = [
         "__init__.py",
     ],
-    data = glob([".env"]),
+    data = glob([
+        ".env",
+        "static/**/*",
```
richard-to, line 20: Not sure if this is the best way to add static folder support for bazel dev, especially in the downstream scenario.

demo/static/tailwind.css
richard-to, line 1: When deploying the demo site, would need to set MESOP_STATIC_FOLDER variable in hugging face config

richard-to (commented):
Update the implementation a bit. Ended up still going with environment variables for now.

I experimented with adding --config flag, but the issue I ran into with that scenario was that it seemed hard to enforce the code that the users would run there.

I also tried add something `me.configure_app(AppSettings(...))` which almost worked, but one issue is that the CSS in me.page gets loaded before me.configure_app gets run. I think this is because we me.configure_app is global and it may not run before all me.page files are executed. I think it's possible that this issue could be resolved. But the other issue with using me.configure_app is that it's cross cutting across different pages, so for the Bazel app it's a bit weird since we have demo app which can be run standalone and also the examples. 

Example code here:

https://github.com/richard-to/mesop/commit/b6df9632daa3775bd4ce5b2a6af1e32700bfd708

mesop/server/server.py
wwwillchen: OK, env var seems OK since we use it for everything else already.

Re: web-component dir structure, I think it depends on how people compile their web component JS code. For example, if each are standalone binaries, then your structure might make sense, but if they're compiling all their web components into 1 JS binary (e.g. to share a common UI library), then it'll be in 1 folder naturally. I think having 1 static dir to start with is good.

wwwillchen (changes requested):
thx - mostly looks good - added some comments

mesop/server/server.py
richard-to: Yeah, it was weird. I tried that initially, but it kept erroring out due to the base path being different than I was expecting.

```
FileNotFoundError: [Errno 2] No such file or directory: '/private/var/tmp/_bazel_/4fbe097e24ba789ddebddac4ca5258ab/execroot/mesop/bazel-out/darwin_arm64-fastbuild/bin/mesop/web/src/app/editor/web_package/static'
```

I was expecting it to be:

```
'/private/var/tmp/_bazel_/4fbe097e24ba789ddebddac4ca5258ab/execroot/mesop/bazel-out/darwin_arm64-fastbuild/bin/mesop/mesop/static'
```

But I'll try to test if os.cwd works downstream and for pip install version. I'm not too confident about it.

mesop/BUILD
richard-to: Yeah, this one is slightly annoying because we have to manually set the folder to mesop/static. I tried to move the static folder down to the root level, but I think I didn't set up the BUILD file right so that it was included in the runfiles.

demo/static/tailwind.css
richard-to: I agree that it would be nicer to default to static. But one reason I didn't was because of the rare case where someone using Mesop has a folder called static that they didn't intend to expose.

Right now flask does expose the static folder, but it doesn't expose the static folder at the user's application root folder (I tested this out manually just to verify).

So we may need some kind of incremental warning or something before we can make it the default.

mesop/static/tailwind.css
```diff
@@ -0,0 +1,732 @@
+*,
```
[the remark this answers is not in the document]
richard-to: Yeah I was wondering about that scenario. Ok. Updated with license.

mesop/server/config.py
```diff
@@ -24,6 +24,8 @@ class Config(BaseModel):
   state_session_backend_firestore_collection: str = "mesop_state_sessions"
   state_session_backend_sql_connection_uri: str = ""
   state_session_backend_sql_table: str = "mesop_state_sessions"
+  static_folder: str = ""
```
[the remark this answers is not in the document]
richard-to: I think we need to go with B for now just for the edge case of someone coincidentally already having a static folder.

I agree that we will need the ability to expose multiple static folders. That would become more important for bigger Mesop apps. I tried a few different interfaces for exposing multiple static folders, but ultimately wasn't sure, so decided to start with just one static folder. And I'm hoping that there will still be a path to expand to multiple static folders in the future

I think for multiple static folders, it would become more of a necessity if there were a pattern where Mesop could be used more modularly. Than those shared components may need their own static folder. In Flask, this would be akin to the Flask Blueprints.

In terms of the web component use case, I think that those shouldn't need to go into the same static folder.

I think would just need to update the expected format for web components.

So for this route, I think we should do something like:

```
 @app.route(f"/{WEB_COMPONENTS_PATH_SEGMENT}/static/<path:path>")
 def serve_web_components(path: str):
```

I think then the python files and javascript/css files can exist in the same folder and be more easily packageble. But the web viewable output would be in the static folder.

This would be a breaking change though.

Agree about updating the docs. Just wanting to make sure we settled on the interface for this first.

mesop/server/server.py
richard-to: Hmm, that's an interesting about point about compiling all their web components into a single JS binary. Is that possible to do now?

But couldn't they then do something like:

```
- web_component1
   - web_component1.py
   - file.js
   - file.js
   - file.scss
- web_component2
  - web_component2.py
  - file.ts
  - file.ts
  - file.css
- combined_web_component
  - web_component1.py
  - web_component2.py
  - static
     - bundle.js
     - bundle.css
```

So in this case combined_web_component, is created through some kind of build process where in the web component python files just get moved.

Granted, I'm hand-waving a bunch of stuff since moving the python files would entail changing import paths. And also is it possible to point to the same bundle.js in the web component path folder?

mesop/server/server.py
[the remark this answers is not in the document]
richard-to: Ok added some tests.

richard-to:
I added some docs. But right now only for the Config section which isn't all that prominent. I'm also considering add another page under Guides / Enhancements. But still thinking what to write that wouldn't be redundant to the Config section. But maybe could add some examples for integrating Bootstrap and Tailwind.

mesop/server/server.py
wwwillchen: I've invited you to my personal Mesop app repo - you can see how I basically compile all the web components into 1 JS binary. Besides being more convenient to setup compilation-wise, it provides a significant benefit in terms of reducing code size because then you can reuse the same libraries (e.g. code mirror) across web components. Note: when you import the same JS module multiple times, the browser automatically uses the cached import after the first time.

This is actually a pretty big issue with our current web component story, which is that if you have N web components which want to use the same library (e.g. codemirror, marked, etc), they will end up compiling it themselves _unless_ they somehow dedupe the library (e.g. CDN, or some other way of dynamically linking the library instead of compiling it in).  Other frameworks like Angular have solved this in a pretty involved way: https://angular.dev/tools/libraries/angular-package-format (not an exact parallel but vaguely resembles what we would need to do to solve this issue)

mesop/server/config.py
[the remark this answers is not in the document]
wwwillchen: OK sounds good.

Not sure if I totally understand the breaking change that you're proposing - is it that web components have to be in a separate directory from the Python code? Maybe it's just me, but I do like having the Python and JS files right next to each other (in the same dir) because they are so closely linked.

We could offer both though:

```py
@mel.web_component(path="./inner_component.js", static_folder="./static/")
```

This way you can keep the entry point file in the same directory (or somewhere else) and optionally define a static folder for JS modules / other things that are imported by the entry point.

Taking a step back though, I'm not super optimistic about shared web components (at least in the short-term) so I I'm not too worried about us solving this soon. On one hand there's the performance / shared library issue (see above) which isn't really easy to solve unless you vendor in web components (i.e. copy paste) and on the other hand, it doesn't seem like anyone in the open-source community is really building non-trivial web components that you would want to share (vs. just vendoring)

mesop/server/server.py
richard-to: Thanks for the context. That's good to know.

mesop/server/config.py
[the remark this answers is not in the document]
richard-to: Yeah I don't have a strong stance on this. So I think having a static folder option could be a nice compromise for people who prefer having the files in the same directory. I think with the static folder, then one could include images or other non-css / js assets a bit easier. Otherwise from what I understand, we'd have to do some kind of allowlisting of the files in the folder that are accessible.

richard-to:
- Ok, updated the docs with some example use cases for using the static folders (Loading logos / Tailwind CSS)
- Tested that the static folder is viewable using a pip install
- Also was able to test the static folder downstream and it worked

mesop/server/server.py
```diff
@@ -270,3 +277,15 @@ def handle_message(message):
     flask_app.socketio = socketio  # type: ignore
 
   return flask_app
+
+
+def get_static_folder() -> str | None:
+  if not app_config.static_folder:
```
wwwillchen, line 283: if static_folder is ".", then I think a user could end up serving their entire mesop app directory as a static asset. I think we should add some kind of validation (probably both here and in config.py) to make sure that users don't configure bad static folder paths

mesop/server/server.py
wwwillchen: One thing that makes me a little nervous is that we're using `os.getcwd()` - although we're already using something similar for web components serving, in that case we have a file extension check.

There could be a security issue if a Mesop app developer (for some reason) has the cwd be in some top-level directory (e.g. `/`) and they do something like this:

```
MESOP_STATIC_FOLDER="usr" mesop /usr/home/local/app.py
```

This is a somewhat contrived example, but I think we should clearly document this both in the guide and api docs. 

I think resolving this relative to the Mesop app entrypoint file could be even better (both here and for web component serving), but that's a larger change and I think something we could defer as long as we clearly document for now. 

For example, here's how flask resolves to the root path (which is what they use for resolving relative paths for static_folder):

https://github.com/pallets/flask/blob/dffe3034825edc4f84aded013fc9c6590e545ed5/src/flask/helpers.py#L569

It seems non-trivial and even then they are falling back to os.getcwd() in some cases.

wwwillchen (commented):
Overall looks good! My two main comments are security-related on server.py

Thanks for documenting this well!

docs/guides/static-assets.md
```diff
@@ -0,0 +1,129 @@
...
+
+Mesop allows you to specify a folder for storing static assets that will be served by
+the Mesop server.
+
+## Enable a static folder
+
+This feature can be enabled using environment variables.
+
+- [MESOP_STATIC_FOLDER](../api/config.md#mesop_static_folder)
+- [MESOP_STATIC_URL_PATH](../api/config.md#mesop_static_folder)
```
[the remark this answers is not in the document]
richard-to: Doh, copy-pasta'd again. 🤦🏾

docs/guides/static-assets.md
```diff
@@ -0,0 +1,129 @@
...
+import mesop as me
+
+@me.page(stylesheets=["/static/tailwind.css"])
+def foo():
+  with me.box(classes="bg-gray-800"):
+    me.text("Mesop with Tailwind CSS.")
+```
+
+Tailwind is able to extract the CSS properties from your Mesop main.py file. This does
+not work for all cases. If you have dynamic properties, then you may need to maintain
```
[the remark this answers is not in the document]
richard-to: Great point. Yes, I believe it's for cases where like string concatenation or string formatting.

But things like `'text-sm' if 'small' else 'text-lg'` would be fine since it's statically defined. At least that's my understanding.

Updated.

docs/guides/static-assets.md
```diff
@@ -0,0 +1,129 @@
...
+- [MESOP_STATIC_URL_PATH](../api/config.md#mesop_static_folder)
+
+Full description of these two settings can be found on the [config page](../api/config.md).
+
+### Enabling a static folder named "assets"
+
+This will make the files in the `assets` directory accessible from the Mesop server
+at `/static`.
+
+Mesop will look for the `assets` directory relative to your application directory.
```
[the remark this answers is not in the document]
richard-to: Ah great point. Totally forgot about that scenario. Updated.

docs/api/config.md
```diff
@@ -6,6 +6,28 @@ Mesop is configured at the application level using environment variables.
...
 
+### MESOP_STATIC_FOLDER
+
+Allows access to static files from the Mesop server.
+
+The specified folder path must be relative to the Mesop app root directory.
+
+By default, this feature is not enabled.
+
+**Default:** `''`
```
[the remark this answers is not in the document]
richard-to: Great points. Updated with examples.

richard-to:
> One thing that makes me a little nervous is that we're using os.getcwd() - although we're already using something similar for web components serving, in that case we have a file extension check.
> 
> There could be a security issue if a Mesop app developer (for some reason) has the cwd be in some top-level directory (e.g. /) and they do something like this:
> 
> MESOP_STATIC_FOLDER="usr" mesop /usr/home/local/app.py
> This is a somewhat contrived example, but I think we should clearly document this both in the guide and api docs.
> 
> I think resolving this relative to the Mesop app entrypoint file could be even better (both here and for web component serving), but that's a larger change and I think something we could defer as long as we clearly document for now.
> 
> For example, here's how flask resolves to the root path (which is what they use for resolving relative paths for static_folder):
> 
> https://github.com/pallets/flask/blob/dffe3034825edc4f84aded013fc9c6590e545ed5/src/flask/helpers.py#L569
> 
> It seems non-trivial and even then they are falling back to os.getcwd() in some cases.

Yeah that's a great point. I totally overlooked the use case where the develop runs the app from a different directory than the entrypoint file.

I think at least for the Mesop case, we currently have fewer ways to run the app than Flask, so our function probably would not need to be as robust (though I could definitely be wrong). Haven't had a chance to look, but I think the main question is if we can get the location of the main.py file at the time the Flask server initializes which I think is fine since we have the argv from the main function before calling configure_flask_app.

I'll look into a bit more tomorrow. Seems like it would be good to wait since this could be a breaking change with the difference between cwd and relative to the main.py file.

richard-to:
I took another look in terms of the finding the path to the entry point file

I guess not surprisingly, we had more entrypoint use cases than I thought:

1. Bazel entrypoint -> I think using os.getcwd() is fine. I guess it makes the path to static folder a bit longer.
2. Pip install -> For the basic case, can we can pass in the absolute path which we already get from bin.py
3. Colab -> I'm not familiar enough with Colab. So not sure what would be the path we'd need here. Would get.getcwd be sufficient here?
4. Gunicorn -> Not sure yet how we'd get the path to main.py. I guess in that case may need to use the approaches you pointed out in the Flask case since gunicorn points a module, so we'd need to know what module to load.
5. Mounted Middleware -> I think this would be the same as the gunicorn case

richard-to:
Actually mounted Mesop case could be trickier since we wouldn't know what the module for the Mesop app would be since we'd load the FastAPI app via uvicorn or fastapi run.

wwwillchen:
> I took another look in terms of the finding the path to the entry point file
> 
> I guess not surprisingly, we had more entrypoint use cases than I thought:
> 
> 1. Bazel entrypoint -> I think using os.getcwd() is fine. I guess it makes the path to static folder a bit longer.
> 2. Pip install -> For the basic case, can we can pass in the absolute path which we already get from bin.py
> 3. Colab -> I'm not familiar enough with Colab. So not sure what would be the path we'd need here. Would get.getcwd be sufficient here?

I'm not really sure because in Colab there isn't a regular Python module - it's stored as a Drive file. I think using `os.getcwd` would be reasonable, but tbh, I'm not sure how much people would want to use Colab + static assets (it feels like an unusual use case).

> 4. Gunicorn -> Not sure yet how we'd get the path to main.py. I guess in that case may need to use the approaches you pointed out in the Flask case since gunicorn points a module, so we'd need to know what module to load which we can get from the argv and apparently the `module:variable` part should be the last part of the command. Save with uvicorn.

and it's not just gunicorn because they could use other WSGI servers too.

> 5. Mounted Middleware -> I think this would be the same as the gunicorn case

In the mounted app case, agree it's complex because it's not clear whether the mesop module or the fast api app should be considered the root. to be fair, if someone is already doing a mounted app, it's fairly trivial to do your own static folder route.

---

overall, I'm thinking we can get just continue to use os.getcwd() - perhaps printing out the absolute path for the static folder on startup + docs to alleviate this concern.

richard-to:
> overall, I'm thinking we can get just continue to use os.getcwd() - perhaps printing out the absolute path for the static folder on startup + docs to alleviate this concern.

Ok, sounds good. I've already updated the code to print the static folder on start up. And I've updated the docs. Let me know if the docs could be more clear.

docs/api/config.md
```diff
@@ -6,6 +6,52 @@ Mesop is configured at the application level using environment variables.
...
+Here are some examples of valid paths. Let's assume the current working directory is
+`/srv/`
+
+- `static` becomes `/srv/static`
+- `static/` becomes `/srv/static`
+- `static/assets` becomes `/srv/static/assets`
+- `./static` becomes `/srv/static`
+- `./static/` becomes `/srv/static`
+- `./static/assets` becomes `/srv/static/assets`
+- `.` becomes `/srv/static/.`
```
wwwillchen, line 38: I don't understand these last two examples. It's unclear how `.` becomes `/srv/static/.`

Also, I think prohibiting "." and "./" could be good because it seems fairly risky (and probably un-intended) to serve all your Mesop app code (e.g. py modules) - https://github.com/google/mesop/pull/989#discussion_r1812006176

wwwillchen (approved):
LGTM - I think the only remaining comment is the "." and "./"

wwwillchen:
Related to the entry point but not blocking I'm drafting #1057 because I've realized for my use case (e.g. a pip package which uses mesop internally), you can't actually rely on os.getcwd because the user can be in any arbitrary directory when invoking the pip package as a CLI tool. I think this is a pretty unusual use case, but it seems reasonable to allow app developers to explicitly set the app base path in these advanced use cases.

Issue #438, opened 2024-06-15
Class/Tailwind support on Box component


ducan-ne:
**Is your feature request related to a problem? Please describe.**
as mentioned in #123.  As a dev has properly understood the web app. I want to reuse my web skills to mesop instead of writing raw styles.

**Describe the solution you'd like**
Support for passing classes to **Box** would be great.

wwwillchen:
I prototyped this: https://github.com/google/mesop/pull/476, but there's a few hurdles which I've noted in the PR. I'm going to mark this as votes needed and see if more people want this before investing more into this. I know tailwind is very popular in the FE community, but it also seems quite duplicative with Mesop's [Style API](https://google.github.io/mesop/api/style/) and I think _most_ of the use cases that Tailwind does can be covered already (there's some exceptions like state selector, pseudo classes, and groups).

richard-to:
We added a classes field to the box component, so this should allow tailwind support on the box component. Just need to specify add a tailwind generated stylesheet to the page object (for now needs to be hosted on CDN or mounted Fast API server). It seems like it should be able to extract the css classes from python files as well.

PR #995, opened 2024-09-29
Add CSS class support for box component

richard-to:
Ref: https://github.com/google/mesop/issues/438

### DIFF
--- a/mesop/web/src/app/styles.scss
+++ b/mesop/web/src/app/styles.scss
@@ -260,6 +260,12 @@ body {
   }
 }
 
+// The box component has div-like semantics, so make it `display: block` by default.
+// Custom elements like Angular component tags are treated as inline by default.
+component-renderer {
+  display: block;
+}
+
 mesop-markdown {
   h1,
   h2,
--- a/mesop/web/src/app/styles.scss
+++ b/mesop/web/src/app/styles.scss
@@ -261,9 +261,20 @@ body {
 }
 
 // The box component has div-like semantics, so make it `display: block` by default.
-// Custom elements like Angular component tags are treated as inline by default.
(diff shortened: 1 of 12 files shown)


mesop/web/src/app/styles.scss
```diff
@@ -260,6 +260,23 @@ body {
...
+component-renderer {
+  display: block;
+}
+
+// Custom elements like Angular component tags are treated as inline by default.
+//
+// Since component-renderer encompasses many different types of a components, we
+// need to add a more specific selector to make these inline by default.
+component-renderer:not([mesop-box]),
+// The first component-renderer object is a box, but it needs to be inline.
```
wwwillchen, line 275: why does it need to be inline?

wwwillchen (approved):

mesop/web/src/app/styles.scss
richard-to: That's a great question. I wasn't quite able to figure that out. But when I looked at the generated markup, that first component-renderer box element does not have display block by default on main branch (which is definitely strange since in component renderer ts code, in the main branch, it should set the style to display block, at least that's what I would have thought).

richard-to: Yeah I noticed this because some things definitely broke when I changed it to display block. I think the primary issue occurred in the chat apps where the full height stuff wasn't working. So seems like something that could be adjusted.

pritam-dey3:
I would like to have a guide [here](https://google.github.io/mesop/guides/state-management/) with tailwind css example. I just started using mesop today, and feel lucky that this feature just dropped, since I know a little bit of tailwind-css and is much easier for me to arrange the layouts.

PR #999, opened 2024-10-01
Add rxjs dep to date range picker

### LINKED ISSUE #997: Add rxjs dependency to date range picker BUILD file

richard-to:
Rxjs dep is used for the debounce behavior.

Closes #997

### DIFF
--- a/demo/input.py
+++ b/demo/input.py
@@ -23,6 +23,7 @@ def load(e: me.LoadEvent):
   path="/input",
 )
 def app():
-  s = me.state(State)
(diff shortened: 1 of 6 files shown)


wwwillchen (approved):
alternatively, you could also add it to https://github.com/google/mesop/blob/f60e66e37ebd90ed8d78d3a07f3b587ab2226e54/mesop/components/defs.bzl#L32 so then it's automatically included in each component (we use it in a few component and right now we compile everything into 1 bundle so there's no real downside to over-including deps).

PR #1000, opened 2024-10-01
Update grid table example to use dark mode theming

richard-to:


demo/grid_table.py
```diff
@@ -222,14 +225,16 @@ def expander(self, df_row_index: int) -> me.Style:
     return me.Style()
 
 
-class GridTableThemeDark(GridTableTheme):
-  _HEADER_BG: str = "#28313e"
-  _CELL_BG: str = "#141d2c"
-  _CELL_BG_ALT: str = "#02060c"
-  _COLOR: str = "#fff"
+class GridTableThemeDefault(GridTableTheme):
```
wwwillchen, line 228: optional: Having an interface `GridTableTheme` seems a bit of an overkill b/c you only have 1 impl now so you could just inline this into GridTableTheme

wwwillchen (approved):

PR #1004, opened 2024-10-02
Add update page title command

### LINKED ISSUE #331: Dynamically set page title
Currently page title can only be set statically using the `@me.page(path="/foo", title="hi")`, but sometimes users may want to set it dynamically, particularly to show badge/notificated-style information (e.g. 44 unread messages)

richard-to:
This change adds a command that can update the page title.

A few notes:

- Turns out we set the page title after every state change, so a web component wouldn't have worked here
- One issue I ran into design-wise is that we need to keep the state of the updated page title. Ideally, it seems like it sent with the UI state every time. In the end, I chose not to do that. Instead just stored this client-side state as a variable on Channel class for now, which I'm not sure is ideal. But I think can be refactored later if we get more uses cases where we have state that is separate from the main mesop state object. I guess we do have query params and it's possible I could have used a similar but simplified pattern for storing the page title.

Closes #331


docs/api/commands/update-page-title.md
```diff
@@ -0,0 +1,17 @@
+# Update page title
+
+If you want to update the page title, you can use `me.update_page_title` which will
```
wwwillchen, line 3: I would name this `set_page_title` b/c:
1. consistency with our existing APIs, e.g. [set_theme_mode/density](https://github.com/google/mesop/blob/1f34f5669ba2ab064d91a6e11e1da5848bf92077/mesop/features/theme.py#L10)
2. in the case where the user hasn't set a page title (e.g. `me.page(` doesn't have the page arg), I think it's clearer calling this `set_page_title` rather than `update_page_title` since there wasn't really a page title to begin with (besides the default page title the framework sets)

mesop/tests/e2e/update_page_title_test.ts
```diff
@@ -0,0 +1,119 @@
...
+      await page.goto('/update_page_title/page1');
+
+      await page
+        .getByRole('button', {
+          name: 'title change before navigate',
+          exact: true,
+        })
+        .click();
+      await expect(page).toHaveURL(
+        'http://127.0.0.1:32123/update_page_title/page2',
```
wwwillchen, line 32: I think you can drop the origin, see: https://github.com/google/mesop/blob/1f34f5669ba2ab064d91a6e11e1da5848bf92077/mesop/tests/e2e/dom_state_test.ts#L13

mesop/web/src/shell/shell.ts
wwwillchen, line 1: (no action needed) food for thought: it's a little weird to me the interaction between channel and shell (which has been a while for a long time) - they both want to call each other (which would be a cyclic dependency) so we use this indirection.

I think this PR doesn't add that much incremental complexity, but perhaps in the future we should refactor shell and channel. It seems like channel is doing too much (it's both the transport layer + a stateful rendering service) and could be split up into 2+ classes (although the exact split is kind of unclear to me now).

wwwillchen (approved):
LGTM - a comment around the naming, otherwise looks good - thanks!

docs/api/commands/update-page-title.md
richard-to: Great point. Updated.

mesop/web/src/shell/shell.ts
richard-to: Yeah, I agree about this. I'll keep this in mind for the future.

PR #1013, opened 2024-10-03
Fix bug where multiple select values could not be initialized.

### LINKED ISSUE #681: Can't initialize me.select with multiple values
Right now, me.select allows the user set to one variable. But not multiple values which may be needed when multiple selections are allowed.

richard-to:
Drive by adjustments:

- Add appearance property to select
- Update autocomplete/input/textarea/select to use outline appearance.

Closes #681

### DIFF
--- a/mesop/components/select/select.ts
+++ b/mesop/components/select/select.ts
@@ -89,8 +89,8 @@ export class SelectComponent {
 
   private _checkInitialValuesChanged(values: string[]): boolean {
     return (
-      JSON.stringify(this.initialSelectOptions.sort()) ===
-      JSON.stringify(values.sort)
+      JSON.stringify(this.initialSelectOptions.sort()) !==
+      JSON.stringify(values.sort())
     );
   }
 }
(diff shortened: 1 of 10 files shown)


mesop/components/select/select.ts
```diff
@@ -12,24 +12,39 @@ import {
...
       this.type.getValue() as unknown as Uint8Array,
     );
+
+    // In order to preserve backwards compatibility, only update values if the incoming
+    // values change.
+    //
+    // This is to preserve the existing behavior that allows the component to behave
+    // reasonably even if the value property is set/updated.
+    const values = this._config.getValueList();
+    if (this.initialSelectOptions !== values) {
```
wwwillchen, line 44: not sure if this working as intended because in JS `!==` is checking for referential equality and not value equality, i.e.

```
var a = [1]
var b = [1]
a !== b
```

note: this is different than python

wwwillchen (approved):
LGTM % comment on possible bug in equality check

mesop/components/select/select.ts
richard-to: Oh dang. Can't believe I've been using javascript for this long and never knew about this behavior. Updated.

PR #1014, opened 2024-10-03
Button toggle

### LINKED ISSUE #343: Add Button Toggle component
https://material.angular.io/components/button-toggle/overview

richard-to:
Adding the basic button toggle so I can finally replace my custom version lol.

Closes #343


mesop/components/button_toggle/button_toggle.ng.html
```diff
@@ -0,0 +1,18 @@
+<mat-button-toggle-group
+  [style]="getStyle()"
+  [hideMultipleSelectionIndicator]="config().getHideMultipleSelectionIndicator()"
+  [hideSingleSelectionIndicator]="config().getHideSingleSelectionIndicator()"
```
wwwillchen, line 4: i don't understand why Angular Material's API separates this out into two properties vs. just having one like `hideSelectionIndicator` because there's already a separate `multiple` property.

Am I missing something about why these should be separate properties? Should we simplify this in our API?

wwwillchen (approved):
Very nice! 1 comment about the API (I'll defer to you on which way to go)

mesop/components/button_toggle/button_toggle.ng.html
richard-to: Great point. Yeah I don't really know what the use case for it would be. Updated to combine them.

PR #1024, opened 2024-10-11
Fix sidenav "escape" key bug

### LINKED ISSUE #1019: Sidenav on esc close menu
**Describe the bug**
Sidenav is closing when you press ESC key. The state of sidenav_open is not change instead only css class .mat-drawer transform: translate3d(-100%, 0, 0); 

**To Reproduce**
Click on Sidenav then press ESC key

**Expected behavior**
Sidenav should have option to not close sidenav or to change state with ESC.

**Desktop System Info**

- OS: Mac, Ubuntu 24.04
- Browser Chrome, Firefox
- Version latest

richard-to:
Changes:

- Add on opened change event handler (so user can update state) when escape pressed
- Add disable_close property to prevent escape from being used
- Drive bys:
  - Add position property so sidenav can be rendered on the right side
  - Minor style adjustments to sidenav demo

Closes #1019

### DIFF
--- a/mesop/components/sidenav/sidenav.py
+++ b/mesop/components/sidenav/sidenav.py
@@ -1,8 +1,35 @@
+from dataclasses import dataclass
+from typing import Any, Callable, Literal
+
 import mesop.components.sidenav.sidenav_pb2 as sidenav_pb
 from mesop.component_helpers import (
   Style,
   insert_composite_component,
+  register_event_handler,
+  register_event_mapper,
   register_native_component,
+)
+from mesop.events import MesopEvent
+
+
+@dataclass(kw_only=True)
+class SidenavOpenedChangedEvent(MesopEvent):
+  """Event representing the opened state change of the select component.
+
+  Attributes:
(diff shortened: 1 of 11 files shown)


mesop/components/sidenav/sidenav.py
```diff
@@ -1,15 +1,45 @@
...
+    opened=event.bool_value,
+  ),
+)
 
 
 @register_native_component
 def sidenav(
   *,
   opened: bool = True,
+  disable_close: bool = False,
```
wwwillchen, line 40: I'm leaning towards we should default this to True because `on_opened_changed` is an optional param (and we can't make it a required param at this point).

As the original issue stated, it's kinda weird to allow the user to Esc and close the sidenav when there isn't proper state handling.

wwwillchen (approved):
thx! 1 comment on the default value here for sidenav

mesop/components/sidenav/sidenav.py
richard-to: Yeah that's a good point. Updated.

PR #1025, opened 2024-10-11
Turn (server) error box into a more standard dialog UI

wwwillchen:
- Dialog provides a more standard UI compared to the somewhat strange fullscreen error box (it's easier to dismiss it and it's clearer from the overlay that there's, potentially, still UI that you can interact with).
- Stop automatically hiding the error UI if there's no more errors in the next render; instead we will close the previous error dialog if there's a new error. 
  - The problem with the previous approach of automatically hiding is that if you're streaming a response, the first chunk could through an error, but the second chunk might not, which results in a flash of the error UI even though the error is still in the code.


richard-to (approved):
I like this change. This is definitely more intuitive than the previous behavior.

Issue #981, opened 2024-09-21
MESOP_CONCURRENT_UPDATES_ENABLED  - concurrent streaming requests has odd behavior maybe


richard-to:
Need to look into this one more to determine if this is a big issue or not. Could be minor. I'm also not sure the reason for the issue yet.

I was testing out the animation demo. Basically I clicked "Transform Color" followed by "Fade In / Fade Out". I also tested out the reverse.

What happens in this scenario is that the second animation that is clicked will not fully finish. The final rendering will be some intermediate state.

wwwillchen:
Good catch - it looks like what's happening (when I inspect the network tab) is that when there's two concurrent event streams, whichever finishes first, will also close the second one, which makes sense because: https://github.com/google/mesop/blob/129e0d011ebb315d98457ab723c185ee1dab823e/mesop/web/src/services/channel.ts#L118

I'm not sure what's the right way of solving this, we could have a map of eventSource (instead of having a single one, it's actually kind of interesting how the concurrent updates even works with the single eventSource instance) or we can have some kind of counter to keep track of the pending requests and still use a single eventSoruce.

richard-to:
This env var is deprecated, so closing.

Issue #980, opened 2024-09-21
MESOP_CONCURRENT_UPDATES_ENABLED - with non-default state-sessions


richard-to:
Just realized that this could cause some problems with how state session backends currently work. Right now the token IDs change after every request. The old state / token pair also get deleted after every request. After this first concurrent request, the session token will not be found and force a retry.

wwwillchen:
That's a good point, I think we'll need to fix this before making MESOP_CONCURRENT_UPDATES_ENABLED the default, but I imagine this combination will be pretty rare for a while.

richard-to:
Yeah, I've been thinking of how to resolve this. I'm thinking what we can do is stop deleting the state session from the cache right away.

We can probably mark it as "accessed" and then perform a deletion of all "accessed" state sessions at a quicker interval than the stale sessions which we currently delete at a 10 minute interval.

Alternatively, we could simplify it further, and just delete all state at a quicker interval. The issue with this is that we may need to balance inactivity time. However, some slight extra latency for inactive sessions may be ok.

I think that could be enough to make the concurrent updates behave the same as with out sessions.

wwwillchen:
yeah. sounds good closing this

PR #1027, opened 2024-10-13
Add basic websockets support

wwwillchen:
**Summary of changes:**
- Adds basic websocket support on both the server and client side.
- The initial render will use a regular SSE HTTP connection because it needs to fetch the experiment flag. After that, if websockets is enabled (i.e. `MESOP_WEBSOCKETS_ENABLED` env var is set to true), then all the subsequent UI requests and responses will happen on a single websocket connection.
- This partially addresses the use case in #1028. One thing that's missing is that different websocket messages are handled by separate Mesop Context instances, which means that in the concurrent request use case, you can still hit the bug detailed in that issue. I've explored ways of addressing this (e.g. reusing the same Mesop context instance for the same websockect connection), but there are some non-trivial edge cases (e.g. how do we avoid data race conditions in mutating the same component tree) that I've deferred it for a future PR.

**Notes:**
- The security issues found by Snyk for `werkzeug` aren't actually introduced in this PR and they only affect Mesop framework developers.
- Remove unused `_requests_in_flight` code.
- Updated a couple e2e tests that were implicitly relying on the blur getting fired first to explicitly trigger blur. When I tested it manually the blur event almost always happens first.
- I'll try to sync this downstream (looks like socket-io.client is available downstream)


mesop/web/src/services/channel.ts
```diff
@@ -117,143 +139,204 @@ export class Channel {
...
+        this.socket = undefined;
+        console.error('WebSocket error:', error);
+        this.status = ChannelStatus.CLOSED;
+      });
+    });
+
+    this.socket.on('disconnect', (reason: string) => {
+      zone.run(() => {
+        this.socket = undefined;
+        console.error('WebSocket disconnected:', reason);
```
richard-to, line 219: I think fine for now, but could be good to notify the user if the socket gets into a disconnected or error state.

mesop/web/src/services/channel.ts
```diff
@@ -117,143 +139,204 @@ export class Channel {
...
+    });
+
+    this.socket.on('response', (data: any) => {
+      const prefix = 'data: ';
+      const payloadData = (data.data.slice(prefix.length) as string).trimEnd();
+      zone.run(() => {
+        if (payloadData === STREAM_END) {
+          this._isHotReloading = false;
+          this.status = ChannelStatus.CLOSED;
+          this.logger.log({type: 'StreamEnd'});
```
richard-to, line 196: [question] What's the difference between using logger and console?

richard-to (approved):
Awesome. Nice that the the overall amount of non-boilerplate changes was pretty minimal from SSE to WebSocket. 

Have you tested on Cloud Run?

I tested it out a bit, mostly with the animation demo and it worked pretty well. I didn't run into the same issues as #981. 

It still has the issue of #980 but I think that's resolvable by not deleting the sessions objects right away. Not sure if that's a good long term approach. But I can try to prioritize that fix.

Been a while since I've used socket io, but does socket io still have fallbacks when web sockets are not possible --- I'm assuming not since back in the day web socket support was not in all browsers yet?

And thanks for offering to handle the downstream sync. 🙇🏾 🙇🏾 🙇🏾

wwwillchen:
> Awesome. Nice that the the overall amount of non-boilerplate changes was pretty minimal from SSE to WebSocket.
> 
> Have you tested on Cloud Run?
I haven't but I'll try to test it out sometime. Theoretically it should work because Cloud Run supports [websockets](https://cloud.google.com/run/docs/triggering/websockets), but I'm not sure if the cutover will work (e.g. a request gets routed to another Cloud Run instance).
> 
> I tested it out a bit, mostly with the animation demo and it worked pretty well. I didn't run into the same issues as #981.
> 
> It still has the issue of #980 but I think that's resolvable by not deleting the sessions objects right away. Not sure if that's a good long term approach. But I can try to prioritize that fix.
> 
> Been a while since I've used socket io, but does socket io still have fallbacks when web sockets are not possible --- I'm assuming not since back in the day web socket support was not in all browsers yet?

Yeah it still has fallback support with [HTTP long-polling](https://socket.io/docs/v4/how-it-works/#transports). We could support making this configurable, although, I think for Mesop, it's probably better to just use the default SSE approach.
> 
> And thanks for offering to handle the downstream sync. 🙇🏾 🙇🏾 🙇🏾

.github/workflows/ci.yml
```diff
@@ -72,6 +72,14 @@ jobs:
           name: playwright-report-with-concurrent-updates-enabled
           path: playwright-report-with-concurrent-updates-enabled/
           retention-days: 30
+      - name: Run playwright test with websockets enabled
+        run: MESOP_WEBSOCKETES_ENABLED=true PLAYWRIGHT_HTML_OUTPUT_DIR=playwright-report-with-websockets-enabled yarn playwright test
```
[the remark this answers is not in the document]
wwwillchen: good catch! done

mesop/web/src/services/channel.ts
wwwillchen: Good point - filed #1033 - definitely should improve this, perhaps some kind of auto-reconnect 3 times and then show error message.

mesop/web/src/services/channel.ts
wwwillchen: logger is displayed in the Mesop DevTools - although TBH, I never really use it and I just use Chrome DevTools - I'm open to just ripping out logger & maybe the entire DevTools?

richard-to:
> Yeah it still has fallback support with [HTTP long-polling](https://socket.io/docs/v4/how-it-works/#transports). We could support making this configurable, although, I think for Mesop, it's probably better to just use the default SSE approach.

I see. Yeah I was thinking the long polling probably wouldn't be a great experience (compared to SSE). Would there be a way to disable that fallback?

In terms of options from there, I guess we could try to fallback to the Mesop SSE in that scenario. But I do wonder if the behaviors/functionality would diverge too much after a while. So if someone enabled Websockets, then they probably are expecting certain functionality potentially. So the other option is to just sort of fail for those users, but that's also not a great experience (but I imagine web socket support is pretty common nowadays so seems like an uncommon scenario.

wwwillchen:
> > Yeah it still has fallback support with [HTTP long-polling](https://socket.io/docs/v4/how-it-works/#transports). We could support making this configurable, although, I think for Mesop, it's probably better to just use the default SSE approach.
> 
> I see. Yeah I was thinking the long polling probably wouldn't be a great experience (compared to SSE). Would there be a way to disable that fallback?

It should be disabled: https://github.com/google/mesop/blob/75edeb8dbc0f007bcdf9c2421ddff36240627bd0/mesop/web/src/services/channel.ts#L169

> 
> In terms of options from there, I guess we could try to fallback to the Mesop SSE in that scenario. But I do wonder if the behaviors/functionality would diverge too much after a while. So if someone enabled Websockets, then they probably are expecting certain functionality potentially. So the other option is to just sort of fail for those users, but that's also not a great experience (but I imagine web socket support is pretty common nowadays so seems like an uncommon scenario.

Yeah I think failing is OK. My guess is websockets will be used only in advanced use cases, where the developer should have a good idea of whether its supported.

Issue #1028, opened 2024-10-13
Support WebSockets (as an alternative to SSE)


wwwillchen:
# Context

Right now Mesop uses [Server-sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) (SSEs) which have been good, as they are widely supported and relatively straightforward implementation-wise, however there's some limitations with more advanced use cases.

# Motivation

- #623 - even though we have an experimental [MESOP_CONCURRENT_UPDATES_ENABLED](https://google.github.io/mesop/api/config/#mesop_concurrent_updates_enabled) mode, there's quite a few issues with it, e.g. #980 #981. 

There's a key problem here which is that our [core rendering loop](https://github.com/google/mesop/blob/9bbb62e04e2c8a8802fbe862636f3903baa270f3/mesop/server/server.py#L200-L241) in handling long user interactions (via generator) assumes that you can generate tree diffs from the previous tree in your context (i.e. SSE request). 

## Example use case

With concurrent updates, you can run into strange bugs like the following:

- SSE Request #1 (long) - after 4 seconds, it tries to update the tree by adding component A
- SSE request #2 (short) - after 1 second, it tries to update the tree by adding component A

In this case you will end up erroneously adding component A twice because the first SSE request (which takes longer) still sent a component tree diff to add component A even though it was added by another request (that completed sooner). I ran into this issue in my personal app and it's not very intuitive to debug and points to a core issue which is that concurrent updates across multiple stateless server instances is bound to be problematic. Instead, I think having a model where there's multiple concurrent updates getting funneled into the same stateful server instance is more robust.

# Implementation plan

I have a PR #1027 which prototypes support but it's failing many of the test cases. Once I take a closer look and can get most/if not all of them passing, I think we can add a new mode to Mesop, e.g. `MESOP_WEB_SOCKETS_ENABLED`. I think this will always be an opt-in because our current setup (SSE) is quite robust and works for most use cases.

richard-to:
Gonna close this since Web Sockets is implemented, but still in experimental for now.

PR #1036, opened 2024-10-16
Add proper concurrency support in websockets mode

wwwillchen:
This follows up from #1027 and it reuses the same Context instance for an entire session (i.e. websockets connection). This fixes the edge case in #1028 because the component diffing bug is fixed. See new test case which covers this.

I've extracted the environmental variables from `server/server_utils.py` into a separate package `env` to avoid cyclic dependencies since many packages like `runtime` may want to depend on `env`.

### DIFF
--- a/mesop/runtime/runtime.py
+++ b/mesop/runtime/runtime.py
@@ -2,13 +2,14 @@ from copy import deepcopy
 from dataclasses import dataclass
 from typing import Any, Callable, Generator, Type, TypeVar, cast
 
-from flask import g
+from flask import g, request
 
 import mesop.protos.ui_pb2 as pb
 from mesop.events import LoadEvent, MesopEvent
 from mesop.exceptions import MesopDeveloperException, MesopUserException
 from mesop.key import Key
 from mesop.security.security_policy import SecurityPolicy
+from mesop.server.server_utils import MESOP_WEBSOCKETS_ENABLED
 from mesop.utils.backoff import exponential_backoff
 
 from .context import Context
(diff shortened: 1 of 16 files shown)


mesop/runtime/runtime.py
```diff
@@ -54,8 +55,15 @@ def __init__(self):
...
     self._loading_errors: list[pb.ServerError] = []
     self._has_served_traffic = False
+    self._contexts = {}
 
   def context(self) -> Context:
+    if MESOP_WEBSOCKETS_ENABLED and hasattr(request, "sid"):
+      # flask-socketio adds sid (session id) to the request object.
+      sid = request.sid  # type: ignore
+      if sid not in self._contexts:
+        self._contexts[sid] = self.create_context()
```
richard-to, line 65: [question] Is runtime created every request or is it a singleton? If it's the latter, I was wondering if _contexts would need to be cleared

mesop/tests/e2e/concurrent_updates_websockets_test.ts
```diff
@@ -0,0 +1,17 @@
+import {testInWebSocketsEnabledOnly} from './e2e_helpers';
+import {expect} from '@playwright/test';
+
+testInWebSocketsEnabledOnly(
```
richard-to, line 4: Any other test cases we should check that wouldn't be covered by existing tests?

richard-to (approved):
Looks good. Think I only had one question I was unsure of in the code.

PR #1038, opened 2024-10-17
Use OS specific path separator

kumaranvpl:
In the [web_component.py](https://github.com/google/mesop/blob/main/mesop/labs/web_component.py#L34-L35) file, an os specific path is constructed at [line 33](https://github.com/google/mesop/blob/main/mesop/labs/web_component.py#L33) using `os.path.join`. 

But immediately the path is checked whether it starts with "/" or not. "/" is a linux specific path separator and ideally the code should use os specific path separator. 

This PR fixes that issue.

Because of this issue, in windows, full_path `\some\dir\file.js` becomes `/\some\dir\file.js` and later when converted to a URL it becomes `http://host:port//some/dir/file.js` which is not really a valid URL.

### DIFF
--- a/mesop/labs/web_component.py
+++ b/mesop/labs/web_component.py
@@ -31,8 +31,8 @@ def web_component(*, path: str, skip_validation: bool = False):
     os.path.dirname(os.path.abspath(caller_module_file))
   )
   full_path = os.path.normpath(os.path.join(caller_module_dir, path))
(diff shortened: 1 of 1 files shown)


wwwillchen:
thanks!

kumaranvpl:
> thanks!

@wwwillchen No problem. Happy to contribute.

PR #1048, opened 2024-10-23
Switch from socket.io to vanilla websockets & fix state updates for websockets

wwwillchen:
Based on a suggestion from a co-worker who used socket.io quite a bit, it's quite a lot of hassle (e.g. protocol version compatibility) compared to using vanilla websocket since we don't need to use an HTTP fallback in our use case.

Notes:

- Uses the native WebSockets API on the client-side and [flask-sock](https://github.com/miguelgrinberg/flask-sock) which is a modern successor to [flask-socketio](https://github.com/miguelgrinberg/flask-socketio) (note: same author)
- Creates a UUID for the session (i.e. WebSockets lifetime) so that we can reuse the same Context instance across multiple user events (this is the same behavior as introduced in #1036, but using a different mechanism because flask-sock doesn't include `request.sid`)
- Launches a background thread _with_ the Flask request context so that multiple user events can be processed concurrently.
- Disables updating state from the client or sending state to the client because the state is already on the server (since have a long-lived Context object in websockets mode).
- Related to the above point, I've added a warning message in the docs for `MESOP_CONCURRENT_UPDATES_ENABLED` which I should have added earlier. I'm not sure if `MESOP_CONCURRENT_UPDATES_ENABLED` is really viable as a standalone option outside of using it with `MESOP_WEB_SOCKETS_ENABLED`

### DIFF
--- a/build_defs/defaults.bzl
+++ b/build_defs/defaults.bzl
@@ -77,10 +77,6 @@ THIRD_PARTY_JS_HIGHLIGHTJS = [
     "@npm//highlight.js",
 ]
 
-THIRD_PARTY_JS_SOCKETIO_CLIENT = [
-    "@npm//socket.io-client",
-]
-
 THIRD_PARTY_PY_ABSL_PY = [
     requirement("absl-py"),
 ]
@@ -93,8 +89,8 @@ THIRD_PARTY_PY_FLASK = [
     requirement("flask"),
 ]
 
-THIRD_PARTY_PY_FLASK_SOCKETIO = [
-    requirement("flask-socketio"),
+THIRD_PARTY_PY_FLASK_SOCK = [
+    requirement("flask-sock"),
 ]
 
 THIRD_PARTY_PY_MATPLOTLIB = [
--- a/build_defs/requirements.txt
+++ b/build_defs/requirements.txt
@@ -8,7 +8,7 @@ python-dotenv
 
 # Optional (lazily-loaded) deps:
 sqlalchemy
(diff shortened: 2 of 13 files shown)


richard-to (approved):
I think this looks reasonable. I think the part I didn't fully understand was the use of the while True loop and the threading.

wwwillchen:
> I think this looks reasonable. I think the part I didn't fully understand was the use of the while True loop and the threading.

the while True loop basically means a thread is allocated per websocket connection and watching for any incoming messages, which isn't super scalable (compared to a proper async approach), but I think good enough for now.

re: the threading, it's to allow multiple user events for the same websocket connection to be processed concurrently, otherwise they would end up getting processed serially.

The concurrency still here is a bit tricky and I might try to simplify/change this in the future.

richard-to:
> re: the threading, it's to allow multiple user events for the same websocket connection to be processed concurrently, otherwise they would end up getting processed serially.

Yes, I understood that part. I guess I was wondering about thread safety. I remember we do have the lock for when the context is updated. So I assume the rest is fine. The other thing I was wondering about was the GIL. Are there scenarios where the threading may not actually work that well (there was one time where I was using threads and it ended up using a ton of compute and being way slower than just single threaded -- but I think I just picked a poor use case since it was more processing intensive versus I/O intensive)? I think for a long LLM call that's hitting the network, I'm assuming that should work. But I'll try to play around with this version of the web sockets implementation a bit more when I get the chance.

PR #1057, opened 2024-10-26
Add env var: MESOP_APP_BASE_PATH 

wwwillchen:
This adds MESOP_APP_BASE_PATH as a new configuration via env var to support the use case where `os.getcwd()` cannot be reliably used as the app base path. Specifically, if you want to build a pip package CLI app that runs Mesop, using the cwd isn't feasible because the app user can be running the CLI command from an arbitrary directory (and not the mesop app directory). I suspect this is a pretty niche use case so I don't think most people will need to use this.

Note: the testing for this isn't that great (it's not trivial to cover the pip CLI use case with our current integration test infra), but it covers the static folder case.

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -252,9 +252,9 @@ By default, this is not enabled. You can enable this by setting it to `true`.
 
 This uses WebSockets instead of HTTP Server-Sent Events (SSE) as the transport protocol for UI updates. If you set this environment variable to `true`, then [`MESOP_CONCURRENT_UPDATES_ENABLED`](#MESOP_CONCURRENT_UPDATES_ENABLED) will automatically be enabled as well.
 
-### MESOP_STATIC_FILES_BASE_PATH
+### MESOP_APP_BASE_PATH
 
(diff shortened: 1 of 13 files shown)


docs/api/config.md
```diff
@@ -252,6 +252,10 @@ By default, this is not enabled. You can enable this by setting it to `true`.
 
 This uses WebSockets instead of HTTP Server-Sent Events (SSE) as the transport protocol for UI updates. If you set this environment variable to `true`, then [`MESOP_CONCURRENT_UPDATES_ENABLED`](#MESOP_CONCURRENT_UPDATES_ENABLED) will automatically be enabled as well.
 
+### MESOP_APP_BASE_PATH
+
+This is the base path used to resolve other paths, particularly for serving static files. This is rarely needed because the default of using the current working directory is usually sufficient.
```
richard-to, line 257: Maybe mention that this needs to be an absolute path. I think sort of implied because it is a base path.

richard-to (approved):

docs/api/config.md
wwwillchen: good idea - done

Issue #903, opened 2024-08-30
Using absl logging library (where applicable)


wwwillchen:
For things that don't  have any terminal formatting and may be useful when running in a production setting, we should use logging library to give app developers more configurability.

PR #1062, opened 2024-10-27
Allow custom favicons

### LINKED ISSUE #563: Add ability to add custom favicon

richard-to:
If static folder is enabled, then we will check if there is a favicon.ico in the designated static folder. If it exists, we will use the user's favicon instead of the default.

Also:

- Moved the static folder functions to server_utils.py
- Improved validation slightly for static URL paths
- Updated docs with custom favicon example

Closes #563

For a future improvement, we can add more complete favicon support. See https://www.emergeagency.com/insights/detail/the-essentials-of-favicons/


wwwillchen (approved):
awesome

mesop/server/server.py
```diff
@@ -44,10 +43,17 @@
 def configure_flask_app(
   *, prod_mode: bool = True, exceptions_to_propagate: Sequence[type] = ()
 ) -> Flask:
+  static_folder = get_static_folder()
+  static_url_path = get_static_url_path()
+  if static_folder and static_url_path:
+    print(f"Static folder enabled: {static_folder}")
+  else:
```
wwwillchen, line 50: imo, I think it's a little spammy to print this out (in the disabled case). we should eventually move most/if not all of our print statements into proper logger statements #903, but for now we should avoid adding prints in the base/default cause.

PR #1065, opened 2024-10-27
Fix race condition in processing commands (navigate)

wwwillchen:
Previously we had an edge case where a navigate & query param mutation (see newly added e2e test case) ran into a race condition because we weren't properly awaiting the Angular router navigation.

This ensures that all commands (including across render events) are processed serially.

### DIFF
--- a/mesop/web/src/services/channel.ts
+++ b/mesop/web/src/services/channel.ts
@@ -36,7 +36,7 @@ interface InitParams {
     jsModules: readonly string[],
   ) => void;
   onError: (error: ServerError) => void;
-  onCommand: (command: Command) => void;
+  onCommand: (command: Command) => Promise<void>;
 }
 
 export enum ChannelStatus {
@@ -237,7 +237,7 @@ export class Channel {
     };
   }
 
-  private handleUiResponse(
+  private async handleUiResponse(
     request: UiRequest,
     uiResponse: UiResponse,
     initParams: InitParams,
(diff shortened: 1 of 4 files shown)


mesop/web/src/services/channel.ts
```diff
@@ -354,6 +355,27 @@ export class Channel {
     }
   }
 
+  private async processCommandQueue(
+    onCommand: (command: Command) => Promise<void>,
+  ) {
+    if (this.commandQueuePromise) {
```
richard-to, line 361: This part was slightly unclear to me. Is this conditional block for the websocket case where we can have concurrent requests and there for concurrent commands? 

Curious what would happen in terms of the order of operations? Would this end up reducing concurrency since it serializes the behavior?

I think the commands we have are relatively fast operations anyway.

So the only question I guess is the navigate behavior. In this case, seems like first command to reach navigate would win?

So if I click on some long running request that eventually navigates and a quick event that navigates, then the second navigation would occur. But for two quick events, I guess the first navigation would usually occur and unless the first event was slightly slower.

I think that seems fine. Mainly just trying to understand if I'm on the right track with my interpretation of the code.

richard-to (approved):

mesop/web/src/services/channel.ts
wwwillchen: actually, this wasn't even for concurrent requests/websockets. I hit an edge case where if you have two render events (e.g. two yields in an on_load block) that happen quickly together, you can run into a race because the first render event is still processing the command (in particular, the angular navigate which is async) and the second render event comes in. This basically serializes all the commands across render events, which I think would reduce concurrency / negatively affect performance, but only in a minor way since our commands, as you point out, are quite fast.

In terms of ordering, whichever render response is received first would be run first, but if there's two navigations, as in your example, then both navigations would get applied and effectively it would be last writer wins (e.g. the last navigation is the final destination). Let me know if that makes sense.

PR #1066, opened 2024-10-27
Fix component renderer: querying web component children

wwwillchen:


mesop/web/src/component_renderer/component_renderer.ts
```diff
@@ -221,8 +221,9 @@ export class ComponentRenderer {
   }
 
   private updateCustomElementChildren() {
-    const existingChildren = Array.from(
-      this.customElement!.querySelectorAll(COMPONENT_RENDERER_ELEMENT_NAME),
+    const existingChildren = Array.from(this.customElement!.children).filter(
+      (child) =>
+        child.tagName === COMPONENT_RENDERER_ELEMENT_NAME.toUpperCase(),
```
richard-to, line 226: Maybe add a comment here on why we need to do toUpperCase here. It's not immediately clear to me what's going on. I see the constant is `component-renderer-element` and when we create the element we use `component-renderer-element`. So why does the tag name change to `COMPONENT-RENDERER-ELEMENT`?

Uh, ok, I looked it up. I see. So tagName returns all uppercase so it can be case-insensitive.

richard-to (approved):
Interesting. Surprised, we haven't run into this issue earlier.

mesop/web/src/component_renderer/component_renderer.ts
wwwillchen: yup - added a comment

wwwillchen:
> Interesting. Surprised, we haven't run into this issue earlier.

yeah, i'm a little surprised too, but it's a somewhat advanced use case where you need to have nested web components with a dynamic number of children to hit it.

PR #1072, opened 2024-10-28
Fix bug with single select value not showing.

### LINKED ISSUE #1071: Select button single select not displaying selected option
Looks like a regression. 

Possibly:

- https://github.com/google/mesop/pull/1037
- https://github.com/google/mesop/pull/1013

richard-to:
Seems like in single select mode, the ngModel is a string rather than an array.

So we need to update the code to handle both cases.

Closes #1071

### DIFF
--- a/demo/select_demo.py
+++ b/demo/select_demo.py
@@ -5,14 +5,20 @@ import mesop as me
 
 @me.stateclass
 class State:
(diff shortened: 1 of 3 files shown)


wwwillchen (approved):
lgtm

PR #1074, opened 2024-10-29
Add content_button equivalent to uploader

### LINKED ISSUE #1068: Add content_button version of uploader
Main use case is to allow the option to use an icon for the uploader. But I think there could be other benefits to providing more flexibility to the uploader stying.

richard-to:
This change adds content_uploader which works similar to content_button. This change allows a little bit more freedom to style the uploader, especially for using icons which is a common use case.

I wasn't too sure about the naming. "content_uploader" could be misconstrued slightly, but it parallels content_button.
Closes #1068

### DIFF
--- a/mesop/components/uploader/e2e/uploader_app.py
+++ b/mesop/components/uploader/e2e/uploader_app.py
@@ -13,14 +13,16 @@ class State:
 def app():
   state = me.state(State)
   with me.box(style=me.Style(padding=me.Padding.all(15))):
-    me.uploader(
-      label="Upload Image",
+    with me.content_uploader(
       accepted_file_types=["image/jpeg", "image/png"],
       on_upload=handle_upload,
       type="flat",
       color="primary",
       style=me.Style(font_weight="bold"),
-    )
+    ):
+      with me.box(style=me.Style(display="flex", gap=5)):
+        me.icon("upload")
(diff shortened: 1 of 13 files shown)


mesop/components/uploader/uploader.py
```diff
@@ -44,24 +46,64 @@ def map_upload_event(event, key):
...
-@register_native_component
+@component
 def uploader(
   *,
   label: str,
   accepted_file_types: Sequence[str] | None = None,
   key: str | None = None,
   on_upload: Callable[[UploadEvent], Any] | None = None,
-  type: Literal["raised", "flat", "stroked"] | None = None,
+  type: Literal["raised", "flat", "stroked", "icon"] | None = None,
```
wwwillchen, line 56: I think you're trying to have parity between uploader and content_uploader API but it doesn't really make sense to have icon as a type for uploader because you can't configure an icon here

mesop/components/uploader/e2e/uploader_app.py
wwwillchen, line 1: Just to make sure that regular "uploader" works, I'd leave this file alone (and the test) and then just copy this app file and the test, this way you cover both uploader and content_uploader (or you can have them both in one file)

wwwillchen (approved):

PR #1076, opened 2024-10-29
Move copy_current_request_context to ensure thread-safety

wwwillchen:
This should fix the error I was getting in websockets mode:

```
ValueError: <Token var=<ContextVar name='flask.request_ctx' at 0x102d56c50> at 0x1138dbac0> was created in a different Context
```

See: https://github.com/pallets/flask/discussions/5254


richard-to (approved):
🙏🏾

PR #1082, opened 2024-11-01
Add Angular Material Card component

### LINKED ISSUE #1080: Add Angular Card Component
Ref: https://material.angular.io/components/card/examples

Simple component to do, but a nice way for people with less HTML/CSS knowledge to add a common component with a consistent styling and behavior.

richard-to:
This component consists of the following components:

- card
- card_header
- card_content
- card_actions

I did not include card_footer since I couldn't tell what benefit it added, but we can added later if needed.

I did not include an card_image component since it was easy enough to use image with width 100%.

Closes #1080

### DIFF
--- a/demo/card.py
+++ b/demo/card.py
@@ -1,0 +1,103 @@
+import mesop as me
+
+
+def load(e: me.LoadEvent):
+  me.set_theme_mode("system")
+
+
+@me.page(
+  on_load=load,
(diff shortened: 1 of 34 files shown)


wwwillchen (approved):
nice

wwwillchen:
btw, looks like one of the card test might be flaky

PR #1093, opened 2024-11-05
Send experiment settings on initial page load

wwwillchen:
It's better to send the experiment settings on initial page load rather than getting them from the first render, particularly for the websockets configuration because it can result in a weird edge case where the first UI request is over SSE (HTTP) and all the subsequent requests is over WebSockets. Because Mesop has very different behavior under WebSockets (e.g. state is kept in-memory and never sent over the network) compared to SSE, mixing SSE and WebSockets over a single browser session can result in weird edge cases.


richard-to (approved):
Nice. I agree this makes more sense.

PR #1097, opened 2024-11-08
Fix clearing query params when navigating

wwwillchen:
The previous implementation had a flaw where it issued a command to clear the query params which resulted in replacing the browser history entry, which means that back navigations would not restore the original query params. (The e2e test case demonstrates this issue).

### DIFF
--- a/mesop/commands/navigate.py
+++ b/mesop/commands/navigate.py
@@ -3,9 +3,6 @@ from typing import Sequence
 from mesop.features.query_params import (
   QueryParams,
 )
-from mesop.features.query_params import (
-  query_params as me_query_params,
-)
 from mesop.runtime import runtime
 from mesop.utils.url_utils import remove_url_query_param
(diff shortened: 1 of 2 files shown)


richard-to (approved):
Why does using the query params in the runtime context preserve the query params when going back?

wwwillchen:
> Why does using the query params in the runtime context preserve the query params when going back?

The problem was that previously we were doing me.query_params.clear() which issued a clear query param command which modified the browser history. By directly deleting the query params from context, we avoid issuing this command, and thus not modifying the query param in the browser history.

richard-to:
> The problem was that previously we were doing me.query_params.clear() which issued a clear query param command which modified the browser history. By directly deleting the query params from context, we avoid issuing this command, and thus not modifying the query param in the browser history.

Ah thanks for the explanation. That makes sense now.

PR #1109, opened 2024-11-19
First pass at implementing cors

### LINKED ISSUE #1110: CORS support
Applications built with Mesop should be embeddable into pages on external sites.  AI (especially chat) UIs are prime candidates for embedding into host websites.

To enable this, [CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) support will need to be added.

CORS can be added to Mesop relatively simply by copying the pattern used for [Iframe Security](https://google.github.io/mesop/guides/web-security/#iframe-security).  Iframe Security (SecurityPolicy) and CORS have similar use cases and implementations. Both of these mechanisms are implemented using HTTP response headers, and configuration via `@me.page` parameters make sense as a configuration mechanism.

I have a preliminary PR here: https://github.com/google/mesop/pull/1109

czcollier:
fixes #1110

### DIFF
--- a/mesop/runtime/runtime.py
+++ b/mesop/runtime/runtime.py
@@ -10,6 +10,7 @@ from mesop.events import LoadEvent, MesopEvent
 from mesop.exceptions import MesopDeveloperException, MesopUserException
 from mesop.key import Key
 from mesop.security.security_policy import SecurityPolicy
+from mesop.security.cors import CORS
 from mesop.utils.backoff import exponential_backoff
 from mesop.warn import warn
 
@@ -33,6 +34,7 @@ class PageConfig:
   title: str
   stylesheets: list[str]
   security_policy: SecurityPolicy
+  cors: CORS
   on_load: OnLoadHandler | None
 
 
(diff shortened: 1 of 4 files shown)


wwwillchen:
Hey - thanks for opening this PR. I left a comment on the issue: https://github.com/google/mesop/issues/1110#issuecomment-2487056529

I don't think I fully understand the requested use case, but it's not clear to me how setting `Access-Control-Allow-Origin` on 
index.html (which is effectively what this PR is doing). It seems pretty unusual to want to do a cross-origin XHR/fetch for the index.html, particularly for a Mesop app.

mesop/runtime/runtime.py
```diff
@@ -33,6 +34,7 @@ class PageConfig:
   title: str
   stylesheets: list[str]
   security_policy: SecurityPolicy
+  cors: CORS
```
wwwillchen, line 37: Instead of a separate parameter in PageConfig, I would put CORS into SecurityPolicy (i.e. `allowed_cors_origins`) since CORS is security-related

PR #1123, opened 2024-12-03
Skip unnecessary trace render loop in WebSockets mode

wwwillchen:
This skips the trace mode render loop in WebSockets mode which is a significant perf improvement, since we can use the last render loop because the Context is long-lived across multiple UI requests (for the same session).

This always registers the event handler regardless of whether we're in trace mode. For non-WebSockets mode, there's no practical difference.


richard-to (approved):
Nice. Great optimization. Makes sense to always register the event handlers whether in trace mode or not. Makes the code a bit easier to read there too.

PR #1136, opened 2024-12-06
Fixes major memory leak in component renderer

wwwillchen:
If you navigate to the same pages over and over again, you can measure an significant increase in JS memory (heap snapshots). The main contributor was that we weren't cleaning up the embedded view refs.

### DIFF
--- a/mesop/version.py
+++ b/mesop/version.py
@@ -1,6 +1,6 @@
 """Contains the version string."""
 
-VERSION = "0.12.9"
+VERSION = "0.12.10beta5"
 
 if __name__ == "__main__":
   print(VERSION)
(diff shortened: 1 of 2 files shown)


mesop/version.py
```diff
@@ -1,6 +1,6 @@
 """Contains the version string."""
 
-VERSION = "0.12.9"
+VERSION = "0.12.10beta5"
```
richard-to, line 3: Is this change necessary?

richard-to (approved):
LGTM - Great find.

mesop/version.py
wwwillchen: let me take it out. I'll cut a new release soon

Issue #929, opened 2024-09-06
me.slot appending in nested content_components


ruttenm:
I tried to nest two content_components, where the me.slot() is not at the end of the component. This leads to unexpected behavior in the layout where it seems the content of both me.slots() is combined in the last me.slot().

```
import mesop as me


@me.page(path="/")
def homepage():
    with layout():
        me.text("hello world")


@me.content_component
def layout():
    with me.box(
        style=me.Style(
            display="flex",
            flex_direction="column",
            height="100vh",
            background="red",
        )
    ):

        with me.box(
            style=me.Style(
                flex="1",
                background="blue",
            )
        ):
            me.slot()

        with nav_bar():
            me.text("navbar")


@me.content_component
def nav_bar():
    me.slot()
```

**To Reproduce**

Run `mesop test.py`

**Expected behavior**
The expected layout is to see "Hello world" in the blue box in the main content, and "navbar" in red at the bottom. However, when using `with me.navbar():` the "Hello world" is rendered in the red navbar at the bottom.

**Desktop System Info**

- OS: MacOS
- Browser: Chrome
- Version: mesop 0.12.3

RodrigoDiasDeOliveira:
HI whats up?! i hope i can help...

The issue you are facing is likely related to the way the `me.slot()` function is being handled within nested content components. When `me.slot()` is used in nested components, especially when the slots are not at the end of the component, it can cause unexpected behavior in rendering. Specifically, the content for the slots might be combined incorrectly due to how the framework interprets and assigns content to the slots.

Here's an analysis of the problem and a potential solution:

 I Think:
In the `layout` component, you are using `me.slot()` within the first `me.box`, and then you have another content component (`nav_bar`) that also contains a `me.slot()`. When the framework processes the slots, it might combine the content meant for the first `me.slot()` with the content for the second `me.slot()`, leading to incorrect rendering.


To fix this issue, you can try explicitly specifying different slot names using the `name` parameter for each `me.slot()`. This way, you can ensure that the content is correctly assigned to the appropriate slot.

Here is the modified code:

```python
import mesop as me

@me.page(path="/")
def homepage():
    with layout():
        me.text("hello world")

@me.content_component
def layout():
    with me.box(
        style=me.Style(
            display="flex",
            flex_direction="column",
            height="100vh",
            background="red",
        )
    ):
        with me.box(
            style=me.Style(
                flex="1",
                background="blue",
            )
        ):
            me.slot(name="main_content")  # Use a named slot

[...]

Issue #810, opened 2024-08-16
Support multiple slots


malmaud:
**Is your feature request related to a problem? Please describe.**
I want to define content components that accept *multiple* arbitrary components. eg:

* if i wanted to define a tab view where the user can pass in multiple arbitrary components for the different tabs
* or a 'content with header' component where the content and header are each arbitrary user-supplied components

**Describe the solution you'd like**
Some way to pass arbitrary components as children or parameters to content components, not sure what the API should look like. Maybe something like

```python
with header_and_body_component() as parent:
   with parent.header():
       me.text("my header")
   with parent.body():
       me.text("my body")
```

PR #1145, opened 2024-12-10
Named slots / Nested slots

richard-to:
# Changes

- Adds support for named slots
- Mostly fixes issues with custom components inside custom components
  - Still has the limitation where you can't defer slot rendering to a higher level component
- Update docs
- 
# Issues

- Closes https://github.com/google/mesop/issues/810
- Closes https://github.com/google/mesop/issues/929

### DIFF
--- a/docs/components/index.md
+++ b/docs/components/index.md
@@ -48,7 +48,7 @@ def scaffold(url: str):
       me.slot()
 ```
 
-Now other components can re-use this scaffold component:
+Now other components can re-use this `scaffold` component:
 
 ```python
 def page1():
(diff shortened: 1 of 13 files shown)


wwwillchen:
Overall, I'm liking the direction of this. Thanks for working on this!

My main two suggestions are:

1. I don't think we should support multiple unnamed slots. It seems: a) no real benefit compared to multiple named slots, b) harder to read the callsites (e.g. what's the difference between the first slot and second slot) and c) deviates from HTML slots: https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_templates_and_slots - based on my reading, in HTML, you can only have one unnamed slot, which acts like a default slot, and it takes in all the content implicitly as the slot (i.e. what our content_component currently does). I think keeping this similar to web API isn't totally necessary, but it's a good sanity check in terms of our API design. Interestingly, Angular itself models their concept (content projection) after the HTML standard, but with some differences to "angularize" it: https://angular.dev/guide/components/content-projection/

2. This isn't 100% necessary for the feature, but ideally I'd like to have some form of stricter typing. For example, an ideal API for this would look something like this:

```py
@me.content_component(named_slots=["header", "body"])
def comp():
  ...

def page():
  with comp() as c:
     with c.header():
       me.text("...")
     with c.body():
       me.text("...")
```

I'm not totally sure how to write this type annotation / implement this (maybe Gemini can help :), but I think this would provide a much nicer API for consumers for named slot content components because then you can get IDE support (e.g. autocomplete, type-checking)

richard-to:
I think it's definitely doable to get this syntax, but does not seem like we'd be able to get the dynamic annotations/type hints in IDE in a nicer way.

```
def page():
  with comp() as c:
     with c.header():
       me.text("...")
     with c.body():
       me.text("...")
```

There is one option which is to use namedtuple

```
@me.content_component(named_slots=namedtuple('SomeName', ["header", "body"]))
def comp():
```

I think this would work, but I think doesn't feel like a good trade off. Since we'd need to create a different namedtuple for each set of named slots. Also wouldn't work if someone did: `namedtuple('SomeName', slot_names_from_var)`

Another option is for users to implement a dataclass and pass that into the named slots. But seems unnecessarily verbose and probably don't want to expose DetachedNodeTreeStateContext

```
@dataclass
class HeaderBodySlots:
  header: DetachedNodeTreeStateContext
  body: DetachedNodeTreeStateContext
```

richard-to:
Hmm, one more idea that could work is to let the user pass in a Literal for the named slots like `Literal['header', 'body']`

```
@me.content_component(named_slots=Literal['header', 'body']))
def comp():
```

Drawback is that this would lead to an api like this:

```
def page():
  with comp() as c:
     with c.slot('header'):
       me.text("...")
     with c.slot('body'):
       me.text("...")
```

So not as a nice c.header, c.body

wwwillchen:
> EDIT: Actually NVM, don't think that would work.
> 
> 
> 
> Hmm, one more idea that could work is to let the user pass in a Literal for the named slots like `Literal['header', 'body']`
> 
> 
> 
> ```
> 
> @me.content_component(named_slots=Literal['header', 'body']))
> 
> def comp():
> 
> ```
> 
> 
> 
> Drawback is that this would lead to an api like this:
> 
> 
> 
> ```
> 
> def page():
> 
>   with comp() as c:
> 
>      with c.slot('header'):
> 
>        me.text("...")
> 
>      with c.slot('body'):
> 
>        me.text("...")
> 
> ```
> 
> 
> 
> So not as a nice c.header, c.body

I think the named slots w/ literals is a good idea. Not quite as ergonomic, but still readable and easy both for the content component producer and consumer.

richard-to:
> I think the named slots w/ literals is a good idea. Not quite as ergonomic, but still readable and easy both for the content component producer and consumer.

Unfortunately, I spoke too soon about using Literals. Don't think it will work. Was hoping something like this could work. But get `(Unknown) -> Any). Passing in a dynamic literal does work when not wrapped in the Callable. But unfortunately not too useful that way. So I'll just go ahead and implement the c.header(), c.footer()

```
from typing import Literal, Callable, NamedTuple, Any


def make_tuple(typeX):
  return NamedTuple("TestT", [("slot", Callable[[typeX], Any])])

TestT = make_tuple(Literal["header", "footer"])
 
t = TestT(slot=lambda a: a)
t.slot()
```

wwwillchen:
I'd avoid namedtuple because it's not great API-wise for long-term maintenance compared to dataclass, see: https://snarky.ca/dont-use-named-tuples-in-new-apis/

Similar to how we do state class, what if we did this?

```py
@me.slotclass
class ScaffoldSlots:
  header: me.NamedSlot
  body: me.NamedSlot
  footer: me.NamedSlot

@me.content_component(named_slots=ScaffoldSlots)
def ...
```

Essentially slotclass is an alias for dataclass, but we could do things like enforce [kw_only](https://docs.python.org/3/library/dataclasses.html#:~:text=in%20version%203.10.-,kw_only,-%3A%20If%20true%20(the).
me.NamedSlot is basically an alias to something like DetachedNodeTreeStateContext

Even though this is a bit more verbose than namedtuple, this should be pretty familiar to Mesop developers since it's quite similar to the stateclass syntax and it avoids the downsides of namedtuples (e.g. indexing).

wwwillchen (approved):
lgtm - nice!

PR #1146, opened 2024-12-12
Fix missing type hints for @content_component

### LINKED ISSUE #912: No typing for functions wrapped by @me.component and @me.content_component
Minor thing but typing of the underlying function does not show up when using the component decorators. Could be a small improvement to dev ex if we could get typing to show up.

richard-to:
I checked @component again and that seems to work. It was only @content_component which was not showing the type hints.

Not sure why removing the annotation made it so thecustom component's annotations showed up.

# Before

Screenshot 2024-12-11 at 6 04 35 PM

# After

Screenshot 2024-12-11 at 6 04 55 PM

Closes #912

### DIFF
--- a/mesop/component_helpers/helper.py
+++ b/mesop/component_helpers/helper.py
@@ -8,7 +8,9 @@ from typing import (
   Any,
   Callable,
   Generator,
+  Generic,
   KeysView,
+  ParamSpec,
   Type,
   TypeVar,
   cast,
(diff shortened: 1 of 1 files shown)


richard-to:
Ugh...

wwwillchen:
I don't think removing the type annotations is the right solution. Take a look at how @component is implemented. I think we need TypeVar and use it as a param and return value.

richard-to:
> I don't think removing the type annotations is the right solution. Take a look at how https://github.com/component is implemented. I think we need TypeVar and use it as a param and return value.

Yeah looks that way.  I find those annotations a bit hard to read. But yeah updated it with the TypeVar stuff. Claude (er...Gemini) explained the usage well enough.

wwwillchen (approved):

PR #1155, opened 2024-12-18
Bump angular to v19

wwwillchen:
note: this removes some unused imports/dependencies b/c Angular now has stricter checking in v19


packages.bzl
```diff
@@ -17,6 +17,7 @@ _CDK_ENTRY_POINTS = [
     "layout",
     "bidi",
     "portal",
+    "private",
```
richard-to, line 20: What's private for?

mesop/tests/e2e/focus_component_test.ts
```diff
@@ -51,7 +51,7 @@ test('focus select', async ({page}) => {
   await page.goto('/focus_component');
   await page.locator('//mat-select[@id="mat-select-0"]').click();
   await page.getByRole('option', {name: 'Select'}).click();
-  await expectIsFocused(page, 'mat-select[id="mat-select-2"]');
+  await expectIsFocused(page, 'mat-select[id="mat-select-1"]');
```
richard-to, line 54: Why did this change to 1?

mesop/web/src/component_renderer/component_renderer.ts
```diff
@@ -39,7 +38,7 @@ const WEB_COMPONENT_PREFIX = '<web>';
   templateUrl: 'component_renderer.ng.html',
   styleUrl: 'component_renderer.css',
   standalone: true,
-  imports: [CommonModule, ComponentLoader],
+  imports: [CommonModule],
```
richard-to, line 41: When was Component Loader used before? Thinking about it now, I don't actually remember looking at the file before.

richard-to (approved):
LGTM - I assume the CI tests will pass once you rebase #1163. Mainly just added a few minor clarification comments to help with my understanding.

mesop/tests/e2e/focus_component_test.ts
wwwillchen: I don't know. It seems weird it _used_ to be mat-select-2 because there's only two selects on the page (note: it's 0-indexed) so mat-select-1 seems right to me.

packages.bzl
wwwillchen: looks like CDK has a new private package: https://github.com/angular/components/tree/main/src/cdk/private

mesop/web/src/component_renderer/component_renderer.ts
wwwillchen: It hasn't been used in a very long time. At one point, I was deferring loading the components (e.g. lazy load only the components that you need) so I wrote ComponentLoader, but I couldn't actually get this working inside google3 because it compiles down into a dynamic import (e.g. `import(...)`) which has special handling in google3

PR #1181, opened 2025-01-14
Make sure .env is loaded before env var are set

### LINKED ISSUE #1180: Env vars from .env file do not seem to be loading
Not sure when it happened, but my .env file seems like it's being ignored. Happening for me both on bazel and from pip mesop.

richard-to:
Looks like the issue is that env.py gets loaded before config.py since the latter contains the code to load the .env. We may want to combine env.py and config.py. But for now, just fixing the bug.

So basically the .env file was being loaded correctly. It just wasn't handling the env vars in env.py correctly which was why I could never enable web sockets. 

Closes #1180

### DIFF
--- a/mesop/env/env.py
+++ b/mesop/env/env.py
@@ -1,6 +1,17 @@
 import os
 
+from dotenv import find_dotenv, load_dotenv
+
 from mesop.exceptions import MesopDeveloperException
+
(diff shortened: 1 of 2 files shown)


wwwillchen (approved):
yeah, agree that merging the files makes sense

PR #1187, opened 2025-01-29
Add Cloud Run tutorial

richard-to:
This change adds a page with the Cloud Run video tutorials


wwwillchen:
Thanks. I would also showcase this on our home page too!

wwwillchen (approved):

richard-to:
Sure, good idea. I can add it to the home page also. I'll add it to the community section.

Issue #1185, opened 2025-01-22
Namedslots support for web_components


richard-to:
### Discussed in https://github.com/google/mesop/discussions/1178

<div type='discussions-op-text'>

<sup>Originally posted by **jmfk** January  7, 2025</sup>
I noticed namedslots isn't implemented for web_components. It is definitely on my wish list for the future. 

Since sending UI events via the server just isn't viable implementing things in web_components seems like the best solution.
But just having a single slot is a bit limiting when building UI components. I'm currently building a resizable layout component, and I want to pass sidebar, content and other pane content to it. The component it self would only handle the resize events and render the other components.

</div>

wwwillchen:
I started prototyping this:
https://github.com/google/mesop/pull/1226

Besides the implementation being pretty non-trivial, there's also an open question of how it will actually work. For example:

```py
with multi_slot_web_component() as c:
  with c.header():
     me.text("header")
  with c.footer():
     me.text("footer-child1")
     me.text("footer-child2")
```

Because web component slots must have exactly one element (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_templates_and_slots#adding_flexibility_with_slots)), this means that we'd need to have some kind of element (probably a div) to wrap the two children components (footer-child1, footer-child2). I don't like creating wrapper elements because if someone is doing flexbox styling, then adding an intermediate DOM element disrupts it (and because this is a "synthetic" component, the app developer can't really customize it either).

So, then we'd need enforce that for named slots (at least for >1) web components, it must have a single child, which feels a little weird to me since we don't enforce this anywhere else in Mesop.

I think we can punt on this for a while because you can technically get away with one slot and just manipulate with the children in that one slot (it's hacky, but possible):

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <title>Footer-Header Component</title>
  </head>
  <body>
    <!-- Example usage:
         The first child will be rendered as the footer,
         and the second child as the header. -->
    <footer-header>
      <div>First Child: Will be rendered as the footer</div>
      <div>Second Child: Will be rendered as the header</div>
    </footer-header>

    <script>
      class FooterHeader extends HTMLElement {
        constructor() {
          super();
          // Attach a shadow root.
          this.attachShadow({ mode: "open" });
          
          // Create header and footer containers.
          this._header = document.createElement("header");
          this._footer = document.createElement("footer");
          
          // Create a single slot for light DOM content.
[...]

PR #1191, opened 2025-02-02
Add example of using Gemini Live API with Mesop.

richard-to:
Demonstrates the following:

- Setting up web socket connection to Gemini Live API via client
- Playing audio responses from Gemini Live API
- Sending audio input via user's microphone to Gemini Live API
- Sending video frames via user's webcam to Gemini Live API
- Using custom tool to interact with Mesop UI

### DIFF
--- a/mesop/examples/web_component/gemini_live/gemini_live.js
+++ b/mesop/examples/web_component/gemini_live/gemini_live.js
@@ -1,0 +1,178 @@
+import {
+  LitElement,
+  html,
+} from 'https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js';
+import {GeminiLiveApi} from './gemini_live_api.js';
+
+class GeminiLive extends LitElement {
+  static properties = {
+    api_config: {type: String},
+    enabled: {type: Boolean},
+    endpoint: {type: String},
+    startEvent: {type: String},
+    stopEvent: {type: String},
+    input_prompt: {type: String},
+    toolCallEvent: {type: String},
+    tool_call_responses: {type: String},
+  };
+
+  constructor() {
+    super();
+    this.onAudioInputReceived = (e) => {
+      if (!this.api) {
+        return;
+      }
+      this.api.sendAudioChunk(e.detail.data);
+    };
+
+    this.onVideoInputReceived = (e) => {
+      if (!this.api) {
+        return;
+      }
+      this.sendVideoChunk(e.detail.data);
+    };
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    window.addEventListener('audio-input-received', this.onAudioInputReceived);
+    window.addEventListener('video-input-received', this.onVideoInputReceived);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    window.removeEventListener(
+      'audio-input-received',
+      this.onAudioInputReceived,
+    );
+
+    window.removeEventListener(
+      'video-input-received',
+      this.onVideoInputReceived,
+    );
+
+    if (this.api) {
+      this.api.ws.close();
+    }
+  }
+
+  firstUpdated() {
+    if (this.enabled) {
+      this.setupWebSocket();
+    }
+  }
+
+  setupWebSocket() {
+    if (this.api) {
+      return;
+    }
+
+    this.api = new GeminiLiveApi(
+      this.endpoint,
+      true,
+      JSON.parse(this.api_config),
+    );
+
+    this.api.onSetupComplete = () => {
+      console.log('Setup complete...');
+    };
+
+    this.api.onAudioData = (base64Data) => {
+      this.dispatchEvent(
+        new CustomEvent('audio-output-received', {
+          detail: {data: base64Data},
+          // Allow event to cross shadow DOM boundaries (both need to be true)
+          bubbles: true,
+          composed: true,
+        }),
+      );
+    };
+
+    this.api.onClose = () => {
+      console.log('Web socket closed...');
+    };
+
+    this.api.onToolCall = (toolCalls) => {
+      this.dispatchEvent(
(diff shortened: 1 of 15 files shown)


mesop/examples/web_component/gemini_live/utils.py
```diff
@@ -0,0 +1,8 @@
+from typing import Any, Callable
+
+import mesop.labs as mel
+
+
```
wwwillchen, line 5: this seems like a good functionality inside mesop

mesop/examples/web_component/gemini_live/gemini_live.js
```diff
@@ -0,0 +1,177 @@
...
+      if (!this.api) {
+        return;
+      }
+      this.sendVideoChunk(e.detail.data);
+    };
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    window.addEventListener('audio-input-received', this.onAudioInputReceived);
```
wwwillchen, line 38: this definitely works, but it's a bit of a strange pattern since it wouldn't work with multiple instances. I'm wondering what your thoughts are about:
1) having gemini live be a master component and basically have everything else be a child, this way instead of listening to window, you would listen to this part of the DOM tree - it does some like logically this is the master/orchestrator component.
2) having everything be 1 web component. It would simplify the cross-component communication, but then you would need to do pretty much all the UI in JS instead of Python, but I wonder if #1185 would solve this issue.

(don't need any changes to this example, but just food for thought)

wwwillchen:
Thanks for creating this! I pulled it down and played with it, pretty neat :)

LGTM - left a couple of brainstorming questions. It feels like to me like quite a lot of work to do multimodal stuff in Mesop, which makes sense since the framework wasn't focused on that. I'm curious if you have thoughts about making it easier to build multimodal apps in mesop

wwwillchen (approved):

mesop/examples/web_component/gemini_live/gemini_live.js
richard-to: Yeah, I agree about this being a not great pattern. Mostly was a quick hack to avoid having to send data to Mesop and then to Gemini Live web component. I did consider some other ideas here, such as dynamically naming the event listeners, but figured I'd punt. Because, yeah, it took much more work than I would have liked to get it running.

1) The orchestrator idea sounds interesting. That sounds like it could work.
2) Yeah, I did consider that option, but agree that that option would kind of defeat the purpose of using Mesop and sort of limit the configurability of the UI. But definitely would have made things easier to implement. I think named slots for web components would sort of solve the problem, but not entirely since controlling the UI would be limited where the slots were. But for demo purposes, that may be enough. And allowing for custom control over how the UI elements look ends up adding more effort.

--------

I think one option is to include Gemini Live, Audio Player, and Audio Recorder in a single web component since those technically don't need any UI elements. The buttons could be done as a Mesop components. Then could trigger click events to that update state to the centralized web component, such as a disabling/enabling microphone. I think that could work. 

I guess Video Player could technically also be controller in the consolidated web component as well since the video preview isn't necessary. So displaying the video preview could potentially be a separate web component. 

-------

But yeah, I think this was definitely a lot more work to get going and definitely more work than I think someone building a demo on Mesop would want to do. Definitely too much javascript required. So would be cool if we could get a web component that would be easy to use and integrate.

I think one drawback of this is that the direct web socket connection on the client side isn't that secure, which means people would likely need a proxy to the web socket connection. So that would lead to more work for Mesop devs in terms of having to set that up.

I also have a version that runs the web socket in the backend Mesop server: https://github.com/richard-to/mesop-gemini-2-experiments. But the problem with this is that we likely don't want to couple the web socket connections to the Mesop server. Probably not too scalable even for demos. I didn't include those demos here since it doesn't work for 3.10. Some async feature I'm using kept throwing errors.

I think in the end, probably for more realistic use cases using video, we'd want some kind of Web RTC server rather than a web socket connection.

------

I don't have any concrete ideas of how to make it easier to build multi-modal Mesop apps as of yet. I think one purpose of trying to do this has been to see how much to see where there potential weak points are.

I think my initial thoughts are:

1. Web socket support seems more and more important (so it's good we have that now)
2. At least for the Gemini live API, seems good to separate out the Web Socket server for that API, but doing so is probably up to the user to set up, though could be helpful to point people to existing proxy implementations. But initially they could use the direct JS connection to the Gemini Live API for development
3. It's good that I was able to get things working with web components. Definitely to make them shareable and usable, they'd need to be cleaned up a lot.
4. I think a lot of it comes down to working with web components, making it easier to build them and share them, especially the latter case since ultimately we don't want the majority of Mesop devs to be digging into the javascript stuff, unfortunately building up that ecosystem of web components is a lot of work.
5. In terms of web components, there's definitely a good amount of use cases where the web component itself has no UI. So I'll either proxy the event from a UI element (Video Player) or use a slot (Audio Recorder). And getting this set up is pretty tedious. Also I'm not sure there's a established pattern yet.
6. I think the web component to web component communication can be a useful pattern, but as discussed it needs a more established design pattern
7. I think still a lot of unknowns when working with larger amounts of data

---------

I still have a couple more things I want to test out. I want to see if we can do something similar to AI Studio where each audio response can be replayed. And I think that scenario is more about how do we store the audio. Do we store it in state? And will it make the state too large after too many turns? Or do we do something like use the direct web component to web component communication since the audio data is already on the client?

mesop/examples/web_component/gemini_live/utils.py
wwwillchen: I think basically it would be updating https://github.com/google/mesop/blob/07c3190997c2d9ae4f853bca3fab433f7cb10564/mesop/component_helpers/helper.py#L468 so that None event handlers are filtered out like you're doing here.

Agree WC are a bit tedious, although I have found pointing an LLM to an existing web component file and say "make something similar to foo.py and foo.js except to do X" gets you most of the boilerplate

PR #1192, opened 2025-02-03
Add Jeopardy Live demo to Showcase

richard-to:
Here's a demo of using the multimodal live API. Basically an updated version of the Jeopardy game but now with audio input and output.

The system instructions still need more work. The tool usage doesn't work that well sometimes. It mostly consistently uses the right tool, but for some reason it doesn't seem to understand how to handle the tool responses correctly like 50% of the time. Like sometimes it will straight up just read the JSON response lol. So I feel like I'm doing something wrong or something.

Also another drawback is that currently the API can only return one modality at a time. So can't have it return both text and audio yet. So custom tool usage is pretty important in order to get the UI updates.


wwwillchen (approved):
neat - interesting example, I wonder if you're running into issues because it's an experimental model

richard-to:
> neat - interesting example, I wonder if you're running into issues because it's an experimental model

Yeah, definitely could be because it's experimental. API itself is also alpha and I remember reading somewhere that the tool usage needed improvement, but can't seem to find the link. 

I think 2.0 Flash is now the default, so maybe I can try changing the model name. Though seems doubtful.

PR #1197, opened 2025-02-11
Add a basic tab group demo

richard-to:
Punting on getting the Angular Material Tab component working for now.

So just creating a very barebones tab group as a starting point.

Screenshot 2025-02-11 at 3 48 39 PM


wwwillchen (approved):
thx

PR #1199, opened 2025-02-13
Add config option for MESOP_PROD_UNREDACTED_ERRORS

wwwillchen:


richard-to (approved):

richard-to:
Maybe a slight nit would be to name it for MESOP_FORCE_SHOW_UNREDACTED_ERRORS. But I think it doesn't matter much. Mainly a slight semantic suggestion that isn't fully accurate either. Mainly if you set it to False, it's still going to show unredacted errors in dev mode. So the env var name doesn't fully match. But I think with the description it's fine.

wwwillchen:
agree the. naming is a little funky, i've renamed it to MESOP_PROD_UNREDACTED_ERRORS which should be clearer.

Issue #301, opened 2024-05-24
Include x/y coordinates on click event


richard-to:
Including the x/y coordinates (clientX, clientY, offsetX, offsetY) on a click event could be helpful (especially with the position properties in https://github.com/google/mesop/issues/280)

VishalGawade1:
Hello @wwwillchen,
I was looking for good first issues and this seems to be a good start can I still work on it?

richard-to:
Hi @VishalGawade1, feel free to work on it.

sepehrrasooli:
Hi @richard-to, I noticed that the latest PR was closed and there's no active PR for this issue. If no one is currently working on it, I'd be happy to take it on. Let me know if it's still available!

richard-to:
@sepehrrasooli - Yes, feel free to work on it. Let me know if you have any questions.

sepehrrasooli:
> [@sepehrrasooli](https://github.com/sepehrrasooli) - Yes, feel free to work on it. Let me know if you have any questions.

Hi @richard-to, Sorry for the late reply! Thanks for the go-ahead on this—I’ll start working on it. I’ll reach out if I run into any questions.

PR #1207, opened 2025-02-17
added x and y coordinates

freakycows:
I added this fix in response to issue request #301, which asked for x and y coordinates to be made for click events.


wwwillchen:
closing because cla has not been signed

PR #1208, opened 2025-02-20
Add chartjs example web component

richard-to:
# Example chart rendering

Screenshot 2025-02-16 at 4 52 57 PM

# Example using a modified Mesop chat with Gemini API

Not the most realistic example. Probably more realistic if the user could upload data.

Screenshot 2025-02-17 at 12 41 27 PM


wwwillchen (approved):
thx

PR #1214, opened 2025-02-27
Add Google Maps example web component

KMarkert:
## Example map rendering and interactions:
google_maps_mesop_example

> Note: This will most likely require linting to be done. If there is a linting tool I should use please let me know.

### DIFF
--- a/mesop/examples/web_component/google_maps/google_maps_app.py
+++ b/mesop/examples/web_component/google_maps/google_maps_app.py
@@ -4,63 +4,65 @@ Note: Does not implement full interactions with a map. This is meant as a
     a web map.
 """
 
-from pydantic import BaseModel
 from typing import Tuple
+
+from pydantic import BaseModel
 
 import mesop as me
 import mesop.labs as mel
-
-from mesop.examples.web_component.google_maps import(
-    google_maps_component,
-) 
+from mesop.examples.web_component.google_maps.google_maps_component import (
+  google_maps_component,
+)
 
 
 @me.page(
-    path="/google_maps_example",
-    security_policy=me.SecurityPolicy(
-        allowed_script_srcs=[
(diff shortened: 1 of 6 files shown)


richard-to:
In terms of the linting and presubmit stuff, you'll need to set up the dev env: https://google.github.io/mesop/internal/development/

Which is a bit annoying to set up.

The other option is to use a GitHub Codespace: https://google.github.io/mesop/internal/codespaces/

It does take some time to initialize and install everything, but I believe that should set up all the pre-commit checks as well as the linting in VSCode.

mesop/examples/web_component/google_maps/google_maps_app.py
```diff
@@ -0,0 +1,66 @@
...
+    latlng: Tuple[float, float]
+
+
+def on_map_click(e: mel.WebEvent):
+    click = MapClick(**e.value)
+    me.state(State).click_location = f"Latitude: {click.latlng[0]}, Longitude: {click.latlng[1]}"
+    print(f"Map clicked at: {click.latlng}")
+
+
+if __name__ == "__main__":
```
richard-to, line 65: This part is not needed -- I ended up removing this for you when I was applying the linting.

mesop/examples/web_component/google_maps/google_maps_app.py
```diff
@@ -0,0 +1,66 @@
...
+    a web map.
+"""
+
+from pydantic import BaseModel
+from typing import Tuple
+
+import mesop as me
+import mesop.labs as mel
+
+from mesop.examples.web_component.google_maps import(
```
richard-to, line 13: `mesop.examples.web_component.google_maps.google_maps_component` -> Also just made this adjustment for you since I was making the edits.

richard-to (approved):
This looks good. I'll push the linting changes for you.

PR #1233, opened 2025-03-12
Fix race condition by queueing message processing in Channel

wwwillchen:
Fixes #1232 

As noted in the bug, none of our tests really detect this flakiness on GitHub. I tested this downstream and it fixes the flakiness.


richard-to (approved):

richard-to:
Thanks for fixing this one!

PR #1235, opened 2025-03-12
Expose web components API to mesop namespace and migrate (most) web component examples

wwwillchen:
Preparation for v1 launch.

Main changes are in:

```
- mesop/__init__.py
- mesop/labs/__init__.py
```

The rest are docs & example changes. I've left one web component example using the mesop.labs import just to make sure we don't break it.


richard-to (approved):
Awesome. Will there be any changes needed downstream. I think there are some people using web components internally.

wwwillchen:
> Awesome. Will there be any changes needed downstream. I think there are some people using web components internally.

Nope, I purposely left all of the web component APIs also exported on `mesop.labs` so this should be a backwards-compatible change.

PR #1243, opened 2025-03-18
Bump to v1.0.0

wwwillchen:
@richard-to let me know if there's anything else we should land before cutting v1.


richard-to (approved):
I was hoping to get the updated Demo site changes. But I've been so busy that I don't think that's likely. So I think it seems good.

PR #1254, opened 2025-03-31
Cache prod_bundle.js

wwwillchen:


docs/api/config.md
```diff
@@ -270,6 +270,18 @@ By default, this is not enabled. You can enable this by setting it to `true`.
...
+
+Enables HTTP caching for the main JS bundle (i.e. prod_bundle.js) served by Mesop. This minimizes unnecessary HTTP requests for the JS bundle.
+
+Note: this does *not* cache other JS files like web components, which are configured separately (see below).
+
+### MESOP_WEB_COMPONENTS_HTTP_CACHE_KEY
+
+Enables HTTP caching for the web component entry point modules.
+
+Note: this does *not* cache JS files imported by the web component entry modules.
```
richard-to, line 283: Slightly confusing. Took me a second to get what you were saying here. At first thought it was copy pasta from above. I think overall it's fine though. Just some people may misread if they aren't reading carefully. Which I think is ok.

richard-to (approved):

docs/api/config.md
wwwillchen: agree it's kind of confusing, but I think it's OK for now since this is experimental.

PR #1258, opened 2025-04-01
chore: update diff_component_spec.ts

eltociear:
mutliple -> multiple


richard-to (approved):
Thanks for the typo fix, @eltociear.

PR #1262, opened 2025-04-09
refactor: remove `afterRender`

JeanMeche:
`afterRenderEffect` is better suited here as it will only be executed when the signal changes.

Note: Angular intends to rename the `afterRender` api.

### DIFF
--- a/mesop/web/src/shell/shell.ts
+++ b/mesop/web/src/shell/shell.ts
@@ -1,11 +1,12 @@
 import {
+  afterRenderEffect,
   ApplicationRef,
   Component,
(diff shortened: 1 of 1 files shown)


JeanMeche:
Back to draft, we actually want to re-asses what the right pattern would be here before suggesting a change of API.

richard-to:
Ok sounds good.

JeanMeche:
The correct solution here seems to be to rely on an `afterRenderEffect` that is trigger when the signal changes.

richard-to (approved):

PR #1267, opened 2025-04-19
fix(demo): capitalize input in text-to-text example

bobtista:
The demo for text_to_text is supposed to capitalize the input. It's actually correct in the readme but was wrong in the code.

### DIFF
--- a/demo/text_to_text.py
+++ b/demo/text_to_text.py
@@ -22,4 +22,4 @@ def app():
 
 
 def upper_case_stream(s: str):
-  return "Echo: " + s
+  return "Echo: " + s.capitalize()


demo/text_to_text.py
```diff
@@ -22,4 +22,4 @@ def app():
 
 
 def upper_case_stream(s: str):
```
richard-to, line 24: Now that I think of it. I think the naming of the function is slightly misleading since it could be interpreted as upper casing all characters. Maybe we should change it to `capitalize_stream` here and the readme. Or we could change s.capitalize to s.upper here and in the readme.

richard-to (approved):
Thanks for the catch. You're right. I added one miscellaneous comment that occurred to me.

PR #1273, opened 2025-04-28
Add coordinate values for click events + add right click event + add context menu example

### LINKED ISSUE #301: Include x/y coordinates on click event

### LINKED ISSUE #1272: Add support for right click event
Right now we have a click event, but we don't know if it is a right click or not. Maybe we can pass that info in.

richard-to:
- This change adds some new properties to the click event for getting x/y coordinates on click event
- Also adds a right click event that overrides contextmenu event (ended up making this a separate event so we know when to disable the contextmenu (though maybe could have used click event and added is_right_click)
- Also added simple example of a context menu in Mesop with a right click.

- Closes #1272 
- Closes #301

### DIFF
--- a/demo/context_menu.py
+++ b/demo/context_menu.py
@@ -1,0 +1,136 @@
+"""Simple context menu example"""
+
+from pydantic import BaseModel
+
+import mesop as me
+
+
+class MenuItem(BaseModel):
+  label: str
+
+
+class Menu(BaseModel):
+  items: list[MenuItem]
+
+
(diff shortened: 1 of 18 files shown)


wwwillchen (approved):
personally, i think doing a context menu impl in web component seems a little better (avoids latency for a normally native browser event), but this works too

PR #1274, opened 2025-04-28
Add support for configuring Cross Origin Opener Policy

richard-to:
- We should probably set the default to same-origin, but worried about it breaking some existing apps, so just left it as the default value

### DIFF
--- a/mesop/examples/testing/__init__.py
+++ b/mesop/examples/testing/__init__.py
@@ -11,6 +11,9 @@ from mesop.examples.testing import (
   conditional_event_handler as conditional_event_handler,
 )
 from mesop.examples.testing import (
+  coop as coop,
+)
+from mesop.examples.testing import (
   csp_escaping as csp_escaping,
 )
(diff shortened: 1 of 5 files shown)


wwwillchen:
what's the motivation for this? (the issue doesn't say much), but change lgtm. might be nice to update https://mesop-dev.github.io/mesop/guides/web-security/ to suggest a safer value than the default

wwwillchen (approved):

richard-to:
Mainly added to support a downstream request. But I think it's a good feature to have and I think defaulting to same-origin should work for most Mesop users.

Seems mainly to resolve the issue of XS Leaks. https://xsleaks.dev/ does a good job of explaining. I do not fully understand it myself, but my understanding is that there are ways to infer implicit information about a website. And to prevent these kinds of inferences, one way to do this is to add a COOP header to enforce access is from the same origin.

Sure I can update the Security Policy docs. Thanks for pointing that out.

Issue #1283, opened 2025-05-03
Query params not being read or updated correctly in Web Sockets mode


richard-to:
**Describe the bug**

- Query params from URL are not read by Mesop in Web Sockets mode
- Also updates to query params in Mesop do not reflect the URL
- Using me.navigate does update the query params though.

**To Reproduce**

Play around with the query params demo: /workspaces/mesop/mesop/examples/query_params.py

**Expected behavior**

- Query params should be read from URL and reflect in me.query_params
- Query params should be updated in the URL when updated in me.query_params

richard-to:
Looked into this a bit yesterday. Seems like some kind of race condition, but haven't pinpointed where yet.

Issue #1276, opened 2025-04-28
Query Params not updated when in web socket mode


richard-to:
- The query params are read on load by the app
- The query params on initial load are correct
- Later there is a me.navigate that updates the query params
- The URL updates correctly here
- Then later the web socket disconnects and connects again. The query params are read on load, but the updated query params are not used. instead the initial query params are still being read.

-------

Seems like this could be an issue with web sockets since I think maybe the updated query param state is not being passed back to the server. Normally this would be passed by SSE mode. But haven't looked too closely at the issue.

-----

richard-to:
Looks like the issue may be here. But not 100% sure. Seems like the onopen call will only ever get the initial request object on reload, rather the latest version. But this assumes onopen is called on a reconnect

```
  private initWebSocket(initParams: InitParams, request: UiRequest) {
    if (this.webSocket?.readyState === WebSocket.OPEN) {
      this.status = ChannelStatus.OPEN;
      const payload = generatePayloadString(request);
      this.webSocket.send(payload);
      return;
    }

    const wsProtocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:';
    const wsUrl = `${wsProtocol}//${window.location.host}/__ui__`;

    this.webSocket = new WebSocket(wsUrl);
    this.status = ChannelStatus.OPEN;
    this.isWaitingTimeout = setTimeout(() => {
      this.isWaiting = true;
    }, WAIT_TIMEOUT_MS);

    const {zone} = initParams;
    this.initParams = initParams;

    this.webSocket.onopen = () => {
      // Send the initial UiRequest upon connection
      const payload = generatePayloadString(request);
      this.webSocket!.send(payload);
      this.wsReconnectAttempts = 0;
    };
```

nirben82:
I debugged this a little and it seems that the query_params that are received as part of the ui_request are always of the previous state. So I believe its not that the query params are not updated during the session. Its that the query params record the state of before the me.navigate. Also, I also turned off web sockets and the issue persists, so unrelated.

nirben82:
[...]

PR #1284, opened 2025-05-03
This is a temp fix for Web Socket reconnect behavior.

richard-to:
- Currently, the web socket will use the initial state when reconnecting.
- Ideally it should try to preserve the current UI state if possible.

This PR only provides a temporary fix to ensure query params are updated
to help a resolve user issue where they would like to reload state from
the db by using the update query params. This is because on Cloud Run
web socket connections last about 10 minutes or something before they
disconenct and reconenction is required.

However, there may be an additional issue with query params. See #1283
that may block this temp fix.

Ref #1276

### DIFF
--- a/mesop/web/src/services/channel.ts
+++ b/mesop/web/src/services/channel.ts
@@ -7,6 +7,7 @@ import {
   UserEvent,
   Component as ComponentProto,
   UiResponse,
+  InitRequest,
   Command,
   ChangePrefersColorScheme,
   HotReloadEvent,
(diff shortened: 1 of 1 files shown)


wwwillchen (approved):
If the websocket stuff is too much of a maintenance burden it may be worth deprecating it since it's only an experimental feature. B/c I only used it for a local app, I never really hardened it for this kind of behavior (but IIRC, this is basically the kind of stuff other websockets-based frameworks like Streamlit do and it becomes pretty non-trivial)

PR #1376, opened 2026-04-12
feat: add me.set_cookie() and me.delete_cookie() API

### LINKED ISSUE #1304: State is reset on browser refresh - breaks authentication flows
## Bug Description
Mesop state is completely reset on browser refresh, making it impossible to maintain authentication tokens or any persistent user data. This renders Mesop unusable for real-world applications requiring user authentication.

## Expected Behavior
- State should persist across browser refreshes within the same session
- Authentication tokens stored in state should remain available until session expires
- State should be tied to browser session/Flask session cookies

## Actual Behavior
- All state values are reset to defaults on browser refresh (F5/Ctrl+R)
- New session files are created with each page reload (when using file backend)
- Users must re-authenticate after every refresh
- No state persistence across page reloads

## Steps to Reproduce

1. Create a Mesop app with authentication state:
```python
import mesop as me

@me.stateclass
class AuthState:
    access_token: str = ""
    user_id: str = ""
    is_authenticated: bool = False

def login_success(token: str, user_id: str):
    state = me.state(AuthState)
    state.access_token = token
    state.user_id = user_id 
    state.is_authenticated = True

@me.page(path="/")
def main():
    state = me.state(AuthState)
    if state.is_authenticated:
        me.text(f"Logged in as: {state.user_id}")
    else:
        me.text("Please login")
        me.button("Login", on_click=lambda e: login_success("token123", "user456"))
```

2. Run the app and click "Login"
3. Verify state shows "Logged in as: user456"
4. Refresh the browser (F5 or Ctrl+R)
5. **Bug:** State is reset, shows "Please login" again

## Investigation Results

**File Backend Behavior:**
When using the file backend, each browser refresh creates a new session file:
```
session_abc123.json  # Created on first visit
session_def456.json  # Created after refresh - previous lost
session_ghi789.json  # Created after another refresh
```

This confirms that Mesop creates new sessions instead of restoring existing ones.

## Tested Solutions (None Work)

**✗ All State Backends Tested:**
- Default memory backend
- File backend
- All other available backends
- **Result: All backends reset state on refresh**

**✗ Manual Flask Session Management:**
```python
from flask import session

# Attempted workaround - doesn't work
def save_to_flask_session():
    state = me.state(AuthState)
    session['access_token'] = state.access_token
    session.permanent = True

def restore_from_flask_session():
    state = me.state(AuthState)
    if 'access_token' in session:
        state.access_token = session['access_token']
        state.is_authenticated = True
```

**✗ Flask Configuration Attempts:**
```python
import os
os.environ['SECRET_KEY'] = 'secret-key'
app.secret_key = 'secret-key'  
app.config['SESSION_PERMANENT'] = True
app.config['PERMANENT_SESSION_LIFETIME'] = 86400 * 30
```

## Environment
- **Mesop version:** 1.1.0
- **Python version:** 3.x
- **Browser:** Chrome/Firefox/Safari (all affected)
- **State backend:** Tested all available backends - same issue
- **OS:** Windows/macOS/Linux (all affected)

## Impact
This bug makes Mesop **unusable for production applications** because:
- Users must re-authenticate after every page refresh
- Shopping carts are lost on refresh
- Form data is lost on accidental refresh
- Any user-specific state is permanently volatile
- Creates terrible user experience

## Proposed Solution
State persistence should be tied to browser sessions:
1. When a page loads, check for existing Flask session cookie
2. If valid session exists, restore corresponding state
3. Only create new state when no valid session cookie present
4. State should persist until session expires (configurable timeout)

## Workaround
**None found.** This appears to be a fundamental architectural limitation that requires a fix in Mesop core.

## Additional Context
This is blocking adoption of Mesop for any application requiring:
- User authentication
- Shopping carts
- Form wizards
- Any persistent user state

Please prioritize this issue as it affects the core usability of the framework for real-world applications.

richard-to:
## Summary

Adds a native `me.set_cookie()` and `me.delete_cookie()` API so Mesop apps can set browser cookies from within event handlers. Closes #1374.

Also closes https://github.com/mesop-dev/mesop/issues/1304.

### Problem

Mesop's `/__ui__` endpoint is an SSE stream — Flask commits HTTP response headers (including `Set-Cookie`) *before* the event-handler generator body runs. This makes `after_this_request` and direct cookie manipulation ineffective from within event handlers, leaving no built-in way to issue `Set-Cookie` headers from app code.

### Solution: two-phase cookie application

1. **`me.set_cookie()` / `me.delete_cookie()`** queue `PendingCookie` objects on the request context (Python side only — no network round-trip yet).
2. After each event-handler yield inside `generate_data()`, any pending cookies are encoded into a short-lived **itsdangerous-signed token** (60 s TTL) that embeds the cookie data directly (no server-side storage). An **`ApplyCookiesCommand`** carrying that token is appended to the render commands. Any server worker holding the same `MESOP_COOKIE_SECRET_KEY` can verify and redeem the token, so no sticky sessions are required.
3. The **TypeScript client** receives the command and makes a **`POST /__apply-cookies`** request with the token in the **form body** (`t=`, `credentials: 'same-origin'`). Sending the token in the body (not the URL) keeps it out of server access logs and browser history.
4. The **`/__apply-cookies` Flask endpoint** validates the Origin header (same-site check, matching `/__ui__`), verifies the itsdangerous signature and TTL, marks the token's nonce as used (single-use within a process), then returns an empty `204` response with the appropriate `Set-Cookie` headers.

This approach works identically in both SSE and WebSocket modes. The signed token with a short TTL and per-process nonce tracking prevents replay attacks; the same-site Origin check prevents a cross-site attacker from replaying a captured token against a victim's browser.

**Note on value visibility:** the token payload is signed but not encrypted, so cookie values are Base64-readable as the token passes through browser JavaScript. For values that must never be visible to JS, use `@me.cookieclass(encrypted=True)`.

**Note on multi-worker replay:** nonces are tracked per-process, so cross-worker replay is theoretically possible but mitigated by the CSRF Origin check and the short TTL. See `MESOP_COOKIE_SECRET_KEY` docs for details.

**`MESOP_COOKIE_SECRET_KEY` is required** for all cookie write operations (`me.set_cookie()`, `me.delete_cookie()`).

## Files changed

| File | Change |
|---|---|
| `mesop/protos/ui.proto` | Add `ApplyCookiesCommand` message (field 8 in `Command` oneof) |
| `mesop/protos/ui_pb2.py` | Regenerated from updated proto |
| `mesop/runtime/context.py` | Add `PendingCookie` dataclass; add `set_cookie()`, `delete_cookie()`, `pending_cookies()`, `clear_pending_cookies()` to `Context` |
| `mesop/commands/set_cookie.py` | New file — public `me.set_cookie()` / `me.delete_cookie()` wrappers |
| `mesop/__init__.py` | Export the two new functions |
| `mesop/server/server.py` | Add `_CookieTokenCache` (itsdangerous-signed + nonce tracking), `maybe_append_apply_cookies_command()`, `/__apply-cookies` route with same-site CSRF guard |
| `mesop/web/src/shell/shell.ts` | Handle `hasApplyCookies()` command with a `fetch()` POST to `/__apply-cookies` |

## Usage example

```python
import mesop as me
from werkzeug.security import check_password_hash

USERS = {"alice": ""}

@me.stateclass
class State:
    username_input: str
    password_input: str

def on_login(e: me.ClickEvent):
    state = me.state(State)
    stored = USERS.get(state.username_input)
    if not stored or not check_password_hash(stored, state.password_input):
        return
    # Cookie is applied via a lightweight follow-up POST request
    me.set_cookie(
        "session",
        create_session_token(state.username_input),
        max_age=86400,   # 1 day
        httponly=True,
        secure=True,
        samesite="Lax",
    )
    me.navigate("/app")
```

## Test plan

- [ ] Call `me.set_cookie()` from a button click handler; verify `Set-Cookie` header appears in the `/__apply-cookies` response in browser DevTools
- [ ] Verify the cookie persists across hard page refreshes and new tabs
- [ ] Call `me.delete_cookie()` and verify the cookie is removed
- [ ] Verify the token is one-time (within a process): POSTing to `/__apply-cookies` a second time with the same token returns `400`
- [ ] Verify the token expires: calling `/__apply-cookies` after 60 s returns `400`
- [ ] Test with `MESOP_WEBSOCKETS_ENABLED=true` — cookie should still be applied correctly
- [ ] Test with `MESOP_BASE_URL_PATH` set — client should call the prefixed endpoint


mesop/web/src/shell/shell.ts
```diff
@@ -243,6 +244,16 @@ export class Shell {
...
+          } else if (command.hasApplyCookies()) {
+            const token = command.getApplyCookies()!.getToken();
+            // Apply cookies via a dedicated endpoint so that Set-Cookie headers
+            // can be sent on a normal HTTP response rather than SSE/WebSocket.
+            await fetch(
+              prefixBasePath(
+                `/__apply-cookies?t=${encodeURIComponent(token || '')}`,
+              ),
+              {method: 'GET', credentials: 'same-origin'},
+            );
```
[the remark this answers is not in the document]
richard-to: Fixed in 55144d4. Added token non-empty check, `if (!applyResp.ok)` error throw, and `cache: 'no-store'`. Also switched from GET to POST (token now sent in the request body rather than the URL, to keep it out of logs).

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/server/server.py
```diff
@@ -101,6 +146,17 @@ def configure_flask_app(
...
+  def maybe_append_apply_cookies_command() -> None:
+    """If the context has pending cookies, cache them and append an ApplyCookiesCommand."""
+    pending = runtime().context().pending_cookies()
+    if not pending:
+      return
+    token = _cookie_token_cache.put(list(pending))
+    runtime().context().clear_pending_cookies()
+    runtime().context().commands().append(
+      pb.Command(apply_cookies=pb.ApplyCookiesCommand(token=token))
+    )
```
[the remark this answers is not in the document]
richard-to: Fixed in 55144d4. `maybe_append_apply_cookies_command()` is now called before `render_loop` in all three `on_load` code paths: the generator on_load loop in `generate_data`, the non-generator on_load branch, and inside `run_page_load` (used when navigating to a page with an on_load handler).

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/server/server.py
```diff
@@ -45,9 +48,51 @@
...
+    Returns None if the token is unknown or has expired.
+    """
+    with self._lock:
+      entry = self._store.pop(token, None)
+    if entry is None:
+      return None
+    cookies, expiry = entry
+    if time.monotonic() > expiry:
+      return None
+    return cookies
```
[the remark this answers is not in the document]
richard-to: Fixed in 55144d4. Added `_evict_expired_locked()` which is called under the lock on every `put()` and `pop()`. This keeps the store bounded without needing a background thread or LRU eviction — in practice the store will only ever have a handful of entries at a time (one per in-flight event handler), so opportunistic cleanup on each write is sufficient.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/server/server.py
```diff
@@ -342,6 +399,35 @@ def ui_stream() -> Response:
...
+        cookie.name,
+        cookie.value,
+        max_age=cookie.max_age,
+        path=cookie.path,
+        domain=cookie.domain,
+        secure=cookie.secure,
+        httponly=cookie.httponly,
+        samesite=cookie.samesite,
+      )
+    return resp
```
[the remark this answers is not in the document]
richard-to: Fixed in 55144d4. Added `resp.headers["Cache-Control"] = "no-store"` to the response. Also switched the endpoint from GET to POST so the token is in the request body rather than the URL, which further reduces exposure through caching/logging infrastructure.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/runtime/context.py
```diff
@@ -241,6 +267,54 @@ def set_page_title(self, title: str) -> None:
...
+        name=name,
+        value=value,
+        max_age=max_age,
+        path=path,
+        domain=domain,
+        secure=secure,
+        httponly=httponly,
+        samesite=samesite,
+      )
+    )
```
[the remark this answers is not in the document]
richard-to: Fixed in 55144d4. `context.set_cookie()` now validates `samesite` against `("Strict", "Lax", "None")` and raises `MesopDeveloperException` for invalid values. It also raises if `samesite="None"` is combined with `secure=False`, since browsers reject SameSite=None cookies that aren't Secure (RFC 6265bis).

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/server/server.py
```diff
@@ -342,6 +399,35 @@ def ui_stream() -> Response:
...
+    from the server.  The token is single-use and expires after
+    _COOKIE_TOKEN_TTL_SECONDS seconds, so replay attacks are not possible.
+    """
+    token = request.args.get("t", "")
+    if not token:
+      abort(400, "Missing token")
+    pending: list[PendingCookie] | None = _cookie_token_cache.pop(token)
+    if pending is None:
+      abort(400, "Invalid or expired token")
+
```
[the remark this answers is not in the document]
richard-to: Fixed in 55144d4. Added a `token is single-use (replay returns 400)` Playwright test that uses `page.route()` to intercept the POST to `/__apply-cookies` and capture the token from the request body, then replays it with `page.request.post()` and asserts the server responds with 400.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/commands/cookie_class.py
```diff
@@ -0,0 +1,330 @@
...
+
+  def decorator(c: type[T]) -> type[T]:
+    if not dataclasses.is_dataclass(c):
+      c = dataclasses.dataclass(c)
+    cookie_name = name if name is not None else _to_snake_case(c.__name__)
+    _COOKIE_CLASSES[c] = _CookieClassMeta(
+      name=cookie_name, signed=signed, encrypted=encrypted
+    )
+    return c
+
```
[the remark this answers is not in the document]
richard-to: Fixed. `cookieclass` now calls `dataclass_with_defaults(c)` at decoration time (same helper used by `stateclass`), so every field is guaranteed to have a default and `cls()` always succeeds without arguments.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/runtime/context.py
```diff
@@ -241,6 +267,63 @@ def set_page_title(self, title: str) -> None:
...
+    path: str = "/",
+    domain: str | None = None,
+  ) -> None:
+    self._pending_cookies.append(
+      PendingCookie(
+        name=name,
+        value="",
+        max_age=0,
+        path=path,
+        domain=domain,
```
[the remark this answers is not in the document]
richard-to: Fixed. `Context.delete_cookie()` now has an explicit `secure: bool = False` parameter (defaulting to `False` rather than relying on `PendingCookie.secure=True`). The public `me.delete_cookie()` API resolves `secure` the same way as `set_cookie` — via `_resolve_secure(None)` which auto-detects `True` on HTTPS and `False` on HTTP.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/commands/set_cookie.py
```diff
@@ -0,0 +1,172 @@
...
+      me.delete_cookie(SessionCookie)     # high-level: class lookup
+
+  Args:
+    name: Cookie name (``str``) or a ``@me.cookieclass``-decorated class.
+    path: URL path scope that was used when creating the cookie.
+    domain: Domain scope that was used when creating the cookie.
+  """
+  if isinstance(name, str):
+    runtime().context().delete_cookie(name, path=path, domain=domain)
+  else:
```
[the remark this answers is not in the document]
richard-to: Fixed. `delete_cookie()` now has `secure: bool | None = None` and resolves it via `_resolve_secure()` (same auto-detection as `set_cookie`): `True` on HTTPS, `False` on HTTP, or explicit override. This ensures the deletion cookie matches the security context of the original cookie.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/server/server.py
```diff
@@ -101,6 +161,17 @@ def configure_flask_app(
...
+  def maybe_append_apply_cookies_command() -> None:
+    """If the context has pending cookies, cache them and append an ApplyCookiesCommand."""
+    pending = runtime().context().pending_cookies()
+    if not pending:
+      return
+    token = _cookie_token_cache.put(list(pending))
+    runtime().context().clear_pending_cookies()
+    runtime().context().commands().append(
+      pb.Command(apply_cookies=pb.ApplyCookiesCommand(token=token))
+    )
```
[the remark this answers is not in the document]
richard-to: Fixed. Added `runtime().context().clear_pending_cookies()` in `generate_data`'s `except` block, so cookies queued during a failed handler are discarded rather than carried over to the next event in long-lived WebSocket contexts.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/guides/auth.md
```diff
@@ -433,3 +434,185 @@ gunicorn --bind 0.0.0.0:8080 'your_module:app'
...
+
+Mesop provides a first-class cookie API so you can persist small pieces of data (session tokens, user preferences, etc.) in the browser without managing raw `Set-Cookie` headers yourself.
+
+### How it works
+
+Because Mesop event-handler responses use Server-Sent Events (SSE) or WebSockets — neither of which supports `Set-Cookie` headers — cookies are applied via a lightweight two-step protocol:
+
+1. Your event handler calls `me.set_cookie()`.
+2. Mesop stores the pending cookies server-side under a short-lived single-use token, then sends an `ApplyCookiesCommand` to the browser.
+3. The Mesop client POSTs the token to `/__apply-cookies`, which responds with the `Set-Cookie` headers.
```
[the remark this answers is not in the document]
richard-to: Fixed. The inaccurate wording ("SSE or WebSockets — neither of which supports Set-Cookie headers") was replaced with the technically correct explanation: HTTP response headers including Set-Cookie are committed before the event-handler body executes, which is the actual constraint. This thread is also marked outdated since the relevant text has since been further revised.

---
_Generated by [Claude Code](https://claude.ai/code)_

mesop/server/server.py
```diff
@@ -45,9 +50,103 @@
...
+    try:
+      payload = URLSafeTimedSerializer(
+        _get_cookie_secret_key(), salt="mesop-apply-cookies"
+      ).loads(token, max_age=_COOKIE_TOKEN_TTL_SECONDS)
+    except (BadSignature, SignatureExpired):
+      return None
+
+    nonce = payload.get("n", "")
+    with self._lock:
+      if nonce in self._used_nonces:
```
richard-to, line 134: I just want to double check. If a nonce gets evicted, the same token cannot be used again since the TTL is enforced by URLSafeTimedSerializer?

docs/guides/auth.md
```diff
@@ -433,3 +434,185 @@ gunicorn --bind 0.0.0.0:8080 'your_module:app'
...
+#### `me.delete_cookie()`
+
+```python
+me.delete_cookie("session_id")      # low-level: explicit string name
+me.delete_cookie(SessionCookie)     # high-level: class lookup
+```
+
+### Security notes
+
+- **`httponly=True` (default):** The cookie is not accessible to JavaScript, which mitigates XSS-based session theft.  It is still visible in browser DevTools.
```
richard-to, line 631: We may want to mention that technically the data is visible when we pass to the apply cookie command? Which we need for applying the cookie?

mesop/server/server.py
richard-to: Yes, exactly. Both expiry windows are the same 60 seconds (`_COOKIE_TOKEN_TTL_SECONDS`). When `_evict_used_locked()` removes a nonce it means 60 s have elapsed since it was first consumed — and at that point `URLSafeTimedSerializer.loads(..., max_age=60)` will raise `SignatureExpired` for the same token, so `pop()` returns `None` before it ever reaches the nonce lookup. The two checks are deliberately aligned so there's no window where a nonce is gone but the signature is still valid.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/guides/auth.md
richard-to: Good call — added a bullet to the Security notes section explaining that cookie values are Base64-readable in the `ApplyCookiesCommand` token while it passes through browser JS, and directing users to `@me.cookieclass(encrypted=True)` if the raw value must never be visible.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/api/config.md
```diff
@@ -286,6 +286,61 @@ Enables HTTP caching for the web component entry point modules.
 
 Note: this does *not* cache JS files imported by the web component entry modules.
 
+### MESOP_COOKIE_SECRET_KEY
+
+!!! warning "Experimental"
+    The signed/encrypted cookie API is experimental and may change in future releases.
+
+Secret key used by the cookie security features.  Required for
```
richard-to: I think this line is unnecessarily specific. In most cases the user needs to write a cookie before they can read one right? So I feel like basically the user needs to set MESOP_COOKIE_SECRET_KEY in order for this to be useful?

mesop/commands/cookie_class.py
```diff
@@ -0,0 +1,332 @@
...
+  The cookie name defaults to the snake_case version of the class name.
+  Override it with the optional ``name`` keyword argument.
+
+  **Signed cookies** (``signed=True``) add an HMAC to the cookie value so
+  any tampering is detected on read.  The contents are still Base64-visible
+  in DevTools.  Requires ``MESOP_COOKIE_SECRET_KEY`` to be set.
+
+  **Encrypted cookies** (``encrypted=True``) use Fernet symmetric encryption
+  so the contents are completely hidden.  Requires the ``cryptography``
+  package (``pip install cryptography``) and ``MESOP_COOKIE_SECRET_KEY`` to be set.
```
richard-to: We can remove requires MESOP_COOKIE_SECRET_KEY to be set for each of these. Just update the comment to say that this class needs MESOP_COOKIE_SECRET_KEY to be set. Also it's a bit redundant when you mention this in the args as well.

mesop/server/server.py
```diff
@@ -45,9 +50,107 @@
...
 logger = logging.getLogger(__name__)
 
+_COOKIE_TOKEN_TTL_SECONDS = 60
+
+
+def _get_cookie_secret_key() -> str:
+  """Return MESOP_COOKIE_SECRET_KEY, raising MesopDeveloperException if unset."""
+  key = os.environ.get("MESOP_COOKIE_SECRET_KEY", "")
+  if not key:
+    raise MesopDeveloperException(
```
richard-to, line 64: Again, we should just make it required for using cookies

docs/guides/auth.md
```diff
@@ -433,3 +434,203 @@ gunicorn --bind 0.0.0.0:8080 'your_module:app'
...
+---
+
+## Cookies
+
+!!! warning "Experimental"
+    The cookie API is experimental. The interface may change in future releases.
+
+Mesop provides a first-class cookie API so you can persist small pieces of data (session tokens, user preferences, etc.) in the browser without managing raw `Set-Cookie` headers yourself.
+
+!!! note "MESOP_COOKIE_SECRET_KEY requirements"
```
richard-to: This note can be removed. This is out dated since the secret key is needed to use cookies. It's true that some operations don't require the secret key, but in practice it's not usable without it. That technicalities create confusion.

Mainly we always need the secret key for the Apply Cookies Command.

docs/guides/auth.md
```diff
@@ -433,3 +434,203 @@ gunicorn --bind 0.0.0.0:8080 'your_module:app'
...
+def on_logout(e: me.ClickEvent):
+    state = me.state(State)
+    state.logged_in = False
+    state.username = ""
+    me.delete_cookie(SessionCookie)
+```
+
+### Signed and encrypted cookies
+
+By default `@me.cookieclass` stores the JSON value in plain text — readable in browser DevTools.  For cookies that carry sensitive data you can add tamper-protection or full encryption by setting `signed=True` or `encrypted=True` on the decorator.  [`MESOP_COOKIE_SECRET_KEY`](../api/config.md#mesop_cookie_secret_key) is already required for all cookie ops; these options add extra protection on top of that.
```
richard-to: Remove last sentence about Mesop Cookie Secret. It's too much repetition.

mesop/commands/cookie_class.py
```diff
@@ -0,0 +1,334 @@
...
+    class SessionCookie:
+        username: str = ""
+
+Encrypted cookies (tamper-proof, contents hidden)::
+
+    @me.cookieclass(encrypted=True)
+    class SessionCookie:
+        username: str = ""
+
+Both ``signed`` and ``encrypted`` require ``MESOP_COOKIE_SECRET_KEY`` to be set.
```
richard-to: Remove this line or update to say that MESOP_COOKIE_SECRET_KEY is required to use the class. I think it should be removed since it's mentioned again below in another comment.

mesop/commands/set_cookie.py
```diff
@@ -0,0 +1,190 @@
...
+  ``None`` auto-detects: returns ``True`` when the request is HTTPS,
+  ``False`` when it is plain HTTP (e.g. local development).
+  """
+  if secure is not None:
+    return secure
+  try:
+    from flask import request
+
+    return request.is_secure
+  except RuntimeError:
```
richard-to, line 23: When can a RuntimeError occur? Is this a common occurrence? Is it possible for a secure request to be marked as not secure?

Maybe should add a warning log here?

mesop/examples/set_cookie.py
```diff
@@ -0,0 +1,75 @@
...
+- me.cookie(SessionCookie) reads the cookie from the current request and
+  returns a populated instance (or a default instance if the cookie is absent).
+- me.set_cookie(instance) serialises the instance and schedules the browser
+  to set the cookie via the /__apply-cookies endpoint.
+- me.delete_cookie(SessionCookie) removes the cookie by class.
+
+The login state persists across hard refreshes and new tabs because the
+cookie is read in on_load.
+
+Note: me.set_cookie() and me.delete_cookie() require MESOP_COOKIE_SECRET_KEY
```
richard-to: Remove this line "Note: me.set_cookie() and me.delete_cookie() require MESOP_COOKIE_SECRET_KEY
to be set so that cookie tokens are cryptographically signed." Although it is true, it's too much detail. Just leave it with the example of how to start Mesop with MESOP_COOKIE_SECRET_KEY

PR #1377, opened 2026-04-12
Support multi-file demos in demo viewer

richard-to:
- Add `extra_files: list[str]` to `Example` dataclass for listing
  additional files (relative to the demo's own directory) to show
  in the code panel
- Add `selected_file` to `State`; reset it on page load so navigating
  between demos always starts on the main Python file
- Build `ALL_EXAMPLES` lookup dict for O(1) name→Example access
- Rewrite `demo_code()` to show a file-select dropdown when a demo
  has extra files; derive file paths via `module.__file__` so both
  flat and subdirectory demos work without special-casing
- Auto-detect language for the code fence (py/js/ts/css/html)

https://claude.ai/code/session_01Noq84VqXg7nnYcDJAnrWPu

### DIFF
--- a/demo/BUILD
+++ b/demo/BUILD
@@ -6,7 +6,10 @@ package(
 
 py_library(
     name = "demo",
-    srcs = glob(["*.py", "*/*.py"]),
+    srcs = glob([
+        "*.py",
+        "*/*.py",
+    ]),
     data = [
         ":screenshots",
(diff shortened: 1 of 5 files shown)


richard-to:
Ref: https://github.com/mesop-dev/mesop/issues/535

PR #1391, opened 2026-05-05
add event-driven auth recovery for WebSockets behind proxies

### LINKED ISSUE #1389: Dead websockets connections don't reconnect when the app is behind an auth proxy
When leaving a Mesop application open for a long period (e.g., overnight) in a corporate environment behind a proxy with Single Sign-On (SSO), the WebSocket connection eventually drops due to session expiry.

When the app attempts to reconnect or send a message:

1. The corporate proxy intercepts the request, sees that the auth cookie has expired, and attempts to return an HTTP 302 Redirect to send the user to the SSO login page.
2. The browser's native WebSocket API does not follow HTTP redirects. If a WebSocket handshake receives anything other than a 101 Switching Protocols response, it immediately aborts.
3. For security reasons, the browser does not expose the HTTP status code or location headers of the failed handshake to JavaScript; it just throws a generic, opaque WebSocket error: Event {isTrusted: true, type: 'error'}.
4. This leaves the app in a disconnected state without informing the user or allowing them to re-authenticate.

nirben82:
Opt-in, event-driven solution to allow applications to handle this gracefully without forcing a destructive full-page reload that loses user state.

1. HTTP Probe: When the connection is detected as dead during user interaction, the client performs a lightweight HTTP probe using fetch with `redirect: 'manual'`.
2. Event Emission: If a redirect is detected (indicating session expiry), the framework emits a `mesop:auth-required` CustomEvent on the `window` object, passing the target URL in the details.
3. Proactive Recovery: The framework listens for the `focus` event on the `window`. When the user returns to the tab (presumably after completing the authentication flow in another window or tab), it probes again and automatically reconnects if successful.

fixes #1389

### DIFF
--- a/mesop/env/env.py
+++ b/mesop/env/env.py
@@ -48,6 +48,11 @@ MESOP_WEBSOCKETS_ENABLED = (
   os.environ.get("MESOP_WEBSOCKETS_ENABLED", "false").lower() == "true"
 )
 
+MESOP_WEBSOCKETS_PROBE_ON_DISCONNECT = (
+  os.environ.get("MESOP_WEBSOCKETS_PROBE_ON_DISCONNECT", "false").lower()
+  == "true"
+)
+
 MESOP_HTTP_CACHE_JS_BUNDLE = (
   os.environ.get("MESOP_HTTP_CACHE_JS_BUNDLE", "false").lower() == "true"
 )
--- a/mesop/web/src/services/channel.ts
+++ b/mesop/web/src/services/channel.ts
@@ -59,6 +59,7 @@ export class Channel {
   private webSocket: WebSocket | undefined;
   private wsReconnectAttempts = 0;
   private wsMaxReconnectAttempts = 3;
+  private isProbing = false;
   private initParams!: InitParams;
   private states: States = new States();
   private stateToken = '';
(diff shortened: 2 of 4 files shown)


mesop/env/env.py
```diff
@@ -48,6 +48,10 @@ def get_base_url_path() -> str:
   os.environ.get("MESOP_WEBSOCKETS_ENABLED", "false").lower() == "true"
 )
 
+MESOP_AUTH_PROXY_ENABLED = (
+  os.environ.get("MESOP_AUTH_PROXY_ENABLED", "false").lower() == "true"
```
richard-to: Let's rename this so it's specific to web sockets since it wouldn't apply to the SSE path.

mesop/web/src/services/channel.ts
```diff
@@ -493,6 +522,28 @@ export class Channel {
...
+   *
+   * We use `redirect: 'manual'` to detect redirects without following them,
+   * which avoids CORS issues if the redirect points to a different domain.
+   *
+   * @returns true if a redirect was detected (meaning session expired), false otherwise.
+   */
+  private async probeConnection(): Promise<boolean> {
+    try {
+      const response = await fetch(window.location.href, {
+        method: 'GET',
```
richard-to: Would a HEAD request work here? Could be worth testing since that would be slightly lighter weight than getting the html page.

We may also want to add a empty health-check endpoint that we set to no cache (like my other PR).

mesop/web/src/services/channel.ts
```diff
@@ -493,6 +522,28 @@ export class Channel {
...
+        redirect: 'manual',
+      });
+      return response.type === 'opaqueredirect';
```
richard-to, line 550: This seems like it's specific to the implementation. Other auth proxy's may have other behaviors. I think that's fine though. We can just log a GitHub issue to add other support.

mesop/web/src/services/channel.ts
```diff
@@ -91,6 +91,21 @@ export class Channel {
       userEvent.setChangePrefersColorScheme(new ChangePrefersColorScheme());
       this.dispatch(userEvent);
     });
+
+    window.addEventListener('focus', async () => {
+      if (
+        this.status === ChannelStatus.CLOSED &&
```
richard-to: I wonder if we can directly check the websocket state instead. It seems like ChannelStatus.CLOSED is used in cases where the connection is fine.

We should also only check this if web sockets is enabled.

mesop/web/src/services/channel.ts
```diff
@@ -91,6 +91,21 @@ export class Channel {
...
+        this.experimentService.authProxyEnabled
+      ) {
+        console.debug('Window focused, probing connection...');
+        const isRedirected = await this.probeConnection();
```
richard-to: Is there any race condition that's possible here? Mainly trying to think what would happen if multiple probe connections were triggered.

nirben82: added isProbing flag to prevent concurrent async operations in TypeScript

mesop/web/src/services/channel.ts
nirben82: Sure thing - https://github.com/mesop-dev/mesop/issues/1392

richard-to:
I'll take another look in a couple days. Busy this week.

PR #1403, opened 2026-09-18
Fix update_dataclass_from_json for None/frozen nested dataclasses

### LINKED ISSUE #1401: update_dataclass_from_json cannot rebuild a nested dataclass that is None, or one that is frozen
`_recursive_update_dataclass_from_json_obj` builds a *list's* elements from the type hint:

```python
item_instance = type_hints[key].__args__[0]()
```

but rebuilds a *nested* dataclass by recursing into whatever the attribute already holds:

```python
setattr(instance, key, _recursive_update_dataclass_from_json_obj(attr, value))
```

Two shapes of state therefore cannot round-trip. Both raise inside `update_state`, **before any event handler runs**, so to a user the interaction silently does nothing at all:

1. **A field typed `X | None` holding `None`** — the walk recurses into the `None` and raises `Unhandled stateclass deserialization where key=…, instance=None`. This bites hardest *inside a list*, because mesop constructs each element itself and a freshly constructed element holds the default `None`. Any list of dataclasses with an optional nested dataclass field is affected on every interaction after one is populated.
2. **A frozen nested dataclass** — mesop populates by `setattr`, so `@dataclass(frozen=True)` raises `FrozenInstanceError`, whatever the attribute holds.

We hit (1) in production: a media picker paginated fine until the page of results included an item carrying an optional usage record, after which every further interaction did nothing, with a 200 going back to the client.

<details>
<summary>Repro — mesop 1.3.5, Python 3.14, stdlib only</summary>

```python
from dataclasses import dataclass, field

from mesop.dataclass_utils import dataclass_utils


@dataclass
class Usage:
    input_tokens: int = 0


@dataclass(frozen=True)
class FrozenUsage:
    input_tokens: int = 0


@dataclass
class Item:
    name: str = ""
    usage: Usage | None = None


@dataclass
class ListState:
    items: list[Item] = field(default_factory=list)


@dataclass
class FrozenState:
    usage: FrozenUsage = field(default_factory=FrozenUsage)


def show(label, sent, fresh):
    payload = dataclass_utils.serialize_dataclass(sent)
    try:
        dataclass_utils.update_dataclass_from_json(fresh, payload)
    except Exception as exc:
        print(f"{label}: {type(exc).__name__}: {exc}")
    else:
        print(f"{label}: ok -> {fresh}")


show("optional, inside a list",
     ListState(items=[Item(name="a", usage=Usage(input_tokens=7))]), ListState())
show("optional, held directly",
     Item(name="b", usage=Usage(input_tokens=7)), Item())
show("frozen nested dataclass",
     FrozenState(usage=FrozenUsage(input_tokens=7)), FrozenState())
```

Output:

```
optional, inside a list: MesopException: Unhandled stateclass deserialization where key=input_tokens, value=7, instance=None
optional, held directly: MesopException: Unhandled stateclass deserialization where key=input_tokens, value=7, instance=None
frozen nested dataclass: FrozenInstanceError: cannot assign to field 'input_tokens'
```

</details>

<details>
<summary>What worked for us, if it is useful</summary>

Constructing the nested value from the payload and assigning it — rather than instantiating an empty one and letting the walk mutate it — covers both cases at once, since it never calls `setattr` on the nested instance itself. Resolving the type from the annotation (stripping `| None`) is the same move the list branch already makes.

We are doing this outside mesop for now, by wrapping `_recursive_update_dataclass_from_json_obj`; it works because mesop resolves its own recursive call through the module global, so the wrapper is re-entered at every level.

</details>

richard-to:
_recursive_update_dataclass_from_json_obj rebuilt a nested dataclass field
by recursing into whatever the attribute currently held. That breaks for a
field typed `X | None` holding `None` (nothing to recurse into, raises
"Unhandled stateclass deserialization") and for a frozen nested dataclass
(setattr raises FrozenInstanceError). Both cases surface inside
update_state, before any event handler runs, so the interaction silently
does nothing.

Build nested dataclass values (including list items) directly from the
JSON payload via the constructor, resolving `X | None` to `X` from the
type hint, instead of instantiating empty and mutating with setattr.

Fixes #1401

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_019XKNGzaycAUrysh8HgHpPX


richard-to:
The `test` check failure is a single Playwright e2e timeout in `mesop/components/box/e2e/box_events_test.ts:18` ("test right click event"), waiting on `getByText('Is Click: True')`. This is unrelated to this PR's change (`mesop/dataclass_utils/dataclass_utils.py` and its test file only) — CI is green on the base commit (`main` @ 9ca3b56), and this test doesn't touch anything in the diff. It looks like a timing flake in an unrelated component.

I don't have permission to trigger a re-run from here (got a 403 on the rerun-failed-jobs API). Flagging so a maintainer can re-run the job; I'll keep watching this PR for further CI/review activity.

---
_Generated by [Claude Code](https://claude.ai/code)_

